PR #4, opened 2020-01-23
Bugfixes + new event: GeneratorUnlockEvent

RogerFK:
- Fixes main game's bug where SCP-268 isn't shown to spectators or SCP-079, [feature stated by the very first patch note of Megapatch 2](https://steamcommunity.com/games/700330/announcements/detail/1610528051343716315)
- Add an extra check for game console commands, so `ev.Color` can be set to `string.Empty` instead of having to null it. *Just in case*.

- Fixed DoorInteractionEvent basically not working, and fixed the identation on it
- Added new GeneratorUnlockEvent
- Now players will recieve a "Command not found." message in red so there's some feedback if there's no console command (if it works as I'm thinking)


EXILED_Events/Patches/GameConsoleQueryOverride.cs
```diff
@@ -14,7 +14,7 @@ public static bool Prefix(RemoteAdmin.QueryProcessor __instance, ref string quer
 					return true;
 				
 				Events.InvokeConsoleCommand(__instance.gameObject, query, encrypted, out string returnMessage, out string color);
-				if (color == null)
+				if (!string.IsNullOrEmpty(color))
```
galaxy119: this should be if (string.IsNullOrEmpty(color)) not if (!...), otherwise it will always change the color to white if the color is *NOT* null/empty, and hit an exception if it is.

RogerFK: Oh well, sure, my bad. Fixed.

RogerFK:
Added a few more commits doing a fix for DoorInteractionEvent basically not working at all. Also, make sure when merging to fix the if (true) I changed (either keep my version and delete the comment or just change the `||` with `&&`)

PR #7, opened 2020-01-26
Refactor projects

AgentBlackout:
### Refactoring code projects and minor code changes

Namely:

- Replaced underscores with dots
- Turned main into shared kernel
- Divided "main" into API shared
- Moved methods which shouldn't have been in Plugin.cs into appropriate helpers in Shared
- Removed exiled_debug config and replaced with compiler preprocessor directive.

This is a breaking change, and will need more testing.


galaxy119:
This would require a complete rewrite of many EXILED plugins for - from what I can see, no actual gain in anything. EXILED is not an API, it is an event framework for plugins to use. In that context I don't feel this restructuring is necessary. Furthermore, many of the changes don't really make sense, exiled_debug need to *not* be a compiler thing, because the Debug() method is used by plugins to send debug code to the server console, when that option is enabled, to allow some users to enable it to help plugin developers diagnose issues without needing to tell them to re-install a "debug" version of the framework. 

Keep in mind - EXILED itself doesn't do anything for the SL server, it simply loads plugins and provides events those plugins can subscribe to. The only "API" is the Plugin class itself, that all plugins must have implemented in order to be loaded.

FruitBoi:
3 words.

What the fuck

KadeDev:
4 words.
Please fucking go away.

PR #34, opened 2020-03-05
Added events, extensions and cleaned up some code

iopietro:
Added "WarheadKeycardAccessEvent" and "Scp079GainExperienceEvent" events
Added "IsNukeLeverEnabled", "IsNukeDetonated" and "IsLCZDecontaminated" map extensions
Added "GetHealth" and "SetHealth" player extensions


galaxy119:
Most of these changes are 1. Superfluous, 2. Not in the interest of thread efficiency or 3. Breaking to existing plugins.
I will add the extensions and possibly the events you did into EXILED, but I'm not going to accept the rest of these changes as they simply aren't necessary, and some will likely break existing plugins.

iopietro:
The only things that could break plugins are the inserted getters and setters, I can make a pull request of everything except that change.
Other changes won't break any plugin, I already tested them with my old ones.

PR #65, opened 2020-03-31
Remove OnReload when reloading all plugins

VirtualBrightPlayz:
Why reload *after* it's disabled *and* it's going to be re-enabled?


VirtualBrightPlayz:
I think I didn't quite get what OnReload is meant for, it's not to reload server configs, my bad.

PR #67, opened 2020-04-02
Sorting plugins by file name during loading and filtering by file extension

Nekonyx:
Sorting the plugins by name will help to set priority in some cases.
Filtering by file extension may also be needed if, for example, you do not want to load a plugin called "AnotherPlugin.dll.off".

### DIFF
--- a/EXILED_Main/PluginManager.cs
+++ b/EXILED_Main/PluginManager.cs
@@ -43,7 +43,10 @@ namespace EXILED
 				Directory.CreateDirectory(path);
 			}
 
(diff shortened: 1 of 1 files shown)


galaxy119:
The only libraries that should be loaded with a higher priority than others is dependencies, plugins themselves should never be doing anything that requires them to be loaded before or after others as it's bad convention.
Also Linux servers *can* load .so files as mono libraries (such as libopus.so for dissonance encoding as an opus stream) so only being able to have .dll libraries loaded is needlessly restricting.

PR #69, opened 2020-04-02
Ability to selectively enable, disable and reload the plugin

Nekonyx:
Adds commands:
"listplugins" - displays a list of all plugins (enabled and disabled plugins are marked with green/red colors)
"enableplugin <name>" - includes a specific plugin
"disableplugin <name>" - turns off a certain plugin
"reloadplugin <name>" - reboots a certain plugin.
"reloadallplugins" is a former "reloadplugins", renamed to avoid confusion.

Unfortunately, it has not been properly checked :(

### DIFF
--- a/EXILED_Events/ReloadCommandHandler.cs
+++ b/EXILED_Events/ReloadCommandHandler.cs
@@ -12,17 +12,126 @@ namespace EXILED
 			string[] args = ev.Command.Split(' ');
 			switch (args[0].ToLower())
 			{
-				case "reloadplugins":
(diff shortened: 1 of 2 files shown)


galaxy119:
Rejected because this feature will be part of the complete rewrite of EXILED Plugin handling in 2.0 after Scopophobia releases, until then we have no way of guarunteeing our 'enableplugin' won't just get a return if it's re-checking it's config, and since we have no global method of getting/setting config values it would likely just break anyways.

PR #71, opened 2020-04-03
Room's GetPlayers method

Nekonyx:


galaxy119:
This should be in the MapExtensions class, or a new extensions class for Rooms, not inside the API object class.
IE:

public static void GetHubs(this Room room) => Player.GetHubs.Where(h => h.GetCurrentRoom().Name == room.Name);

as an example, this can be called both with Map.GetHubs(room); or room.GetHubs();
Also using ReferenceHub.Hubs can return the hub of the server itself, please use Player.GetHubs() instead.

PR #72, opened 2020-04-04
 Added properties, command, bug fixes and code cleanup

iopietro:
- Added **AppDataDirectory**, **PluginsDirectory**, **ExiledDirectory**, **DependenciesDirectory**, **LoadedDependenciesDirectory**, **ConfigsPath**, **LogsPath** Properties.
- Added **reloadconfigs** command, to reload **plugin configs**.
- General **code cleanup**.
- Fixed a bug that would've throwed exceptions, if a server owner or a server moderator, had executed two or more commands in rapid succession.

### DIFF
--- a/EXILED_Events/EventHandlers.cs
+++ b/EXILED_Events/EventHandlers.cs
@@ -92,9 +92,8 @@ namespace EXILED.Patches
 				if (!EventPlugin.DeadPlayers.Contains(ev.Player))
 					EventPlugin.DeadPlayers.Add(ev.Player);
 
-				if (EventPlugin.DropInventory) {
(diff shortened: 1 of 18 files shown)


galaxy119:
Plugin.Config.Reload() doesn't force already read config values that are stored as variables to update, so it's use would be more or less pointless.

PluginManager.LogsPath's existance isn't verified before it is attempted to be appended to.

Renaming the class in EXILED_Permissions will break any plugins using those extensions.

You don't need to lock the file object before appending to it since the only time it's accessed is within the RaCommand event, which doesn't run asyncronously even if fired multiple times in rapid succession.

PR #79, opened 2020-04-14
Smoll fixes

SebasCapo:
- Added Lift.Elevator list
- Fixed StartNuke not working
- Fixed NukeDetonationTimer not working


galaxy119:
The fix for StartNuke was already made just wasn't pushed yet.
You never clear your Lift.Elevator list, so after a round ends and the map regenerates, they will be nullrefs
NukeDetonation timer *should* always use network syncvars, but will double check.

PR #82, opened 2020-04-28
New extensions

SebasCapo:
- Added new Extensions for ItemTypes.
- Added new health-related extensions for ReferenceHubs.
- Added a maybe unnecessary bool for ReferenceHubs, but might make things simpler overall.

### DIFF
--- a/EXILED_Main/Extensions/Item.cs
+++ b/EXILED_Main/Extensions/Item.cs
@@ -6,7 +6,7 @@ namespace EXILED.Extensions {
         public static float GetWeaponAmmo( this Inventory.SyncItemInfo item ) => item.durability;
 
         public static bool IsAmmo( this ItemType item ) =>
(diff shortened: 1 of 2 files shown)


FruitBoi:
Before I merge this, can i ask why you’re taking the maxHp of the player and subtracting the health of the player?

SebasCapo:
Because setting the health value would allow you to go over the Max HP amount, which might be nice for some plugins, but you can still do that with the SetHealth method, in case Adrenaline HP isn't good enough for you :P 

Just thought that there should be a method that let's you add HP without surpassing the MaxHP limit.

FruitBoi:
If we’re trying to not go over the max, wouldn’t Mathf.Clamp be a better option?

SebasCapo:
Oh, didn't know those two existed, I'mma check those two out and edit the file

SebasCapo:
Done, should be better now!

PR #83, opened 2020-04-29
Roundlock, Lobbylock, FriendlyFire variables & fix AddHealth

SebasCapo:
Easier and more intuitive access to Round-lock, Lobby-lock & Friendly-fire booleans.

### DIFF
--- a/EXILED_Main/Extensions/Player.cs
+++ b/EXILED_Main/Extensions/Player.cs
@@ -604,10 +604,9 @@ namespace EXILED.Extensions
 		/// <param name="player"></param>
 		/// <param name="amount"></param>
 		public static void AddHealth(this ReferenceHub player, float amount, bool capAtMax = false) {
-			if(capAtMax) {
-				if(player.playerStats.health + amount > player.playerStats.maxHP) amount = player.playerStats.maxHP - player.playerStats.health;
-				player.playerStats.health += amount;
-			} else player.playerStats.health += amount;
+			if(capAtMax && player.playerStats.health + amount > player.playerStats.maxHP)
+				amount = player.playerStats.maxHP - player.playerStats.health;
+			player.playerStats.health += amount;
 		}
 		
 		/// <summary>
--- a/EXILED_Main/Extensions/Player.cs
+++ b/EXILED_Main/Extensions/Player.cs
@@ -603,9 +603,7 @@ namespace EXILED.Extensions
 		/// </summary>
 		/// <param name="player"></param>
 		/// <param name="amount"></param>
-		public static void AddHealth(this ReferenceHub player, float amount) {
-			player.playerStats.health += amount;
-		}
+		public static void AddHealth(this ReferenceHub player, float amount) => player.playerStats.health += amount;
 		
 		/// <summary>
 		/// Adds the specified amount of health to a <see cref="ReferenceHub">player</see> without exceeding the maximum health value.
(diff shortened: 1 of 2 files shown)


js6pak:
Can't you just use properties instead of Get-/Set- methods? Is it even needed? Also incorrect formatting

galaxy119:
For extensions that change something that aren't instanced (ie, static variables/etc), it's prefered for you to use a get/set.
Something like (pseudo code below)
```
public static bool FriendlyFire
{
    get
    {
        return ServerConsole.FriendlyFire;
    };

    set
    {
        ServerConsole.FriendlyFire = value;
    }
}```

**Comment by SebasCapo** [comment]
Ah, I see, makes sense! I'mma make another commit later with those changes!

**Comment by js6pak** [comment]
Please fix formatting :< (what are you using for editing code? notepad?)

**Comment by SebasCapo** [comment]
I have to edit the github file :( Can't access my pc rn

File: EXILED_Main/Extensions/Player.cs [hunk 1]
```diff
@@ -603,8 +603,12 @@ public static Room GetCurrentRoom(this ReferenceHub player)
 		/// </summary>
 		/// <param name="player"></param>
 		/// <param name="amount"></param>
-		public static void AddHealth( this ReferenceHub player, float amount ) =>
-			player.playerStats.health = Mathf.Clamp(amount, 1, player.playerStats.maxHP);
+		public static void AddHealth(this ReferenceHub player, float amount, bool capAtMax = false) {
```
js6pak, line 606: Won't changing signature break old plugins?

SebasCapo: AFAIK it shouldn't 🤔, it'll just set capAtMax to false by default, but could be wrong

EXILED_Main/Extensions/Player.cs
```diff
@@ -603,8 +603,12 @@ public static Room GetCurrentRoom(this ReferenceHub player)
...
+			if(capAtMax) {
+				if(player.playerStats.health + amount > player.playerStats.maxHP) amount = player.playerStats.maxHP - player.playerStats.health;
```
js6pak, line 608: Whats the point of doing that?

SebasCapo: MaxHp cap :P

EXILED_Main/Extensions/Player.cs
js6pak: Still signature is changed, so plugins compiled with old version of EXILED will break

js6pak:
And AddHealth fix couldn't be just simply `player.playerStats.health = Mathf.Clamp(player.playerStats.health + amount, 1, player.playerStats.maxHP);`?

SebasCapo:
Uh, sure, I didn't really know how to use Mathf.Clamp, if that works I'd rather use it

SebasCapo:
Alright, should be better now!

js6pak:
`amount += player.playerStats.health; player.playerStats.health = capAtMax ? Mathf.Min(amount, player.playerStats.maxHP) : amount` would be shortest

galaxy119:
For setting FriendlyFire - 
Firstly, you need to set NetworkfriendlyFire, not friendlyFire

Secondly instead of doing a GetComponentsOfType<> why not 
```
foreach (ReferenceHub hub in Player.GetHubs())
    hub.SetFriendlyFire(value);
``` 
?

SebasCapo:
Damn, you're right, that's better! Updated with that method! ^^

FruitBoi:
Looks good.

PR #86, opened 2020-04-30
Add way to get players badge easier, Checks for players being NTF added

DefyTheRush:
An extension to ReferenceHub that will get the badge they have on. Instead of doing ReferenceHub.GetRank().BadgeText, they can just do ReferenceHub.GetBadgeName() to simplify it some more

### DIFF
--- a/EXILED_Main/Extensions/Item.cs
+++ b/EXILED_Main/Extensions/Item.cs
@@ -76,7 +76,7 @@ namespace EXILED.Extensions {
             || ItemType.KeycardScientistMajor == type || ItemType.KeycardSeniorGuard == type || ItemType.KeycardZoneManager == type;
         
         /// <summary>
-        /// Check if an <see cref="ItemType">picked up or dropped item</see> is a gun.
+        /// Check if an <see cref="Pickup">picked up or dropped item</see> is a gun.
         /// </summary>
         /// <param name="type"></param>
         /// <returns></returns>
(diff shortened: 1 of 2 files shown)


galaxy119:
I'd prefer if you changed IsNtf to use a switch-case instead of a long if statement, to reduce the number of method calls used to resolve it
```
switch(hub.GetRole())
{
    case NtfScientist:
    case NtfCadet:
    case NtfLieutenant:
    case NtfCommander:
        return true;
    default:
        return false;
}```
Something like that

DefyTheRush:
I will do that right away, thank you for letting me know! I was doing switch statements for my own projects but I don't know why I decided to make an if-statement for this now that I think about it

DefyTheRush:
I also added another extension which is useful for checking items if they area certain type if you pick up or drop them (ItemDroppedEvent and PickupItemEvent)

galaxy119:
IsGun is the same thing as IsWeapon, but doesn't check for a MicroHID, I think instead it would be more appropriate to attach a bool checkMicro = false onto the IsWeapon one, and then change the || ItemType.MicroHID to || (checkMicro && ItemType.MicroHID) rather than make an entirely new extension for that change.

Since IsWeapon was added in the current test build phase, and isn't an existing method from the full release, changing it's method signature isn't a big deal.

DefyTheRush:
Hmm, I see. I was doing this because Pickup and Inventory.SyncItemInfo was different but then I realized that ItemID returns an ItemType as SyncItemInfo does

DefyTheRush:
It says there's a confliction even though when I added these fixes 1-2 days ago, it said the branch could be merged with no issues. I'll look into this

FruitBoi:
Alright looks good

PR #93, opened 2020-05-08
Revert Mutefixer

sanyae2439:


FruitBoi:
We have to sync all the mutes to new players that are joining.

PR #95, opened 2020-05-15
Added option to register custom Damage Types

kirun9:
Added option to add custom Damage Types with death cause that is displayed on player ragdoll

### DIFF
--- a/EXILED_Events/EXILED_Events.csproj
+++ b/EXILED_Events/EXILED_Events.csproj
@@ -56,6 +56,7 @@
     <Compile Include="Patches\AnnounceScpTerminationEvent.cs" />
     <Compile Include="Patches\CancelMedicalItemEvent.cs" />
     <Compile Include="Patches\CcmAnnoyingLoadFix.cs" />
+    <Compile Include="Patches\DamageTypesPatch.cs" />
     <Compile Include="Patches\DoubleSpawnEventFix.cs" />
     <Compile Include="Patches\PlaceBloodEvent.cs" />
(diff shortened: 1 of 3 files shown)


galaxy119:
After deliberating abit on the idea, the contributors have agreed the feature, while well implemented, is not necessary, and could potentially cause issues as damage types are not synced to the clients. In the best case scenario, it would default to "UNDEFINED" on the client, and thus render the custom damage types merely cosmetic for the code.

kirun9:
I have done some checks and tests and here is what I have found:
`PlayerStats.HitInfo.GetDamageType()` is the only way that `DamageTypes.DamageType` is obtained in game code

Its obtained mostly by `RagdollManager` and `PlayerStats` classes (and all other calls are called in `NineTailedFoxAnnouncer` to announce case of death of SCP)

While spawning Ragdoll (during which process every call to `PlayerStats.HitInfo.GetDamageType()` in `RagdollManager` and `PlayerStats` classes is made) it done by `NetworkServer.Spawn(gameObject)` call and setting `gameObject.GetComponent<Ragdoll>().Networkowner` variable;

`gameObject.GetComponent<Ragdoll>().Networkowner` is type of `Ragdoll.Info` that contains `PlayerStats.HitInfo` which contains `DamageTypes.DamageType`

`Ragdoll.Info` constructor is still called on server side

By all that I can assume that `DamageTypes.DamageType` is server side variable and can be safely modified.
Yes. `DamageTypes.damageTypes` array is not synced with client, although it's synced by other variables (for example `Ragdoll.Info`)

I have run some tests with my friends and they were able to see custom death causes on ragdolls

PR #100, opened 2020-05-20
Fix calling a death event when a player has godmode & has unsyncedHp

iRebbok:


iRebbok:
Don't accept this PR yet, I'll make sure the HP calculation math is accurate, and make edits or let you know if it's correct.

iRebbok:
I've fixed some parts, but I'm still waiting for confirmation about the functionality.

iRebbok:
The tester tested and confirmed the functionality, no problems were found.

PR #118, opened 2020-06-05
New LocalReportEvent

iRebbok:
The same as #117.

### DIFF
--- a/Exiled.Events/Handlers/EventArgs/LocalReportEventArgs.cs
+++ b/Exiled.Events/Handlers/EventArgs/LocalReportEventArgs.cs
@@ -1,0 +1,49 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="LocalReportEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.Handlers.EventArgs
+{
+    using System;
+
+    /// <summary>
+    ///     Contains information about the report to local administrators.
+    /// </summary>
+    public class LocalReportEventArgs : EventArgs
+    {
+        /// <param name="issuer"><inheritdoc cref="Issuer"/></param>
+        /// <param name="target"><inheritdoc cref="Target"/></param>
+        /// <param name="reason"><inheritdoc cref="Reason"/></param>
+        /// <param name="isAllowed"><inheritdoc cref="IsAllowed"/></param>
+        public LocalReportEventArgs(API.Features.Player issuer, API.Features.Player target, string reason, bool isAllowed = true)
+        {
+            Issuer = issuer;
+            Target = target;
+            Reason = reason;
+            IsAllowed = isAllowed;
+        }
+
+        /// <summary>
+        ///     Report sender.
+        /// </summary>
+        public API.Features.Player Issuer { get; internal set; }
+
+        /// <summary>
+        ///     Target of the report.
+        /// </summary>
+        public API.Features.Player Target { get; internal set; }
+
+        /// <summary>
+        ///     Gets or sets the report reason.
+        /// </summary>
+        public string Reason { get; internal set; }
+
+        /// <summary>
+        ///     Returns or sets a value indicating whether the process can be processed or not.
+        /// </summary>
+        public bool IsAllowed { get; set; }
+    }
+}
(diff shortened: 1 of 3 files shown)


Exiled.Events/Handlers/EventArgs/LocalReportEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+            Issuer = issuer;
+            Target = target;
+            Reason = reason;
+            IsAllowed = isAllowed;
+        }
+
+        /// <summary>
+        ///     Report sender.
+        /// </summary>
+        public API.Features.Player Issuer { get; internal set; }
```
iopietro, line 32: You can directly put `using Exiled.API.Features` above, no to rewrite the namespace every time.

Exiled.Events/Handlers/EventArgs/LocalReportEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    ///     Contains information about the report to local administrators.
+    /// </summary>
+    public class LocalReportEventArgs : EventArgs
+    {
+        /// <param name="issuer"><inheritdoc cref="Issuer"/></param>
+        /// <param name="target"><inheritdoc cref="Target"/></param>
+        /// <param name="reason"><inheritdoc cref="Reason"/></param>
+        /// <param name="isAllowed"><inheritdoc cref="IsAllowed"/></param>
+        public LocalReportEventArgs(API.Features.Player issuer, API.Features.Player target, string reason, bool isAllowed = true)
```
iopietro, line 21: Pre-events should have `-ing` inside them, so like:
`LocalReportingEventArgs`

Exiled.Events/Handlers/Server.cs
```diff
@@ -49,6 +49,11 @@ public class Server
         /// </summary>
         public static event CustomEventHandler<ReportingCheaterEventArgs> ReportingCheater;
 
+        /// <summary>
+        ///     Called when a complaint about a player is sent to the local server administrators.
+        /// </summary>
+        public static event CustomEventHandler<LocalReportEventArgs> LocalReport;
```
iopietro, line 55: This will follow as above: `LocalReporting`

Exiled.Events/Handlers/Server.cs
```diff
@@ -97,6 +102,8 @@ public class Server
         /// <param name="ev">The <see cref="ReportingCheaterEventArgs"/> instance.</param>
         public static void OnReportingCheater(ReportingCheaterEventArgs ev) => ReportingCheater.InvokeSafely(ev);
 
+        internal static void OnLocalReport(LocalReportEventArgs ev) => LocalReport.InvokeSafely(ev);
```
iopietro, line 105: And `OnLocalReporting`, also don't make `On` methods `internal`, so plugin devs can have their patches and invoke them as they want.

Exiled.Events/Patches/Events/LocalReport.cs
```diff
@@ -0,0 +1,97 @@
...
+    ///         60:        return;
+    ///         61:     ...
+    ///         **: }
+    ///     </code>
+    ///
+    ///     Line numbering is shown, as are names of variables with object types.
+    /// </remarks>
+    [HarmonyPatch(typeof(CheaterReport), nameof(CheaterReport.CallCmdReport),
+        typeof(int), typeof(string), typeof(byte[]), typeof(bool))]
+    internal static class LocalReport
```
iopietro, line 44: Same as above, `LocalReporting` to match the EventArgs name.

Exiled.Events/Handlers/EventArgs/LocalReportEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+            IsAllowed = isAllowed;
+        }
+
+        /// <summary>
+        ///     Report sender.
+        /// </summary>
+        public API.Features.Player Issuer { get; internal set; }
+
+        /// <summary>
+        ///     Target of the report.
```
iopietro, line 35: Extra spaces.

Exiled.Events/Handlers/Server.cs
```diff
@@ -17,7 +17,7 @@ namespace Exiled.Events.Handlers
     /// <summary>
     /// Server related events.
     /// </summary>
-    public class Server
+    public static class Server
```
iopietro, line 20: For now, the class will remain non-static, but if in the 2.0.0 full release we won't change the event system again (like making it per-plugin instead of global) we could just set all events classes to static.

PR #121, opened 2020-06-06
RoleColor & GetSide accesibility

SebasCapo:
Added methods to get a RoleType's Color and more ways to access a Role/Team's Side

### DIFF
--- a/Exiled.API/Extensions/Role.cs
+++ b/Exiled.API/Extensions/Role.cs
@@ -13,7 +13,8 @@ namespace Exiled.API.Extensions
     /// <summary>
     /// A set of extensions for <see cref="RoleType"/>.
     /// </summary>
-    public static class Role {
+    public static class Role 
+        {
 
         /// <summary>
         /// Get a <see cref="RoleType">role's</see> <see cref="Color"/>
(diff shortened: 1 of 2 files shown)


Exiled.API/Extensions/Role.cs
```diff
@@ -5,13 +5,53 @@
 // </copyright>
 // -----------------------------------------------------------------------
 
+using Exiled.API.Enums;
```
galaxy119, line 8: Your using directives should be inside the namespace.

Exiled.API/Extensions/Role.cs
```diff
@@ -5,13 +5,53 @@
...
+            role.GetTeam().GetSide();
+
+        /// <summary>
+        /// Get a <see cref="Team">team's</see> <see cref="Side"/>
+        /// </summary>
+        /// <param name="team"></param>
+        /// <returns></returns>
+        public static Side GetSide(this Team team) 
+            {
+            switch(team) {
```
galaxy119, line 38: The return messages are not indented properly.

SebasCapo: Fixed!

Exiled.API/Extensions/Role.cs
SebasCapo: IHateThisButFixed!

SebasCapo:
Should be better now!

PR #130, opened 2020-06-22
doors lol

PintTheDragon:
I added the 079 door event. It might need some renaming to look better but overall, it works fine. You can use the event like:
`        public void Scp079TriggerDoor(ref Scp079TriggerDoorEvent ev)
        {
           //stuff here
        }`
Then register it with:
`Events.Scp079TriggerDoorEvent += Scp079TriggerDoor;`

It works very similarly to the tesla gate event, only it provides the Door object and tells you if the event opened/closed the door.


galaxy119:
The master branch is frozen, pending the release of Exiled 2.0 and Scopophobia. If you wish to make additions to EXILED, please do so on the 2.0.0 branch.

PR #138, opened 2020-07-06
fix ConsoleCommandEvent to work with the main game command system

ced777ric:


ced777ric:
Added ev.Allow which defaults to true, thus making it run the main game method for client commands, so the new .comand handler that got added in beta 18 gets used.
I did it in a way that it doesnt break current plugins, but they will need to be updated to set ev.Allow to false so the command won't also print "Command not found", this is basicly like the RemoteAdminCommandEvent.

PR #141, opened 2020-07-10
Kade can't spell dynamic

PintTheDragon:
I fixed a lot of spelling mistakes.


PintTheDragon:
Is that whole extensions part needed since 9ac6c7527098ed7f9533667a3c9e3c13ecd8beb3?

galaxy119:
No, the extensions class is no longer needed, though it would be nice to include a more legitimate, useful example of making an extension, so people know how to do them anyways.

PintTheDragon:
I'll remove that part then since it's just confusing. If we want a real extension, we can add it later.

PR #143, opened 2020-07-12
Some changes

iRebbok:
* packing in an object and using reflection when calling events is no longer required, which is slow
* CustomEventHandler<TEventArgs> no longer resolves any type in TEventArgs
* Harmony debug restored using debug build
----------------------
Also now there is a clean StackTrace and no manual cleaning is required as in #140.


Exiled.API/Exiled.API.csproj
```diff
@@ -38,7 +38,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Assembly-CSharp-Publicized">
-      <HintPath>$(EXILED_REFERENCES)\Assembly-CSharp-Publicized.dll</HintPath>
+      <HintPath>..\Refs\Assembly-CSharp-Publicized.dll</HintPath>
```
galaxy119, line 41: All references should use $(EXILED_REFERENCES) - the assembly provided in /Refs is only for those who need the publicized assembly for their own plugin projects.

iRebbok: Why do I need to duplicate assembly in `$(EXILED_References)`, if it can be imported from `Refs/`, I see this as superfluous action, if we provide the latest `Assembly-CSharp-Publicized.dll`, then why move it every update?

Exiled.Events/Handlers/Map.cs
```diff
@@ -7,14 +7,15 @@
 
 namespace Exiled.Events.Handlers
 {
-    using Exiled.API.Extensions;
     using Exiled.Events.EventArgs;
+    using Exiled.Events.Extensions;
+
```
galaxy119, line 12: Inconsistent spacing. Some files have a space blank line before static using declarations, some do not. I don't care either way or the other but make them all the same :pepeangry:

iRebbok: Yes, this happened when changing from `API` to `Events`, I'll fix it.

Exiled.API/Exiled.API.csproj
galaxy119: Consistency with all the other references?

iRebbok: Developer must have a complete set of updated refs according to the freshness of `Assembly-CSharp-Publicized.dll`?

PR #156, opened 2020-07-16
Fix config seralization if there is no override for IConfig

iRebbok:
Desc: When you uninstall a plugin, or there is another problem related to the inability to serialize the required config, we receive errors of this kind that break all serialization & configs -

The solution is that we just override the default configuration as we cannot instantiate the interface.


iRebbok:
I ran into another problem, the config was overwritten and I got an error about type inconsistencies. I'll watch it later.

iRebbok:
I think it's done, now I just delete the default config implementations.
Although I do it, I don't think it's the best implementation, I spent some time looking at the YamlDotNet source code, but there is a lot of it. I think the best option is to deserialize into object instead of IConfig (this will save all the fields), when serializing, insert the original tag (since it is replaced by DefaultConfigImplementation) and, if the configuration block is broken, replace it with the original one provided by this or that plugin.

iRebbok:
I also set the `Description` attribute in `IConfig` to `IsEnabled`, I think we should provide comments on the types that we operate on.

Exiled.API/Interfaces/IConfig.cs
```diff
@@ -15,6 +17,7 @@ public interface IConfig
         /// <summary>
         /// Gets or sets a value indicating whether the plugin is enabled or not.
         /// </summary>
+        [Description("Indicates whether the plugin is enabled or not")]
```
iopietro, line 20: The "problem" with putting the DescriptionAttribute here is that it won't be auto-generated inside configs by YamlDotNet.

iRebbok: Yes, wouldn't it be better to supply it?

iRebbok: I'll remove it. I thought it can get inherit attributes.

PR #158, opened 2020-07-17
Added reload events

PintTheDragon:
Adds a few events that are ran when the `reload` command is ran, so that plugins can respond to them easier.


iopietro:
I know it has already been merged, but can you please name every event with +ing or +ed, based on their order? Ed are fired after, ing are fired before, you can look at other events to find how to name them consistently.

PintTheDragon:
#165 changes the names.

PR #159, opened 2020-07-17
Revise the Map & Room API

iRebbok:
- Using ReadOnly collections instead of Lists to take away the ability to modify them.
- Take away the ability to assign a position and a name to the room.
- Updating API objects whenever the original unity object was destroyed (unity objects are managed from outside the managed CRL C# code, so they aren't collected by the GC and immediately destroyed).
- Change lazy initialization of a room zone to immediate.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -8,6 +8,7 @@
 namespace Exiled.API.Features
 {
     using System.Collections.Generic;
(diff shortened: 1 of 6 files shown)


iRebbok:
- Clearing the cache before changing the scene (restarting the round), GC is forcibly called there, which allows to instantly dispose of garbage.

PR #166, opened 2020-07-20
Fixed some events

PintTheDragon:
Fixed the 079 level and the 106 containing event.


FruitBoi:
penis

PintTheDragon:
It's only one commit now!!

PintTheDragon:
I'm going to have to test these out later, but they _should_ work.

PR #168, opened 2020-07-20
Revert "Fixed a bug that caused Map.Rooms Map.Doors Map.Lifts and Map…

iRebbok:
….TeslaGates to always return a null readonlycollection."

This reverts commit 8d448c13a3537c6584e9e45dba95e6824c89880e.

`ReadOnlyCollection`'s is a wrapper around accessing the main collections without copying all values.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -27,6 +27,11 @@ namespace Exiled.API.Features
         private static readonly List<Lift> LiftsValue = new List<Lift>();
         private static readonly List<TeslaGate> TeslasValue = new List<TeslaGate>();
 
+        private static readonly ReadOnlyCollection<Room> ReadOnlyRoomsValue = RoomsValue.AsReadOnly();
+        private static readonly ReadOnlyCollection<Door> ReadOnlyDoorsValue = DoorsValue.AsReadOnly();
+        private static readonly ReadOnlyCollection<Lift> ReadOnlyLiftsValue = LiftsValue.AsReadOnly();
+        private static readonly ReadOnlyCollection<TeslaGate> ReadOnlyTeslasValue = TeslasValue.AsReadOnly();
+
         /// <summary>
         /// Gets a value indicating whether the decontamination has been completed or not.
         /// </summary>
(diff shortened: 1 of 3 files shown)


galaxy119:
You're initializing the values of the ReadOnlyCollection's without ever updating them before returning their value when the property getter is called.

iRebbok:
It doesn't matter, the `ReadOnlyCollection` contains a reference to the original `IList` collection, so all read calls go to the original collection. 
As written on MSDN:
> After demonstrating the Count, Contains, Item[], and IList.IndexOf members, the code example shows that the ReadOnlyCollection<T> is just a wrapper for the original List<T> by adding a new item to the List<T> and displaying the contents of the ReadOnlyCollection<T>.

galaxy119:
Then we need to debug why the original list was always empty, and once we solve that we can re-implement this, although I'm not entierly sure why we need to in the first place.

iRebbok:
I'll look at this later.

iRebbok:
I used this code:
```csharp
            Exiled.Events.Handlers.Server.WaitingForPlayers += () =>
            {
                var doors = Exiled.API.Features.Map.Doors;
                var rooms = Exiled.API.Features.Map.Rooms;
                var teslas = Exiled.API.Features.Map.TeslaGates;
                var lifts = Exiled.API.Features.Map.Lifts;
                Log.Info($"DoorsCount{doors.Count} FirstDoorName{doors[0].name}");
                Log.Info($"RoomsCount{rooms.Count} FirstRoomName{rooms[0].Name}");
                Log.Info($"TeslasCount{teslas.Count} FirstTeslaName{teslas[0].name}");
                Log.Info($"LiftsCount{lifts.Count} FirstLiftName{lifts[0].name}");
            };
```

and got this result:
```yaml
[2020-07-21 09:51:37.303 +03:00] Incoming connections will be no longer delayed.
[2020-07-21 09:51:37.314 +03:00] [INFO] [**] DoorsCount138 FirstDoorNamePrisonDoor (2)
[2020-07-21 09:51:37.324 +03:00] [INFO] [**] RoomsCount76 FirstRoomNameLCZ_ClassDSpawn (1)
[2020-07-21 09:51:37.334 +03:00] [INFO] [**] TeslasCount4 FirstTeslaNameGate
[2020-07-21 09:51:37.345 +03:00] [INFO] [**] LiftsCount8 FirstLiftNameTestElevator
[2020-07-21 09:51:37.356 +03:00] Waiting for players...
```

I see no error.

iRebbok:
Ah, I see. It seems that not scripts are null, but their objects, but in this case it's a logic bug.

iRebbok:
It works fine now.

PR #171, opened 2020-07-24
Fix small spelling error

KoukoCocoa:


iopietro:
**D:\a\EXILED\EXILED\Exiled.Events\Exiled.Events.csproj" (default target) (2) ->
(CoreCompile target) -> 
  Patches\Generic\AntiFly.cs(23,64): error CS1061: 'Config' does not contain a definition for 'IsAntyFlyEnabled' and no accessible extension method 'IsAntyFlyEnabled' accepting a first argument of type 'Config' could be found (are you missing a using directive or an assembly reference?) [D:\a\EXILED\EXILED\Exiled.Events\Exiled.Events.csproj]**

KoukoCocoa:
It makes no sense, I fixed that and that was the only area where it was required x-x

PR #177, opened 2020-07-26
Fix tesla multi fire of death event

jeppevinkel:


iopietro:
For now, I'm going to merge it as a hotfix, but I'll probably rework the Died event to fix that bug in another way.

PR #181, opened 2020-07-29
Added inheritance support for Exiled.Permissions

jeppevinkel:
Added inheritance support for the permissions.
It works by adding the name of any groups to inherit from in the `Inheritance` list.

### DIFF
--- a/Exiled.Permissions/Features/Group.cs
+++ b/Exiled.Permissions/Features/Group.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------
+// -----------------------------------------------------------------------
 // <copyright file="Group.cs" company="Exiled Team">
 // Copyright (c) Exiled Team. All rights reserved.
 // Licensed under the CC BY-SA 3.0 license.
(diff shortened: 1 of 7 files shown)


Exiled.Permissions/Features/Group.cs
```diff
@@ -31,5 +38,25 @@ public class Group
         /// Gets or sets the group permissions.
         /// </summary>
         public List<string> Permissions { get; set; } = new List<string>();
+
+        /// <summary>
+        /// Gets the combined permissions of the group plus all inherited groups.
+        /// </summary>
+        [YamlIgnore]
+        public List<string> CombinedPermissions
```
iopietro, line 46: Shouldn't we set this Property only when `Permissions.Reload()` get called?

jeppevinkel: The current solution essentially does that, but it might be better to explicitly do it.

jeppevinkel:
I've changed it to all be done in `Permissions.Reload()` now.

I also added support for nested inheritance in descending order like this:
```yml
owner:
  inheritance:
  - admin
admin:
  inheritance:
  -  moderator
moderator:
  inheritance: []
```
It only works in that order, not the other way.
To make it work all ways would be more complicated, time consuming, and could result in infinite loops if not careful.
Currently if someone does it the wrong way, they simply don't get the permissions.

PR #190, opened 2020-08-02
Optimize the player global badge API

iRebbok:
- Using the structure prevents the `Badge` from getting into the managed heap.
- Optimization of getting `Badge` without creating a dictionary and manual parsing.

### DIFF
--- a/Exiled.API/Features/Badge.cs
+++ b/Exiled.API/Features/Badge.cs
@@ -30,21 +30,21 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets the badge text.
         /// </summary>
-        public string Text { get; }
+        public string Text { get; private set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Exiled.API/Features/Badge.cs
```diff
@@ -30,21 +30,21 @@ public Badge(string text, string color, int type, bool isGlobal = false)
         /// <summary>
         /// Gets the badge text.
         /// </summary>
-        public string Text { get; private set; }
```
iopietro, line 33: I was thinking about giving the possibility to use the `Badge` struct in configs, we shouldn't remove setters and we should create a public default constructor, so YamlDotNet will be able to deserialize it.

iRebbok: I've restored the private setters.

PR #191, opened 2020-08-02
Fix RA vulnerability

Qidan475:
Commands can bypass plugins' ra command handlers, if contains REQUEST_DATA PLAYER_LIST SILENT


iRebbok:
What do you mean? Your change doesn't affect work in any way.

PR #202, opened 2020-08-06
Implement IEquatable on Player.

sirmeepington:
Allows developers to use the equality operator when checking if two Player objects represent the same person.
- Implements `IEquatable<Player>` on Player


iopietro:
You can already do that by using default operators `!=` and `==` to compare player references, which is perfectly fine, since players are created only once, when they join and stored inside a dictionary.

galaxy119:
As pietro said, equality operator usage is already perfectly functional. We aren't SMOD, every event doesn't create new player objects, the are created when the player joins and stored in a dictionary until the player leaves. These changes are unnecessary.

PR #208, opened 2020-08-08
Update InteractingLocker event

iRebbok:
I've updated the events and added some useful properties, also marked the conflicting properties as obsoleted so as not to break the existing plugins.


iRebbok:
Uh, I see, in the last release we didn't have `InteractingLockerEventArgs::AccessToken`, I'll remove it.

PR #217, opened 2020-08-13
Add the Player::DisplayNickname API

iRebbok:
These changes are due to the fact that plugins, after assigning a display nickname, cannot get the player's original nickname, although in some cases this is necessary.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -141,7 +141,7 @@ namespace Exiled.API.Features
             get => ReferenceHub.characterClassManager.UserId;
             set
             {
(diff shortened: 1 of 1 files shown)


galaxy119:
Firstly - What did I say about PR's being made to master after a release candidate has been pre-released? :anger: 

Secondly - You need to make the Player.Nickname a get-only property. Northwood does not want plugins to be changing it, and have asked that we at least not include the ability to do so in the API.

iRebbok:
I remembered this today. And yes, I made the `Player::Nickname` get-only as you can see the diff.

Also, what about `Player::UserId`? It can be assigned. 🤔

PR #221, opened 2020-08-14
Added HczCurve RoomType.

14130820:


galaxy119:
Since the inclusion of this can be considered a bug fix, I will accept this PR. But note, Exiled is now in release candidacy, and will not be accepting PR's into the master branch for anything other than critical bug fixes. (I am making an exception for this, even though it's not critical)

Also I need your discord username so I can give you the Exiled Dev role.

14130820:
@Arith7544

PR #229, opened 2020-08-18
Added RoundSummary_ClassList to EndingRoundEventArgs as readonly.

ArithFeather:


iopietro:
Good addition, I forgot to add it during the rewrite of Exiled.

PR #234, opened 2020-08-23
Added Scp173.TurnedPlayers and Scp096.TurnedPlayers, fixed the tutorial can trigger 096 config option

Cyanox62:


iRebbok:
You haven't to ignore the warnings in `.editorconfig`, it'll violate our code conventions.
@Cyanox62 Please remove that from `.editorconfig`.

Cyanox62:
I accidentally added to the editor config, the last two commits removed the changes.

Exiled.API/Features/Scp096.cs
```diff
@@ -16,5 +18,10 @@ public class Scp096
         /// Gets or Sets a value indicating the max shield amount Scp096 can have.
         /// </summary>
         public static float MaxShield { get; set; } = 500;
+
+        /// <summary>
+        /// Gets a list of player ids who will be turned away from SCP-096.
+        /// </summary>
+        public static List<int> TurnedPlayers { get; private set; } = new List<int>();
```
galaxy119, line 25: Since the ghostmode patch already uses player objects for 173's shouldRotate stuff, why is this a List<int> instead of a List<Player> ?

Exiled.API/Features/Scp173.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// A set of tools to modify SCP-173's behaviour.
+    /// </summary>
+    public static class Scp173
+    {
+        /// <summary>
+        /// Gets a list of player ids who will be turned away from SCP-173.
+        /// </summary>
+        public static List<int> TurnedPlayers { get; private set; } = new List<int>();
```
galaxy119, line 20: Same as above

Exiled.API/Features/Scp096.cs
Cyanox62: A List instead of a List? I don't understand what you're saying.

galaxy119: Thanks github very cool..

I meant to ask, why is it a List int 
instead of a List Player

Cyanox62: I figured I'd make it a `List<int>` since `TargetGhosts` was also an int. I'll change them both though.

PR #236, opened 2020-08-24
Added Room.Doors and Player.IsUsingStamina properties

gamehunt:
Room.Doors property gets all doors in the room
 Player.IsUsingStamina property allows disabling stamina for certain players

### DIFF
--- a/Exiled.Events/Patches/Generic/StaminaUsagePatch.cs
+++ b/Exiled.Events/Patches/Generic/StaminaUsagePatch.cs
@@ -1,0 +1,27 @@
+// -----------------------------------------------------------------------
+// <copyright file="StaminaUsagePatch.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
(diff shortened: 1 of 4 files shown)


Exiled.Events/Patches/Generic/StaminaUsagePatch.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary>
+    /// Patches <see cref="Stamina.ProcessStamina()"/>.
+    /// </summary>
+    [HarmonyPatch(typeof(Stamina), nameof(Stamina.ProcessStamina))]
+    internal class StaminaUsagePatch
+    {
+        private static bool Prefix(Stamina __instance)
+        {
+            return Player.Get(__instance._hub).IsUsingStamina;
```
galaxy119, line 24: While this probably won't be an issue, you should do
```
return Player.Get(__instance._hub)?.IsUsingStamina``` since Player.Get() can return null.

gamehunt: Fixed

PR #241, opened 2020-08-29
Fixed rld plugins command

gamehunt:
I found out that the "rld plugins" command is not working correctly. Here is the fixed version

### DIFF
--- a/Exiled.Loader/Loader.cs
+++ b/Exiled.Loader/Loader.cs
@@ -242,12 +242,16 @@ namespace Exiled.Loader
                     plugin.Config.IsEnabled = false;
 
                     plugin.OnDisabled();
+
+                    plugin.OnUnregisteringCommands();
                 }
                 catch (Exception exception)
                 {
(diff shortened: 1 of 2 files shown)


Exiled.Loader/Loader.cs
```diff
@@ -242,13 +242,17 @@ public static void ReloadPlugins()
                     plugin.Config.IsEnabled = false;
 
                     plugin.OnDisabled();
+
+                    plugin.OnUnregisteringCommands();
```
galaxy119, line 246: I'd recommend unregistering commands before calling OnDisabled(), simple because OnDisable could null objects or instances of classes that the commands need to use, and if (albiet unlikely), the OnDisable() call is made, and then a plugin command is invoked before OnUnregisteringCommands() is called, it could cause a nullref etc.

It's **extremely** unlikely, but might aswell be safe than sorry.

gamehunt: Fixed, also changed that in DisablePlugins() method

PR #245, opened 2020-09-03
Fixed Player.IsZooming property and added API for weapon attachments

gamehunt:


galaxy119:
Why not replace the switch-cases for the enum into string parsing to just ``enum.ToString()`` ? Will that not work and look much neater?

gamehunt:
enum.ToString() will produce names like BlueDot, Sniper e.t.c, Isnt it? We need to get names like Blue Dot Sight or Sniper Scope, because game stores mods like that. I think using switch-case in this situation is much simpler than splitting camel case. Correct me if it can be done only with ToString(), my c# knowledge is rather poor

galaxy119:
For getting the modifications, you can do something like:
```
string modName = wep.mod_barrels[weapon.modBarrel].name.Replace(" ", string.Empty);
return (BarrelType)Enum.Parse(typeof(BarrelType), modName);
```
instead of
```
switch (wep.mod_barrels[weapon.modBarrel].name)
{
    case "Suppressor":
        return BarrelType.Suppressor;
    case "Silencer":
        return BarrelType.Silencer;
    case "Muzzle Brake":
        return BarrelType.MuzzleBrake;
    case "Heavy Barrel":
        return BarrelType.HeavyBarrel;
    case "Muzzle Booster":
        return BarrelType.MuzzleBooster;
 }
```
If the enum names are made to match the string of mod_x[weapon.modX].name

For setting them, you can basically do this in reverse like so:
```
public static string SplitCamelCase(this string input)
{
    return System.Text.RegularExpressions.Regex.Replace(input, "([A-Z])", " $1", System.Text.RegularExpressions.RegexOptions.Compiled).Trim();
}
```

Then convert the enum tostring like so:
```
string modName = BarrelType.MuzzleBreak.ToString().SplitCamelCase();
```

But again, would only work if all of the Enum names were made to match the string names used by base-game for mods.

For now, I'll accept the PR so this is in the next beta release (which will be done shortly), but I would recommend testing out these changes and implementing them in the future, as it will make the code look much cleaner.

PR #246, opened 2020-09-08
Events for workstation Activation/Deactivation and Weapon attachments code cleanup

gamehunt:
I've also made DisabledPatches HashSet for situations when two or more plugins want to disable single patch.


Exiled.Events/Events.cs
```diff
@@ -53,9 +53,9 @@ private Events()
         public static Events Instance => LazyInstance.Value;
 
         /// <summary>
-        /// Gets a list of types and methods for which EXILED patches should not be run.
+        /// Gets a set of types and methods for which EXILED patches should not be run.
         /// </summary>
-        public static List<MethodBase> DisabledPatches { get; } = new List<MethodBase>();
+        public static HashSet<MethodBase> DisabledPatches { get; } = new HashSet<MethodBase>();
```
galaxy119, line 58: This is a plugin breaking change.

While breaking changes are not always rejected, it's highly recommended to re-write things in a way that new features won't break existing plugins, and then marking the old one with the [Obsolete()] attribute, so that it still exists and functions for a few Exiled versions before being removed, to give plugins time to update to the new method.

In this case, a simple work around for this would be add the HashSet ontop of the List, change the processing of disabling patches to add each element in the now Obsolete list, to the hashset, and then use the hashset to disable the patches.

The list would then get marked with an attribute
```
[Obsolete("Use DisabledPatchHashSet instead.")]
List<MethodBase> DisabledPatches { get; } = new List<MethodBase>();
HashSet<MethodBase DisabledPatchHashSet { get; } = new HashSet<MethodBase();
```
That way using either one will work for plugins to disable patches, without breaking plugins that already exist and use the List<> version.
After a few releases, the obsolete List<> can then be removed with less risk of breaking plugins.

Exiled.Events/Patches/Events/Player/ActivatingWorkstation.cs
```diff
@@ -0,0 +1,65 @@
...
+        private static bool Prefix(WorkStation __instance, GameObject tabletOwner)
+        {
+            try
+            {
+                if (!__instance.CanPlace(tabletOwner) || __instance._animationCooldown > 0f)
+                {
+                    return false;
+                }
+
+                global::Inventory component = tabletOwner.GetComponent<global::Inventory>();
```
galaxy119: While not a problem, global:: declarations aren't necessary in C#, and usage of it doesn't really match the code style of the rest of the project.

This will not contribute to rejection of the PR, just merely an informational comment that you don't need to define global::

gamehunt: Oops, didn't notice this when deleting dnspy ​​stuff.

gamehunt: Accidentally copied it it from dnspy, fixed in 2fc1f4f

Exiled.Events/Events.cs
gamehunt: Fixed

PR #249, opened 2020-09-10
replace Scp173::TurnedPlayers with HashSet

iRebbok:
Replaced as in Scp096 in favor of performance.

Note: this is a breaking change, all plugins that used it need to be recompiled.


galaxy119:
Since this variable was added in a beta release, the breaking change can be done without needing to obsolete the original.

PR #254, opened 2020-09-21
Improve the updater version checker

iRebbok:
So, currently we have to notify people of the re-release to update a server. It can be inconvenient, especially on hosting platforms such as exiled hosting, because literally not all people know how to install Exiled manually.

This PR plans to fix it, including things:
- Comparison of the commit with the current version and the release version, and if the release commit was made earlier, then it updates Exiled.
- Refactoring the Updater code:
    - to use a thread instead of a task.
    - to separate the code that works with the GitHub API.


iRebbok:
At the moment I've decided that I wouldn't implement version checking with a commit because it's hard from the API side because for that I need to execute 3 requests (get all releases, get all tags, get a tag commit) and those are all unauthorized request, i.e. GH ratelimit goes brrr after some server restarts.

PR #258, opened 2020-10-02
Fixed Map.StartDecontamination()

gamehunt:
`DecontaminationController.Singleton.NetworkRoundStartTime = -1f;` will prevent client from useless updates (it wont do time/decontamination broadcasts and update timer after forced decontamination). There are no side-effects, because that var is not used anywhere except of DecontaminationController.


iRebbok:
Next time use fast-forward if you wanna merge into an exiled branch

PR #260, opened 2020-10-02
Added properties and methods to Player.cs & Map.cs

SebasCapo:
- Optimized Map.GetRandomSpawnPoint() method.

- Player.RemoteAdmin, Player.IsJumping & Player.MoveState properties.

- Map.ShowHint() method to show a hint to all players.


Exiled.API/Features/Map.cs
```diff
@@ -139,7 +167,14 @@ public static void Broadcast(ushort duration, string message, global::Broadcast.
         /// <returns>Returns the spawn point <see cref="Vector3"/>.</returns>
         public static Vector3 GetRandomSpawnPoint(this RoleType roleType)
         {
-            GameObject randomPosition = Object.FindObjectOfType<SpawnpointManager>().GetRandomPosition(roleType);
+            GameObject randomPosition;
+
+            if (spawnpointManager == null)
```
iRebbok, line 172: Don't assume that this works as expected, because after the scene change, the cached component isn't disposed and when we try to get a GameObject of that component we'll get a NRE.

You've to reset that component every restart round.

Exiled.API/Features/Map.cs
```diff
@@ -127,6 +129,32 @@ public static void Broadcast(ushort duration, string message, global::Broadcast.
...
+        /// </summary>
+        /// <param name="duration">The duration in seconds.</param>
+        /// <param name="message">The message that will be broadcast (supports Unity Rich Text formatting).</param>
+        public static void ShowHint(float duration, string message)
+        {
+            foreach (Player player in Player.List)
+            {
+                player.ShowHint(message, duration);
+            }
+        }
```
iRebbok, line 143: Two methods, why? I prefer `ShowHint(string, float)` as in the player class.

iRebbok (changes requested):

Exiled.API/Features/Map.cs
SebasCapo: Was trying to go for more flexibility, I was thinking of maybe requesting a change for ShowHint on the Player class, since there's not much of a reason for ShowHint to have a different Syntax from the Broadcast method, but that may be better to be discussed on the Discord chat. :p

Exiled.API/Features/Map.cs
SebasCapo: Didn't know that! Will revert it back if I don't find any workaround.

Exiled.API/Features/Map.cs
iRebbok: I really don't like implementing a method with a different order of the same arguments. This may be confusing for a developer as you can see the issue https://github.com/galaxy119/EXILED/issues/112

iRebbok (approved):

PR #262, opened 2020-10-04
Get current room fix

ArithFeather:
This should fix the issue with GetCurrentRoom returning new rooms.


iRebbok (approved):
Great job! But configure your ReSharper so that it doesn't make useless changes that create diff noise next time.

PR #263, opened 2020-10-04
Adding pocket dimension room + Fix outside/Map room getter

ArithFeather:
Using SECTR_Sector ensures it will get all rooms (Including pocket dimension).
Added error for calling Map.Room before rooms are created.
Added pocket dimension enums and setter for Room.Type().
Fixed the surface transform, need to use "Outside" because that's the correct transform for the surface.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -74,7 +74,7 @@ namespace Exiled.API.Features
                     if (roomTransforms.Count == 0)
                     {
                         ListPool<Transform>.Shared.Return(roomTransforms);
-                        throw new NullReferenceException("Plugin is trying to access Rooms before they are created.");
+                        throw new InvalidOperationException("Plugin is trying to access Rooms before they are created.");
                     }
 
                     // Add the surface transform "room".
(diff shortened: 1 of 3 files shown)


ArithFeather:
The Ids were unnecessary, my apologies.

Exiled.API/Features/Map.cs
```diff
@@ -60,14 +63,28 @@ public static ReadOnlyCollection<Room> Rooms
...
+                    // Search for SECTR_Sector instead of tag. It is faster and includes Pocket dimension.
+                    roomTransforms.AddRange(
+                        Object.FindObjectsOfType<SECTR_Sector>()
+                            .Select(sector => sector.transform));
+
+                    // If no rooms were found, it means a plugin is trying to access this before the map is created.
+                    if (roomTransforms.Count == 0)
+                    {
+                        ListPool<Transform>.Shared.Return(roomTransforms);
+                        throw new NullReferenceException("Plugin is trying to access Rooms before they are created.");
```
iRebbok, line 77:
```suggestion
                        throw new InvalidOperationException("Plugin is trying to access Rooms before they are created.");
```

Because that isn't really a null ref

iRebbok (changes requested):

Exiled.API/Features/Map.cs
ArithFeather: Sounds good!

iRebbok (approved):

PR #264, opened 2020-10-05
Changed Room to component (monobehaviour).

ArithFeather:
+Fixed lookup for Rooms, this should get -all- the rooms a player can possibly be in.
+Added public method "Map.FindParentRoom(GameObject)" An easy and efficient way to get the room for any game object inside of a room, including players and pickups.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -10,7 +10,6 @@ namespace Exiled.API.Features
     using System;
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
-    using System.Linq;
 
     using Exiled.API.Extensions;
 
@@ -36,6 +35,8 @@ namespace Exiled.API.Features
         private static readonly ReadOnlyCollection<Door> ReadOnlyDoorsValue = DoorsValue.AsReadOnly();
         private static readonly ReadOnlyCollection<Lift> ReadOnlyLiftsValue = LiftsValue.AsReadOnly();
         private static readonly ReadOnlyCollection<TeslaGate> ReadOnlyTeslasValue = TeslasValue.AsReadOnly();
+
+        private static readonly RaycastHit[] CachedFindParentRoomRaycast = new RaycastHit[1];
 
         private static SpawnpointManager spawnpointManager;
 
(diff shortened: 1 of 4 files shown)


Exiled.API/Features/Map.cs
```diff
@@ -37,6 +36,8 @@ public static class Map
         private static readonly ReadOnlyCollection<Lift> ReadOnlyLiftsValue = LiftsValue.AsReadOnly();
         private static readonly ReadOnlyCollection<TeslaGate> ReadOnlyTeslasValue = TeslasValue.AsReadOnly();
 
+        private static readonly RaycastHit[] CachedFindParentRoomRaycast = new RaycastHit[1];
```
iRebbok, line 39: Private fields shouldn't start with uppercase

Exiled.API/Features/Map.cs
```diff
@@ -63,28 +64,38 @@ public static ReadOnlyCollection<Room> Rooms
...
+                        ListPool<GameObject>.Shared.Return(roomObjects);
                         throw new InvalidOperationException("Plugin is trying to access Rooms before they are created.");
                     }
 
-                    // Add the surface transform "room".
+                    // Add the pocket dimension since it is not tagged Room.
+                    const string PocketPath = "HeavyRooms/PocketWorld";
+                    var pocket = GameObject.Find(PocketPath);
+                    if (pocket == null)
+                        throw new NullReferenceException("Pocket Dimension not found. The name or location in the game's hierarchy might have changed.");
```
iRebbok, line 83: Print just a warn log to avoid breaking the API when the game is updated

Exiled.API/Features/Map.cs
```diff
@@ -63,28 +64,38 @@ public static ReadOnlyCollection<Room> Rooms
...
+                    roomObjects.Add(pocket);
+
+                    // Add the surface since it is not tagged Room. Add it last so we can use it as a default room since it never changes.
                     const string surfaceRoomName = "Outside";
-                    roomTransforms.Add(GameObject.Find(surfaceRoomName).transform);
+                    var surface = GameObject.Find(surfaceRoomName);
+                    if (surface == null)
+                        throw new NullReferenceException("Surface not found. The name in the game's hierarchy might have changed.");
```
iRebbok, line 90: Same as above

Exiled.API/Features/Room.cs
```diff
@@ -73,49 +53,33 @@ public Room(string name, Transform transform, Vector3 position)
...
-        /// <summary>
-        /// Equality Comparer.
-        /// </summary>
-        /// <param name="obj">Other object.</param>
-        /// <returns>If the rooms are equal.</returns>
-        public override bool Equals(object obj) => obj is Room room && this == room;
+            room.Zone = FindZone(roomGameObject);
+            room.Type = FindType(roomGameObject.name);
+            room.Doors = FindDoors(roomGameObject);
+            room.FlickerableLightController = roomGameObject.GetComponentInChildren<FlickerableLightController>();
```
iRebbok, line 79: Why not just use the `Awake()` method since it's a component? I.e. move these actions to method

iRebbok (changes requested):

Exiled.API/Features/Room.cs
ArithFeather: This is a common technique in instantiating components where you want to pass them parameters (Just like a constructor), but don't want those parameters to be public.
Right now it's not important, but in case we ever want to pass private data to the room when we instantiate it, we can use the static factor method.

iRebbok:
Merged into development with my commit which resolves conversations. Thank you.

PR #267, opened 2020-10-07
Added 2 new events (fixed?)

xRoier:
- (Player) UsingRadioEvent
- (Map) DamagingWindow


Exiled.Events/Patches/Events/Player/UsingRadioBattery.cs
```diff
@@ -0,0 +1,35 @@
...
+            var ply = API.Features.Player.Get(__instance.gameObject);
+            if (ply != null && __instance.CheckRadio() && __instance.inv.items[__instance.myRadio].id == ItemType.Radio)
+            {
+                var ev = new EventArgs.UsingRadioBatteryEventArgs(ply, __instance, __instance.isTransmitting);
+                Player.OnRadioBatteryUsing(ev);
+                if (!ev.IsAllowed)
+                    return false;
+            }
+
+            return true;
```
iRebbok, line 32: Implement full functionality in prefixes because we won't check every patch with every update, so we can update the patch as soon as something breaks

Exiled.Events/Patches/Events/Map/DamagingWindow.cs
```diff
@@ -0,0 +1,37 @@
...
+                var ev = new EventArgs.DamagingWindowEventArgs(__instance, damage);
+                Handlers.Map.OnDamagingWindow(ev);
+                __instance.health -= ev.Damage;
+                if (__instance.health <= 0f)
+                {
+                    __instance.StartCoroutine(__instance.BreakWindow());
+                }
+            }
+
+            return false;
```
iRebbok, line 34: Look at (for an example) here: [clickable](https://github.com/galaxy119/EXILED/blob/af4697b39c33b8a7158e3ed0183bab722653491b/Exiled.Events/Patches/Events/Map/AnnouncingDecontamination.cs)

You don't need to implement the whole method, it can just be a short event call

iRebbok (changes requested):

PR #270, opened 2020-10-11
Fixed bug where MaximumRespawnAmount is always 15

RinTheWeeb:
Because of

` private void ReissueNextKnownTeam()
        {
            var team = SpawnableTeam;
            if (team != null)
            {
                MaximumRespawnAmount = team.Value.MaxWaveSize;
            }
}`

`
 set
            {
                nextKnownTeam = value;
                ReissueNextKnownTeam();
            }
`

And added some extra checks

### DIFF
--- a/Exiled.Events/Patches/Events/Server/RespawningTeam.cs
+++ b/Exiled.Events/Patches/Events/Server/RespawningTeam.cs
@@ -46,7 +46,7 @@ namespace Exiled.Events.Patches.Events.Server
                             .Where(p => p.IsDead && !p.IsOverwatchEnabled));
 
                     if (__instance._prioritySpawn)
-                        list.OrderBy(item => item.ReferenceHub.characterClassManager.DeathTime).ToList();
+                        list = list.OrderBy(item => item.ReferenceHub.characterClassManager.DeathTime).ToList();
                     else
                         list.ShuffleList();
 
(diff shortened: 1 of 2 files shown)


Exiled.Events/Patches/Events/Server/RespawningTeam.cs
```diff
@@ -46,16 +46,9 @@ private static bool Prefix(RespawnManager __instance)
                             .Where(p => p.IsDead && !p.IsOverwatchEnabled));
 
                     if (__instance._prioritySpawn)
-                    {
-                        var tempList = ListPool<API.Features.Player>.Shared.Rent();
-                        tempList.AddRange(list.OrderBy(item => item.ReferenceHub.characterClassManager.DeathTime));
-                        ListPool<API.Features.Player>.Shared.Return(list);
-                        list = tempList;
-                    }
+                        list.OrderBy(item => item.ReferenceHub.characterClassManager.DeathTime).ToList();
```
iRebbok, line 49: as far as I can tell, ToList() creates a new list and that list isn't assigned to the `list` variable

iRebbok:
IDK at all, it should work like setting `MaximumRespawnAmount` after `NextKnowTeam`

Exiled.Events/Patches/Events/Server/RespawningTeam.cs
RinTheWeeb: Fixed

RinTheWeeb:
> 
> 
> IDK at all, it should work like setting `MaximumRespawnAmount` after `NextKnownTeam`

After the line Handlers.Server.OnRespawningTeam(ev);

The limit got reset back to the default value `team.Value.MaxWaveSize` instead of keeping its original value. 

That was because of the of that function you made when ev.Spawnable is getted? (I think).

I wasn't 100% which part of that code caused it, but this resolved it at least.

iRebbok:
Ah, then it requires another fix, I'll close this PR once I get it to work correctly

iRebbok:
Fixed in development with another implementation

PR #275, opened 2020-10-24
API for SCP-330

gamehunt:
Idk why git decided to do something with .csproj files, but they are unchanged lol


iRebbok:
Okay, something went wrong, let me fix it

PR #278, opened 2020-10-28
Add all subcommand to the reload command

thomasjosif:
This just executes all of the subcommands and prints their responses for the reload command.

### DIFF
--- a/Exiled.Events/Commands/Reload/All.cs
+++ b/Exiled.Events/Commands/Reload/All.cs
@@ -19,6 +19,11 @@ namespace Exiled.Events.Commands.Reload
     /// </summary>
     public class All : ICommand
     {
+        /// <summary>
+        /// Gets static instance of the <see cref="All"/> command.
+        /// </summary>
+        public static All Instance { get; } = new All();
+
         /// <inheritdoc/>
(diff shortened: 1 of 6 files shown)


Exiled.Events/Commands/Reload/All.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <inheritdoc/>
+        public string Description { get; } = "Reload all configs and plugins.";
+
+        /// <inheritdoc/>
+        public bool Execute(ArraySegment<string> arguments, ICommandSender sender, out string response)
+        {
+            bool success = true;
+
+            if (!new Configs().Execute(arguments, sender, out string responsetemp))
```
iRebbok, line 36: Don't create a new instance for one call, make a static instance variable/property in command classes and also use the variable/property to register them in a parent command

iRebbok (changes requested):

Exiled.Events/Commands/Reload/All.cs
thomasjosif: Done!

iRebbok: Thank you!

PR #279, opened 2020-10-28
Player.DropItems() & Player.ClearInventory() modification

Thundermaker300:
* Added Player.DropItems() as a quick way to drop all items as opposed to clearing them.
* Modified Player.ClearInventory() to use Inventory.Clear() instead of clearing the items from the inventory manually.


Exiled.API/Features/Player.cs
```diff
@@ -1146,7 +1146,12 @@ public void ResetInventory(List<ItemType> newItems)
         /// <summary>
         /// Clears the player's inventory.
         /// </summary>
-        public void ClearInventory() => Inventory.items.Clear();
+        public void ClearInventory() => Inventory.Clear();
```
iRebbok, line 1149: `Inventory::Clear()` also affects the dropping of `AmmoBox`. I think you shouldn't have changed it.

iRebbok (changes requested):
Why did you move the `Exiled.nuspec` file? You shouldn't have changed its location.

Exiled.API/Features/Player.cs
```diff
@@ -1146,7 +1146,12 @@ public void ResetInventory(List<ItemType> newItems)
         /// <summary>
         /// Clears the player's inventory.
```
iRebbok, line 1147: Update the summary, like
```
Clears the player's inventory. Includes ammo and items.
```

Thundermaker300:
I deleted the nuspec file?? I didn't even touch it.. Okay, I'll make the changes

Thundermaker300:
I think I fixed it now.

iRebbok (approved):

PR #283, opened 2020-10-29
Creating three methods for enabling & disabling status effects

Thundermaker300:
* `Player::EnableEffect` (PlayerEffect effect, int duration) - Enable
* `Player::EnableEffect` (string effect, int duration) - Enable
* `Player::DisableAllEffects` () - Disable

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -12,6 +12,8 @@ namespace Exiled.API.Features
     using System.Linq;
     using System.Reflection;
 
+    using CustomPlayerEffects;
+
     using Exiled.API.Enums;
     using Exiled.API.Extensions;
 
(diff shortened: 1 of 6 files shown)


Exiled.API/Features/Player.cs
```diff
@@ -1184,6 +1186,31 @@ public void ShowHint(string message, float duration = 3f)
...
+                effect.Value.ServerDisable();
+            }
+        }
+
+        /// <summary>
+        /// Enables a status effect on this player.
+        /// </summary>
+        /// <param name="effect">The <see cref="PlayerEffect"/> to enable.</param>
+        /// <param name="duration">The amount of time the effect will be active for.</param>
+        public void EnableEffect(PlayerEffect effect, float duration = 0f) => ReferenceHub.playerEffectsController.EnableEffect(effect, duration);
```
iRebbok, line 1205: The best implementation is `EnableEffect<T>(float duration = 0f, bool 
addDurationIfActive = false) where T : PlayerEffect` because literally nobody will cache an effect object

iRebbok (changes requested):

Exiled.API/Features/Player.cs
Thundermaker300: true

Exiled.API/Features/Player.cs
```diff
@@ -1184,6 +1186,39 @@ public void ShowHint(string message, float duration = 3f)
...
+        /// Enables a status effect on this player.
+        /// </summary>
+        /// <typeparam name="T">The <see cref="PlayerEffect"/> to enable.</typeparam>
+        /// <param name="duration">The amount of time the effect will be active for.</param>
+        /// <param name="addDurationIfActive">If the effect is already active, setting to true will add this duration onto the effect.</param>
+        public void EnableEffect<T>(float duration = 0f, bool addDurationIfActive = false)
+        {
+            PlayerEffect playerEffect;
+            if (!ReferenceHub.playerEffectsController.AllEffects.TryGetValue(typeof(T), out playerEffect))
+                return;
```
iRebbok, line 1210: That's useless code, `PlayerEffectController::EnableEffect<T>(float, bool)` already does that

iRebbok (changes requested):

PR #286, opened 2020-10-31
Adding Map.SpawningItem event

Thundermaker300:
* Map.SpawningItem
  - ItemType Id (get, set): The Item ID to spawn.
  - Vector3 Position (get, set): The position of the item being spawned
  - Quaternion Rotation (get, set): The rotation of the item being spawned
  - bool IsAllowed (get, set)

### DIFF
--- a/Exiled.Events/EventArgs/SpawningItemEventArgs.cs
+++ b/Exiled.Events/EventArgs/SpawningItemEventArgs.cs
@@ -1,0 +1,57 @@
+// -----------------------------------------------------------------------
+// <copyright file="SpawningItemEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+    using System.Collections.Generic;
+
+    using Exiled.API.Features;
+
+    using UnityEngine;
+
+    /// <summary>
+    /// Contains all informations before the server spawns an item.
+    /// </summary>
+    public class SpawningItemEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SpawningItemEventArgs"/> class.
+        /// </summary>
+        /// <param name="id"><inheritdoc cref="Id"/></param>
+        /// <param name="pos"><inheritdoc cref="Position"/></param>
+        /// <param name="rot"><inheritdoc cref="Rotation"/></param>
+        /// <param name="isAllowed"><inheritdoc cref="IsAllowed"/></param>
+        public SpawningItemEventArgs(ItemType id, Vector3 pos, Quaternion rot, bool isAllowed = true)
+        {
+            Id = id;
+            Position = pos;
+            Rotation = rot;
+            IsAllowed = isAllowed;
+        }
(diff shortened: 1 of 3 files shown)


Exiled.Events/EventArgs/SpawningItemEventArgs.cs
```diff
@@ -0,0 +1,57 @@
...
+    public class SpawningItemEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SpawningItemEventArgs"/> class.
+        /// </summary>
+        /// <param name="id"><inheritdoc cref="Id"/></param>
+        /// <param name="pos"><inheritdoc cref="Position"/></param>
+        /// <param name="rot"><inheritdoc cref="Rotation"/></param>
+        /// <param name="isAllowed"><inheritdoc cref="IsAllowed"/></param>
+        public SpawningItemEventArgs(ItemType id, Vector3 pos, Quaternion rot, bool isAllowed = true)
```
iRebbok, line 29: Add the locked parameter too please <3

Exiled.Events/Patches/Events/Map/SpawningItem.cs
```diff
@@ -0,0 +1,41 @@
...
+        private static bool Prefix(RandomItemSpawner.SpawnerItemToSpawn __instance)
+        {
+            SpawningItemEventArgs ev = new SpawningItemEventArgs(__instance._id, __instance._pos, __instance._rot, true);
+
+            Handlers.Map.OnSpawningItem(ev);
+
+            if (!ev.IsAllowed)
+                return false;
+            if (ev.Id == __instance._id && ev.Position == __instance._pos && ev.Rotation == __instance._rot)
+                return true;
```
iRebbok, line 35: Useless code, instantiating `SpawnerItemToSpawn` doesn't affect anything because it's a struct, so we don't need to do that kind of check

iRebbok (changes requested):

Exiled.Events/Patches/Events/Map/SpawningItem.cs
```diff
@@ -0,0 +1,46 @@
...
+    using UnityEngine;
+
+    /// <summary>
+    /// Patches <see cref="RandomItemSpawner.SpawnerItemToSpawn.DoorTrigger"/>.
+    /// Adds the <see cref="Handlers.Map.SpawningItem"/> event.
+    /// </summary>
+    [HarmonyPatch(typeof(RandomItemSpawner.SpawnerItemToSpawn), nameof(RandomItemSpawner.SpawnerItemToSpawn.Spawn))]
+    internal static class SpawningItem
+    {
+        private static IEnumerator<float> Prefix(RandomItemSpawner.SpawnerItemToSpawn __instance)
```
iRebbok, line 29: This won't work, it you want to implement that kind of patch then it must be a Postfix and the first parameter must be `IEnumerator<float> values` (keep in mind, harmony is name sensitive)

iRebbok (changes requested):

Exiled.Events/Patches/Events/Map/SpawningItem.cs
iRebbok: Anyway, the requirements were in 2.0.1, correct me if I'm wrong and it works

iRebbok (approved):

PR #291, opened 2020-11-04
incorrect scp 096 ahp rate fixed

ismail0234:


ismail0234:
Pull-request number #292 continues.

PR #297, opened 2020-11-04
Adding Scp096::DestroyingDoor event

Thundermaker300:
Adding event `Scp096::DestroyingDoor` (Scp096 Scp096, Player Player, Door Door, bool IsAllowed) - Fires when a door is being charged by Scp096.

### DIFF
--- a/Exiled.API/Features/Cassie.cs
+++ b/Exiled.API/Features/Cassie.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------
+// -----------------------------------------------------------------------
 // <copyright file="Cassie.cs" company="Exiled Team">
 // Copyright (c) Exiled Team. All rights reserved.
 // Licensed under the CC BY-SA 3.0 license.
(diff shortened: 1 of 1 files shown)


Exiled.Events/Patches/Events/Scp096/DestroyingDoor.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    using HarmonyLib;
+
+    using PlayableScps;
+
+    /// <summary>
+    /// Patches the <see cref="Scp096.ChargeDoor"/> method.
+    /// Adds the <see cref="Handlers.Scp096.DestroyingDoor"/> event.
+    /// </summary>
+    [HarmonyPatch(typeof(Scp096), nameof(Scp096.ChargeDoor))]
```
iRebbok, line 21: I'd suggest using `Door::DestroyDoor096` as a target method of the patch in this case and patch it with a transpiler. The only reason of using ``Door::DestroyDoor096` is that there are some checks which determine whether a door can be destroyed or not. I can implement it myself if you don't know how.

iRebbok (changes requested):

Exiled.Events/Patches/Events/Scp096/DestroyingDoor.cs
Thundermaker300: If you'd like sure. I chose this method because patching Door::DestroyDoor096 would not actually give the 096 who destroyed it.

iRebbok: I mean, if we patch this method, then we can't be sure that IsAllowed = true will destroy a door, that's because the method has some checks before it destroys a door

PR #308, opened 2020-11-09
Additions to `Scp096::StartPrying` & `Scp096::AddingTarget` & `Scp049::FinishingRecall` & `Scp049::StartingRecall` events

Thundermaker300:
* Adding `Round::ForceSpawnTeam(SpawnableTeamType team, bool spawnVehicle = false)` - Quick way to spawn a reinforcement team.

### DIFF
--- a/Exiled.Events/EventArgs/AddingTargetEventArgs.cs
+++ b/Exiled.Events/EventArgs/AddingTargetEventArgs.cs
@@ -11,6 +11,8 @@ namespace Exiled.Events.EventArgs
 
     using Exiled.API.Features;
 
+    using Scp096 = PlayableScps.Scp096;
+
     /// <summary>
     /// Contains all informations before adding a target to SCP-096.
     /// </summary>
@@ -20,11 +22,13 @@ namespace Exiled.Events.EventArgs
         /// Initializes a new instance of the <see cref="AddingTargetEventArgs"/> class.
         /// </summary>
         /// <param name="scp096"><see cref="Player"/> who is SCP-096.</param>
+        /// <param name="player">The player who is playing as SCP-096.</param>
         /// <param name="target"><see cref="Player"/> who is the target to be added.</param>
         /// <param name="ahpToAdd"><see cref="int"/> amount of temporary health to add to <see cref="Scp096"/>.</param>
         /// <param name="enrageTimeToAdd"><see cref="float"/> amount of time to add to <see cref="Scp096"/>'s enrage timer. Note: This does not affect anything if he doesn't already have any targets before this event is called.</param>
-        public AddingTargetEventArgs(Player scp096, Player target, int ahpToAdd, float enrageTimeToAdd)
+        public AddingTargetEventArgs(Scp096 scp096, Player player, Player target, int ahpToAdd, float enrageTimeToAdd)
         {
+            Player = player;
             Scp096 = scp096;
             Target = target;
             AhpToAdd = ahpToAdd;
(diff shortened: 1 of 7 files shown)


iopietro:
You should move it to `Exiled.API.Features.Respawn`.

Thundermaker300:
I forgot that was a thing lol

Exiled.API/Features/Respawn.cs
```diff
@@ -23,6 +23,20 @@ public static class Respawn
         /// </summary>
         public static RespawnEffectsController Controller => RespawnEffectsController.AllControllers.Where(controller => controller != null).FirstOrDefault();
 
+        /// <summary>
+        /// Forces a reinforcement respawn.
+        /// </summary>
+        /// <param name="team">The team to respawn.</param>
+        /// <param name="spawnVehicle">Determines whether or not to spawn the vehicle as well as the team.</param>
+        public static void ForceSpawnTeam(SpawnableTeamType team, bool spawnVehicle = false)
```
iopietro, line 31: I would simply call it `Force(...)`, to call it by doing `Respawn.Force`.
You can use the `Respawn.SummonChaosInsurgencyVan` method instead of `ExecuteAllEffects`.

Thundermaker300: I feel like `Respawn.Force` is too simple of a name. That would be like having `Respawn.Summon` instead of `Respawn.SummonChaosInsurgencyVan`

As of the method, I think it's good how it is because I don't want to have to add more code to use the methods individually based on the team.

iRebbok: I'd call it `ForceSpawn` without the `Team` postfix, it might be clearer to others, also rename the second parameter to `executeSpawnEffect` (the main game devs call it the `effect`) or something, because `spawnVehicle` is more suitable for CI, but not for NTF.

iRebbok (approved):

Exiled.API/Features/Respawn.cs
iopietro: Yeah, but `Respawn.ForceSpawn` sounds like a repetition, so we should name it like `Respawn.ForceWave` or something similar.

Thundermaker300:
I meant to create a new branch & PR for that.. if you want to review it though there it is

iRebbok:
Why did you just force pushed other changes? We've lost `Round::ForceSpawn`...

iRebbok (changes requested):
Currently those changes I see are API breaking changes, I don't approve them in any way. What's the advantage of adding the player type into the event args?

Exiled.API/Features/Respawn.cs
iRebbok: Literally lost change. Marked as resolved.

Thundermaker300:
I found it odd that some APIs passed the PlayableScp and others didn't, albeit it would break the plugins using these events unless the fields were renamed. Regarding the ForceSpawn, that's something I can add back quickly later today, I accidentally pushed here instead of on a new branch (which is on me).

iRebbok:
I literally have no idea about the changes. @iopietro what do you think?

iopietro:
Adding a new property to `[...]EventArgs` classes won't break any plugin, unless the plugin instances the `[...]EventArgs` object itself.
If we don't want to use the constructor, we can get Scp049, Scp096 objects from the Player's ReferenceHub, by doing something like this:

`public Scp096 Scp096 => Player.ReferenceHub.scpsController.CurrentScp as Scp096`

Thundermaker300:
I think the concern iRebbok has is changing the fields to be different types (eg previously EnragingEventArgs.Scp096 was Player, now it is Scp096).

iopietro:
Right, I didn't notice it in `AddingTargetEventArgs`.
So, as @iRebbok said, that change would break every plugin which uses that property.

iRebbok:
I revised the changes and my opinion is: these changes doesn't really improve our event args, I can't say they are useless, but really, who needs an instance of Scp096 or Scp049? The instance doesn't contain extra information and putting the Player type there was correct. Although, I haven't seen other event args, so my opinion is only about this PR.

iopietro:
Yeah, those changes are unnecessaries, we could simply add a `Player.CurrentScp` property (witih null checks).

Thundermaker300:
Alright, in that case I'll go ahead and close this, thanks

PR #310, opened 2020-11-11
Update Exiled.Events.EventArgs.StartingEventArgs

TrickyBestia:
Added IsAutoNuke property.

### DIFF
--- a/Exiled.Events/EventArgs/StartingEventArgs.cs
+++ b/Exiled.Events/EventArgs/StartingEventArgs.cs
@@ -7,6 +7,8 @@
 
 namespace Exiled.Events.EventArgs
 {
+    using System.Reflection;
+
     using Exiled.API.Features;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


galaxy119:
Rejected - You can get the info needed for these without using reflection.

TrickyBestia:
@galaxy119, If you mean
`if (ev.Player == null || ev.Player == Server.Host)`
If you press Server events - Start round, ev.Player will be null (or Server.Host, i didn't check). Reflection works in 100%

galaxy119:
I mean you can access the AlphawarheadController singleton via AlphawarheadController._host, rather than using reflection, which is slow.

TrickyBestia:
@galaxy119, are you talking about I am using Warhead.Controller in my code?

galaxy119:
No, I'm talking about accessing the warhead controller via AlphaWarheadController.Host instead of using reflection, which is slow. You also made this PR against master, instead of against our development branch. Either way, I implemented this myself via f6096bb

PR #314, opened 2020-11-16
Add location to plugins

PintTheDragon:
This will let you get the location of plugin dlls, because Plugin.Assembly.Location doesn't work.

### DIFF
--- a/Exiled.API/Features/Plugin.cs
+++ b/Exiled.API/Features/Plugin.cs
@@ -39,9 +39,6 @@ namespace Exiled.API.Features
 
         /// <inheritdoc/>
         public Assembly Assembly { get; } = Assembly.GetCallingAssembly();
-        
-        /// <inheritdoc/>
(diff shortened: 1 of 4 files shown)


iopietro:
Why adding a `method` and a `Dictionary` instead of just a property in the `Plugin` class and `IPlugin` interface?

iRebbok:
I'd suggest `Dictionary<Assembly, string>` to keep track of all locations including dependencies. I think it's better than a simple property in the `Plugin` class. @iopietro doesn't it?

iopietro:
Sure, if we want to keep track of every assembly location, we can do that.

PintTheDragon:
Yeah. I'm going to redo these changes with assembly keys.

PintTheDragon:
There's a dictionary. People can use it. No helper stuff to get it but you can pretty easily get the assembly's location.

PR #319, opened 2020-11-22
Adding try/catch to ChangingMuteStatus and ChangingIntercomMuteStatus

Thundermaker300:


Exiled.Events/Patches/Events/Player/ChangingIntercomMuteStatus.cs
```diff
@@ -22,25 +23,33 @@ internal static class ChangingIntercomMuteStatus
...
+
+                    return false;
                 }
 
-                return false;
+                return true;
+            }
+            catch (Exception e)
+            {
+                Exiled.API.Features.Log.Error($"Exiled.Events.Patches.Events.Player.ChangingIntercomMuteStatus: {e}\n{e.StackTrace}");
```
iRebbok, line 50:
```suggestion
                Exiled.API.Features.Log.Error($"{typeof(ChangingMuteStatus).FullName}.{nameof(Prefix)}:\n{e}");
```
And do the same with another patch

iRebbok (changes requested):

Exiled.Events/Patches/Events/Player/ChangingIntercomMuteStatus.cs
Thundermaker300: gotcha, copied this from a different patch

iRebbok (approved):

PR #322, opened 2020-11-23
Added CalculateDuration method to Cassie

TrickyBestia:


Exiled.API/Features/Cassie.cs
```diff
@@ -40,5 +40,13 @@ public static void DelayedMessage(string message, float delay, bool isHeld = fal
...
             Timing.CallDelayed(delay, () => RespawnEffectsController.PlayCassieAnnouncement(message, isHeld, isNoisy));
         }
+
+        /// <summary>
+        /// Calculates duration of C.A.S.S.I.E message.
+        /// </summary>
+        /// <param name="message">The message, which duration will be calculated.</param>
+        /// <returns>Duration (in seconds) of specified message.</returns>
+        public static float CalculateDuration(string message)
+            => NineTailedFoxAnnouncer.singleton.CalculateDuration(message, false);
```
iRebbok, line 50:
```suggestion
        /// <param name="rawNumber">Determines when a number won't be converted to its full pronunciation.</param>
        /// <returns>Duration (in seconds) of specified message.</returns>
        public static float CalculateDuration(string message, bool rawNumber = false)
            => NineTailedFoxAnnouncer.singleton.CalculateDuration(message, rawNumber);
```

iRebbok (changes requested):

iRebbok (approved):

PR #326, opened 2020-11-24
Added SCP-173 Blinking event

TrickyBestia:


iRebbok (commented):
I don't approve this PR.

We already have `Scp173::TurnedPlayers`, which allows you to turn away specified players from Scp173. But for Scp173 itself it literally doesn't work due to client-sided checks.
Also we've already patched `Scp173PlayerScript::FixedUpdate` via a transpiler [here](https://github.com/galaxy119/EXILED/blob/a92eefa457da458dcfc11da80cbd641f988dfae1/Exiled.Events/Patches/Generic/Scp173BeingLooked.cs).

Your PR doesn't look good to me because it does the same thing as in the `Scp173PlayerScript::FixedUpdate` method. If you really want that event, use 
`Scp173PlayerScript::DoBlinkingSequence` instead -> we mustn't patch generated by Mirror methods.

galaxy119:
This could still be useful for doing other things, not related to who's triggering the event, whenever he blinks. 

For example

`
if (ev.Targets.Length > 4)
    ev.Player.Health += 5;
`
or something.

iRebbok:
Sure, but I can't accept that this event does the same as in the `FixedUpdate` method.

@TrickyBestia CC me after you finish your work, I'll improve performance.

TrickyBestia:
> @TrickyBestia CC me after you finish your work, I'll improve performance.
@iRebbok i finished. What do you want to do?

iRebbok:
Finally a good event..

Firstly - creating a new list on every call makes no sense.
Secondly - doing the same as in the `FixedUpdate` method doesn't needed, just copy & paste the code from the RPC.

iRebbok:
About the `Duration` property: I'll be doing refactoring this event into a transpiler and will add it.

iRebbok:
[[Maintenance] Exiled](https://trello.com/c/AMfeUCdl/24-maintenance-exiled)

PR #327, opened 2020-11-24
Some useful properties in Player API (IsHuman, IsScp)

iRebbok:
Why didn't we have that before...


galaxy119:
Might aswell add an IsScp aswell

PR #330, opened 2020-11-30
Added Player::IsConnected property

TrickyBestia:
It should be useful when you are doing something with player in the coroutine.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -253,6 +253,11 @@ namespace Exiled.API.Features
         public bool DoNotTrack => ReferenceHub.serverRoles.DoNotTrack;
 
         /// <summary>
+        /// Gets a value indicating whether the player is connected to the server or not.
+        /// </summary>
+        public bool IsConnected => List.Contains(this);
+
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Player.cs
```diff
@@ -252,6 +252,11 @@ public string CustomPlayerInfo
         /// </summary>
         public bool DoNotTrack => ReferenceHub.serverRoles.DoNotTrack;
 
+        /// <summary>
+        /// Gets a value indicating whether the player is connected to the server or not.
+        /// </summary>
+        public bool IsConnected => List.Contains(this);
```
iRebbok, line 258: Bad implementation, when a client is disconnected, Mirror destroys its GameObject (destroying a GameObject includes its components), the simple thing you must do is check a component for nullability, it can be done with ReferenceHub itself

iRebbok (changes requested):

TrickyBestia:
Changed it. Now I'm doing an equality check with null, because it returns true when GameObject is destroyed.

iRebbok (approved):

PR #332, opened 2020-12-01
Adding Lift.Type() extension

Thundermaker300:
Adding Lift.Type() extension to match Door.Type(). Previously, ElevatorType enum was only used in InteractingElevatorEventArgs. Now it can be used anywhere via the Lift.Type() method. I also marked InteractingElevatorEventArgs.Type as obsolete, as the DoorType isn't in the InteractingDoorEventArgs, it shouldn't be here either (now that Lift.Type() is available).

### DIFF
--- a/Exiled.API/Extensions/LiftTypeExtension.cs
+++ b/Exiled.API/Extensions/LiftTypeExtension.cs
@@ -1,0 +1,99 @@
+// -----------------------------------------------------------------------
+// <copyright file="LiftTypeExtension.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Extensions
+{
+    using System.Collections.Generic;
+
+    using Exiled.API.Enums;
+    using Exiled.API.Features;
+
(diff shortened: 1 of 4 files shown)


Exiled.API/Extensions/LiftTypeExtension.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// Gets the <see cref="ElevatorType"/>.
+        /// </summary>
+        /// <param name="lift">The <see cref="Lift"/> to check.</param>
+        /// <returns>The <see cref="ElevatorType"/>.</returns>
+        public static ElevatorType Type(this Lift lift) => OrderedElevatorTypes.TryGetValue(lift.GetInstanceID(), out var elevatorType) ? elevatorType : ElevatorType.Unknown;
+
+        /// <summary>
+        /// Gets all the <see cref="ElevatorType"/> values for for the <see cref="Lift"/> instances using <see cref="Lift.elevatorName"/> and <see cref="UnityEngine.GameObject"/> name.
+        /// </summary>
+        internal static void RegisterElevatorTypesOnLevelLoad()
```
iRebbok, line 33: Call that method from the `Map::Lifts` getter, the same implementation as with the doortype exception

iRebbok (approved):

PR #333, opened 2020-12-02
Development docs & Miscellaneous API addition

Thundermaker300:
* Doc stuff: as always
* Moving Player::GetCameraById to (static) Map::GetCameraById (having it as a non-static in the player class literally made no sense)
* Adding an overload to Player::SetCamera to take a camera object

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -949,6 +949,13 @@ namespace Exiled.API.Features
         public void SetCamera(ushort cameraId) => ReferenceHub.scp079PlayerScript?.RpcSwitchCamera(cameraId, false);
 
         /// <summary>
+        /// Sets the camera the player is currently located at.
+        /// Only applies if the player is SCP-079.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


PintTheDragon (changes requested):
This should be added back with a deprecation notice. Any plugins using it will break.

PintTheDragon:
Wait I think I did it wrong fuck. It's the line where you remove the function from Player.

Exiled.API/Features/Player.cs
```diff
@@ -925,28 +937,24 @@ public static Player Get(string args)
             }
         }
 
-        /// <summary>
```
PintTheDragon, line 928: This should be added back with a deprecation notice. Any plugins using it will break.

PintTheDragon:
I think I did it right this time.

PintTheDragon:
Wait I'm blind lmao you already did it

PintTheDragon (approved):
Yeah it's good

Thundermaker300:
Thanks, Pint

iRebbok (approved):

PR #339, opened 2020-12-07
Adding and modifying methods for Player effects

Thundermaker300:
* <s>`Player::GetActiveEffects()` (PlayerEffect[]) - Returns an array of active player effects</s>
* <s>`Player::EnableEffects(string[], float, bool)` (bool) - Enables a list of effects for duration, returns true if all successful</s>
* `Player::EffectActive<T>()` (bool) - Returns if given type is active
* `Player::ChangeEffectIntensity<T>(byte)` (void) - Changes the effect intensity
* `Player::ChangeEffectIntensity(string, byte)` (void) - Changes the effect intensity
* `Player::GetEffectIntensity<T>()` (byte)` - Returns effect intensity
* Modified: `Player::DisableAllEffects` to only call disabled on the enabled effects
(Apologies for like my 20th PR in the past week)

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1221,7 +1221,6 @@ namespace Exiled.API.Features
             HintDisplay.Show(new TextHint(message, parameters, null, duration));
         }
 
-
         /// <summary>
         /// Gets a <see cref="bool"/> describing whether or not the given <see cref="PlayerEffect">status effect</see> is currently enabled.
         /// </summary>
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1222,12 +1222,33 @@ namespace Exiled.API.Features
         }
 
         /// <summary>
+        /// Returns a list of active status effects on the player.
+        /// </summary>
+        /// <returns>A <see cref="PlayerEffect"/> array indicating active status effects.</returns>
+        public PlayerEffect[] GetActiveEffects()
+        {
+            PlayerEffect[] arr = new PlayerEffect[] { };
+            foreach (KeyValuePair<Type, PlayerEffect> data in ReferenceHub.playerEffectsController.AllEffects)
+            {
+                if (data.Value.Enabled)
+                {
+                    arr.Append(data.Value);
+                }
+            }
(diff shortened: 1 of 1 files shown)


Thundermaker300:
I think that's enough commits lol

Exiled.API/Features/Player.cs
```diff
@@ -1222,23 +1222,58 @@ public void ShowHint(string message, float duration = 3f)
...
+            foreach (KeyValuePair<Type, PlayerEffect> data in ReferenceHub.playerEffectsController.AllEffects)
+            {
+                if (data.Value.Enabled)
+                {
+                    arr.Append(data.Value);
+                }
+            }
+
+            return arr;
+        }
```
iRebbok, line 1240: Is it really useful? It literally allocates N arrays, where N is the number of enabled effects. I don't find it so useful.

Exiled.API/Features/Player.cs
```diff
@@ -1247,14 +1282,70 @@ public void EnableEffect<T>(float duration = 0f, bool addDurationIfActive = fals
...
+            {
+                bool success = ReferenceHub.playerEffectsController.EnableByString(str, duration, addDurationIfActive);
+                if (!success)
+                {
+                    flag = false;
+                }
+            }
+
+            return flag;
+        }
```
iRebbok, line 1313: Same as with `GetActiveEffects()`, does anyone need to enable a bunch of events with the same options?

Thundermaker300: I've had a case before where I could've used this but it wasn't available, although I see your point

Exiled.API/Features/Player.cs
Thundermaker300: fair enough

Exiled.API/Features/Player.cs
iRebbok: I mean, it'd be very useful.. but using types instead of string

iRebbok:
@Thundermaker300 ready?

Thundermaker300:
@iRebbok yeah, I don't think there's anything else necessary for this PR, unless you have ideas

PR #341, opened 2020-12-07
Move isNTF from Player to Role

TheVoidNebula:
I think it would fit more in that class. ^^

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -359,11 +359,6 @@ namespace Exiled.API.Features
         }
 
         /// <summary>
-        /// Gets a value indicating whether the player's role type is any NTF type <see cref="ReferenceHub"/>.
-        /// </summary>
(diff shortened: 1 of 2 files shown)


Exiled.API/Features/Player.cs
```diff
@@ -358,11 +358,6 @@ public Camera079 Camera
             set => SetCamera(value.cameraId);
         }
 
-        /// <summary>
-        /// Gets a value indicating whether the player's role type is any NTF type <see cref="ReferenceHub"/>.
-        /// </summary>
-        public bool IsNTF => Team == Team.MTF;
```
Thundermaker300, line 364: Don't remove API like this, as this breaks plugins. Add the `[Obsolete("Message here")]` attribute and it'll be removed before the next full release.

TheVoidNebula: should I make a new commit?

PR #343, opened 2020-12-07
move isNTF

TheVoidNebula:
it would be fitting if it would be removed from the Player class and instead being used in the Role class. It fits there more then in the Player class.


iRebbok:
It's not necessary. The property was made about a quick way to check if a player is ntf. Now you can make it easer to check if a player is ntf by using `if (ev.Player.IsNTF) { /* ... */ }` instead of other solutions.

PR #345, opened 2020-12-09
EffectType Enum

TheVoidNebula:
I think we could do with effects as an enum then just in the Player class. And also it would help newer developers to understand and find all different effect types instead of searching a long time for them (they are nowhere documented) ^^


TheVoidNebula:
while being here, I also have another Idea. What about having data, global data. Like data you can set and get from things like doors and players. maybe you do not only want to get all players but also all doors? or like all ragdolls etc.

iRebbok (commented):
I like this idea. But it requires correct documentation and API. So I don't see any API additions in this PR.

About global data: not understandable. If you are talking about something like a dictionary with keys for a player in their object - it won't be implemented. Use other ways.

TheVoidNebula:
what exactly you mean by API additions? ^^

iRebbok:
At least a small part of API that uses the enum.

TheVoidNebula:
like this?

iRebbok:
@Thundermaker300 do you have any docs suggestions/corrections? Review if you do 🙃

TheVoidNebula:
thanks, iRebbok for fixing my messy code ^^

TheVoidNebula:
I do not think he wants to add anything :) if so he would have already done it xD

Thundermaker300:
Haven't really had the chance to look at it yet

Exiled.API/Enums/EffectType.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// </summary>
+        Burned,
+
+        /// <summary>
+        /// Blurs the player's screen when rotating.
+        /// </summary>
+        Concussed,
+
+        /// <summary>
+        /// Teleports thet player to the pocket dimension and drains health.
```
iRebbok, line 46:
```suggestion
        /// Teleports the player to the pocket dimension and drains health.
```

PR #346, opened 2020-12-10
Bunch of stuff related to cameras

Thundermaker300:
Looking for some feedback here
* Not sure about some of the enum docs
* The cameras have been stored as an array whereas everything else is lists, and it feels weird - I know Scp079PlayerScript provides all the cameras as an array, but I'm not sure how to go about changing it (if we even should).


Thundermaker300:
@iRebbok Any advice for this?

iRebbok:
@Thundermaker300 Wdym changing the array in Scp079PlayerScript?

Exiled.API/Enums/CameraType.cs
```diff
@@ -0,0 +1,338 @@
...
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Enums
+{
+    /// <summary>
+    /// Unique identifier for the different types of SCP-079 cameras.
+    /// </summary>
+    public enum CameraType
+    {
+        // Light
```
iRebbok, line 15: Use regions instead, also disable the warning StyleCop gives about it

Thundermaker300:
> Wdym changing the array in Scp079PlayerScript?
Not changing the array in the Scp079PlayerScript, but changing it from an array to a List<Camera079> in the actual Map class, so that it matches the rest of the API already existing.

iRebbok:
> Not changing the array in the Scp079PlayerScript, but changing it from an array to a List<Camera079> in the actual Map class, so that it matches the rest of the API already existing.

Yep, free fell to change it for API cleanliness

Thundermaker300:
Should I depreciate the old API for the array or just change it to a list? If I do depreciate it, what should the new API be called? I'm honestly not sure if just replacing it would be a bad idea - I'm not sure many plugins use the array (and we could make an announcement telling people to expect the change)

iRebbok:
> Should I depreciate the old API for the array or just change it to a list? If I do depreciate it, what should the new API be called? I'm honestly not sure if just replacing it would be a bad idea - I'm not sure many plugins use the array (and we could make an announcement telling people to expect the change)

Go ahead and change it if the API hasn't been released as full release yet.

Thundermaker300:
Nono, I mean the Map.Cameras API (Camera079[]), I think it's been here for quite a bit

Exiled.API/Enums/CameraType.cs
```diff
@@ -0,0 +1,343 @@
...
+    /// Unique identifier for the different types of SCP-079 cameras.
+    /// </summary>
+    public enum CameraType
+    {
+        /// <summary>
+        /// Represents an unknown camera.
+        /// </summary>
+        Unknown = 0,
+
+#pragma warning disable SA1124 // Do not use regions
```
iRebbok, line 20: Move this line up after the namespace, as we do in event files 😏

Thundermaker300:
Probably best to make this change, however you might want to announce that it's being changed w/o warning

Exiled.API/Extensions/CameraExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <param name="camera">The <see cref="Camera079"/> to check.</param>
+        /// <returns>The <see cref="CameraType"/>.</returns>
+        public static CameraType Type(this Camera079 camera) => OrderedCameraTypes.TryGetValue(camera.GetInstanceID(), out CameraType cameraType) ? cameraType : CameraType.Unknown;
+
+        /// <summary>
+        /// Returns the <see cref="ZoneType"/> the camera is in.
+        /// </summary>
+        /// <param name="camera">The <see cref="Camera079"/> to check.</param>
+        /// <returns>The <see cref="ZoneType"/> of the zone the camera is in.</returns>
+        public static ZoneType GetZone(this Camera079 camera) => GetRoom(camera).Zone;
```
iRebbok, line 43: The GetRoom method may return null, make it null safety

iRebbok (changes requested):

iRebbok (approved):

PR #347, opened 2020-12-11
Fixes of the ChangingCameraStatus Event Docs

TheVoidNebula:
Yeah, some documentation was wrong ^^


TheVoidNebula:
btw because I have already a pull request open I created another branche (don't wonder why I deleted my changes in EffectType and Player class) ^^

PR #350, opened 2020-12-12
Adding Server.IpAddress

Thundermaker300:
yeah


iopietro:
`Server::IpAddress` would be better, to reflect `Player::IpAddress`.

iRebbok (approved):

PR #351, opened 2020-12-13
Add 106 to entering pocket dimension

PintTheDragon:
It adds the player who put the other player in the pocket dimension to the event


PintTheDragon:
Don't make any comments related to how the first commit says 1/3 instead of 1/2.

Exiled.Events/EventArgs/EnteringPocketDimensionEventArgs.cs
```diff
@@ -36,6 +38,11 @@ public EnteringPocketDimensionEventArgs(Player player, Vector3 position, bool is
         /// </summary>
         public Player Player { get; }
 
+        /// <summary>
+        /// Gets the Scp106 who sent the player to the pocket dimension.
```
Thundermaker300, line 42: the majority of API have the scps formatted as "SCP-XXX" in the documentation (not ScpXXX)

iRebbok (approved):

PR #352, opened 2020-12-14
Added comment about returned value of ICommand.Execute

TrickyBestia:


Exiled.Example/Commands/Test.cs
```diff
@@ -38,6 +38,7 @@ public bool Execute(ArraySegment<string> arguments, ICommandSender sender, out s
             else
                 response = "The command has been sent from the server console!";
 
+            // Return true if command executed successfully, otherwise return false.
```
iRebbok, line 41:
```suggestion
            // Return true if the command was executed successfully; otherwise, false.
```

iRebbok (approved):

PR #354, opened 2020-12-15
Adding Respawn APIs

Thundermaker300:
* `Respawn::TimeUntilNext` (int) - Returns the amount of time until the next respawn wave will spawn.
* `Respawn::CurrentlySpawning` (bool) - Returns whether or not the current respawn phase is spawning team or playing animations.

### DIFF
--- a/Exiled.API/Features/Respawn.cs
+++ b/Exiled.API/Features/Respawn.cs
@@ -33,7 +33,7 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets a value indicating whether or not a team is currently being spawned or the animations are playing for a team.
         /// </summary>
-        public static bool CurrentlySpawning => RespawnManager.Singleton._curSequence == RespawnManager.RespawnSequencePhase.PlayingEntryAnimations || RespawnManager.Singleton._curSequence == RespawnManager.RespawnSequencePhase.SpawningSelectedTeam;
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Respawn.cs
```diff
@@ -23,6 +25,16 @@ public static class Respawn
...
 
+        /// <summary>
+        /// Gets the amount of seconds before the next respawn will occur.
+        /// </summary>
+        public static int TimeUntilRespawn => Mathf.RoundToInt(RespawnManager.Singleton._timeForNextSequence - (float)RespawnManager.Singleton._stopwatch.Elapsed.TotalSeconds);
+
+        /// <summary>
+        /// Gets a value indicating whether or not a team is currently being spawned or the animations are playing for a team.
+        /// </summary>
+        public static bool CurrentlySpawning => RespawnManager.Singleton._curSequence == RespawnManager.RespawnSequencePhase.PlayingEntryAnimations || RespawnManager.Singleton._curSequence == RespawnManager.RespawnSequencePhase.SpawningSelectedTeam;
```
iRebbok, line 36: It'd better to rename the variable to `IsSpawning`

iRebbok (approved):

Exiled.API/Features/Respawn.cs
Thundermaker300: True, I didn't think about this. I won't be at my pc for a bit if you'd like to change it.

iRebbok (approved):

PR #355, opened 2020-12-15
Convert mute status changing events to IL patches.

steven4547466:
Converts `ChangingIntercomMuteStatus` and `ChangingMuteStatus` to IL patches. Their functionality remains the same.

### DIFF
--- a/Exiled.Events/Patches/Events/Player/ChangingIntercomMuteStatus.cs
+++ b/Exiled.Events/Patches/Events/Player/ChangingIntercomMuteStatus.cs
@@ -35,20 +35,14 @@ namespace Exiled.Events.Patches.Events.Player
             // Define the return label.
             var returnLabel = generator.DefineLabel();
 
-            // Define the continue label.
+            // Define the continue label and add it.
             // Used if IsAllowed is true.
             var continueLabel = generator.DefineLabel();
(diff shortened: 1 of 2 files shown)


iRebbok (changes requested):
Use the new Harmony API like `WithLabels` and etc

steven4547466 (commented):
Changed in latest commit

iopietro:
Is this useful? Does it improve maintainability?
Transpilers are only useful when injecting code inside other portions of codes, to prevent copy-pasting things that you'll have to update everytime base-game devs change them.

steven4547466:
Transpilers are more efficient than prefixes.

Thundermaker300:
I feel as though prefixes are better for legibility and can be maintained much easier (there's more of us that can fix a prefix patch as opposed to a transpiler). However, in the case of what iopietro said, if northwood changes something often and we need to consistently keep up with it, transpilers are pretty neat

iRebbok:
As the setters we patch don't contain any extra code that we might move/remove if it's useless, we don't need to use transpilers.
Transpilers are good when we want to patch a method in the middle (primary cause) or change the behavior of the method (second cause), they are also useful to patch at the end of the method, again, when we need to change the behavior.

In this case, transpilers make it difficult to maintain the code in the future.

PR #356, opened 2020-12-16
Added Player::SessionProperties

TrickyBestia:


Thundermaker300:
Would we want methods for this (eg. `Ply.AddCustomData(string, object)`, `Ply.GetCustomData(string, object)`, etc)? I've seen some people mention it

TrickyBestia:
@Thundermaker300 `Dictionary` provides us the same methods

Thundermaker300:
Yeah I know, wasn't sure if we wanted to give direct access to the dictionary or provide methods that do it

Exiled.API/Features/Player.cs
```diff
@@ -242,6 +242,11 @@ public string CustomPlayerInfo
             set => ReferenceHub.nicknameSync.Network_customPlayerInfoString = value;
         }
 
+        /// <summary>
+        /// Gets the dictionary of player's session properties. It is not being saved on player disconnect.
+        /// </summary>
+        public Dictionary<string, object> SessionProperties { get; } = new Dictionary<string, object>();
```
iRebbok, line 248:
```suggestion
        /// <summary>
        /// Gets the dictionary of player's session variables. It is not being saved on player disconnect.
        /// </summary>
        public Dictionary<string, object> SessionVariables { get; } = new Dictionary<string, object>();
```

iRebbok:
Can't say that this is very very useful, but the plugin devs may use this to tell a plugin that the player is (for example) SCP-035 or something.

PR #364, opened 2020-12-22
Possible NRE fix in Ghost Mode patch

steven4547466:
This should fix a null reference exception being thrown if the player is null.


Exiled.Events/Patches/Generic/GhostMode.cs
```diff
@@ -73,6 +73,9 @@ private static bool Prefix(PlayerPositionManager __instance)
                 foreach (GameObject gameObject in players)
                 {
                     Player player = GetPlayerOrServer(gameObject);
+                    if (player == null)
```
iRebbok, line 76: When and why the variable is being null? It will never happen or accepting this PR would hide the actual issue.

steven4547466: Well, the method `GetPlayerOrServer` can return null, why make it return null if you have no null check in the first place?

iRebbok: If this is related to your plugin you recently mentioned, so keep in mind that we remove the player when they left: https://github.com/galaxy119/EXILED/blob/d57bbdbc71661e43008bf33d3d63eaceb1f5a4f6/Exiled.Events/Patches/Events/Player/Left.cs#L47-L49
In this context, it means that you're forced to return the player object to the cache or remove they from the `PlayerManager.players` list.

steven4547466: While it does happen on my plugin (which is why I patched over it), I have seen it happen to other people as well. It is a plugin related thing, but it would make sense to have a null check if the method that gets the player has one that makes it explicitly return null as well. In my opinion, anyways.

PR #366, opened 2020-12-22
Right file, added Recontain079Status

TypicalIllusion:


Thundermaker300:
Why are you adding the enum if you don't use it anywhere??

PR #376, opened 2020-12-25
Added new Map::SpawnRagdoll methods

RogerFK:
# Changelog

- `Map::SpawnRagdoll(Player victim, DamageTypes.DamageType deathCause, Vector3 position, Quaternion rotation = default, Vector3 velocity = default, bool allowRecall = true)`

Spawns a ragdoll for a player (useful if: 1. Ragdolls aren't really spawned after a round restart; 2. You want SCP-049 to be able to respawn them.)

- `Map::SpawnRagdoll( RoleType roleType, DamageTypes.DamageType deathCause, string victimNick, Vector3 position, Quaternion rotation = default, Vector3 velocity = default, bool allowRecall = false, int playerId = -1, string mirrorOwnerId = null, float damage = -1f, string killerName = "Killer", int killerId = -1)`

Spawns a ragdoll based on a series of parameters. Most of them aren't actually used, but some other plugins might use them, so I felt like it was a good idea to keeping them there. The `HitInfo` used by the ragdoll's death cause is split into different parameters, so you don't have to manually do `new PlayerStats.HitInfo` *if you don't want to*.

- `Map::SpawnRagdoll(RoleType roleType, string victimNick, Vector3 position, Quaternion rotation = default, Vector3 velocity = default, global::PlayerStats.HitInfo hitInfo = default, bool allowRecall = false, int playerId = -1, string mirrorOwnerId = null)`

Another big boy method where half of the parameters don't matter, but the `HitInfo` is squashed into one variable instead of split into different optional parameters like the overload above.

- `Map::SpawnRagdoll(global::Role role, Ragdoll.Info ragdollInfo, Vector3 position, Quaternion rotation = default, Vector3 velocity = default, bool allowRecall = false)`

Uses the main game's parameters so you can re-use pre-loaded objects: every other method uses the logic in this one, this is its main purpose. Perfect if EXILED ever needs to change the code, or if someone wants to implement their own SpawnRagdoll which could call this one so, for example, every ragdoll has the same settings.

## Remarks
I haven't been able to *really* test the new methods and they are pretty weird to work with since the overloads are slightly messed up. Either way, the only "problem" I'd see would be the `mirrorOwnerID` field.

My suggestion, if ragdolls stop spawning through this method after a round restart, is to just stop using the `DefaultRagdollInfo` property and then just do the `GetComponent` every time this command gets called: I just don't know if this is necessary. Either way, you could always spawn a ragdoll for a player using `SpawnRagdoll(Player, DamageTypes.DamageType, Vector3)`, and this is 99% guaranteed to work.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -256,7 +256,52 @@ namespace Exiled.API.Features
         /// Spawns a ragdoll on the map based on the different arguments.
         /// </summary>
         /// <remarks>
-        /// Tip: You can do '<paramref name="ownerNick"/>: "MyNick"' to skip parameters.
+        /// Tip: You can do '<paramref name="allowRecall"/>: true, <paramref name="playerId"/>: MyPlayer.Id' to skip parameters.
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// // Code to spawn a fake ragdoll
+        /// if (ev.Player == MyPlugin.TheInmortalPlayer)
+        /// {
+        ///     var fakeRagdoll = Map.SpawnRagdoll(ev.Player.Role, ev.Player.Position, deathCause:DamageTypes.Fall, victimNick: ev.Player.DisplayNickname, playerId: ev.Player.Id);
+        /// }
+        /// </code>
+        /// </example>
+        /// <param name="roleType">The <see cref="RoleType"/> to use as ragdoll.</param>
+        /// <param name="deathCause">The death cause, expressed as a <see cref="DamageTypes.DamageType"/>.</param>
+        /// <param name="victimNick">The name from the victim, who the corpse belongs to.</param>
+        /// <param name="position">Where the ragdoll will be spawned.</param>
+        /// <param name="rotation">The rotation for the ragdoll.</param>
+        /// <param name="velocity">The initial velocity the ragdoll will have, as if it was exploded.</param>
+        /// <param name="allowRecall">Sets this ragdoll as respawnable by SCP-049. Must have a valid <paramref name="playerId"/>.</param>
+        /// <param name="playerId">Used for recall. The <see cref="Player.Id"/> to be recalled.</param>
+        /// <param name="mirrorOwnerId">The <see cref="Dissonance.Integrations.MirrorIgnorance.MirrorIgnorancePlayer"/>'s PlayerId field, likely used in the client.</param>
+        /// <param name="damage">Usually ignored, the killing blow damage.</param>
+        /// <param name="killerName">Usually ignored, the killer's name.</param>
+        /// <param name="killerId">Usually ignored, the killer's ID.</param>
+        /// <returns>The Ragdoll component (requires Assembly-CSharp to be referenced).</returns>
+        public static global::Ragdoll SpawnRagdoll(
+                RoleType roleType,
+                DamageTypes.DamageType deathCause,
+                string victimNick,
+                Vector3 position,
+                Quaternion rotation = default,
+                Vector3 velocity = default,
+                bool allowRecall = false,
+                int playerId = -1,
+                string mirrorOwnerId = null,
+                float damage = -1f,
+                string killerName = "Killer",
+                int killerId = -1)
+        {
+            return SpawnRagdoll(roleType, victimNick, position, rotation, velocity, new PlayerStats.HitInfo(damage, killerName, deathCause == default ? DamageTypes.Com15 : deathCause, killerId), allowRecall, playerId, mirrorOwnerId);
+        }
+
+        /// <summary>
+        /// Spawns a ragdoll on the map based on the different arguments.
+        /// </summary>
+        /// <remarks>
+        /// Tip: You can do, for example, '<paramref name="velocity"/>: "Vector3.up * 3"' to skip parameters.
         /// </remarks>
         /// <example>
         /// <code>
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Map.cs
```diff
@@ -31,6 +31,18 @@ public static class Map
         private static readonly List<Camera079> CamerasValue = new List<Camera079>(250);
         private static readonly List<Lift> LiftsValue = new List<Lift>(10);
         private static readonly List<TeslaGate> TeslasValue = new List<TeslaGate>(10);
+        private static readonly Lazy<Ragdoll.Info> LazyDefaultRagdollOwner = new Lazy<Ragdoll.Info>(() =>
```
iRebbok, line 34: You don't need the lazy instancing, assign the value to the property directly

Exiled.API/Features/Map.cs
```diff
@@ -31,6 +31,18 @@ public static class Map
...
+        {
+            return new Ragdoll.Info()
+            {
+                ownerHLAPI_id = PlayerManager.localPlayer.GetComponent<Dissonance.Integrations.MirrorIgnorance.MirrorIgnorancePlayer>().PlayerId,
```
iRebbok, line 38: What will cause null/empty string as a value?

Exiled.API/Features/Map.cs
```diff
@@ -229,6 +253,189 @@ public static Room FindParentRoom(GameObject objectInRoom)
...
+        /// <param name="rotation">The rotation for the ragdoll.</param>
+        /// <param name="velocity">The initial velocity the ragdoll will have, as if it was exploded.</param>
+        /// <param name="allowRecall">Sets this ragdoll as respawnable by SCP-049. Must have a valid <paramref name="playerId"/>.</param>
+        /// <param name="playerId">Used for recall. The <see cref="Player.Id"/> to be recalled.</param>
+        /// <param name="mirrorOwnerId">The <see cref="Dissonance.Integrations.MirrorIgnorance.MirrorIgnorancePlayer"/>'s PlayerId field, likely used in the client.</param>
+        /// <param name="damage">Usually ignored, the killing blow damage.</param>
+        /// <param name="killerName">Usually ignored, the killer's name.</param>
+        /// <param name="killerId">Usually ignored, the killer's ID.</param>
+        /// <returns>The Ragdoll component (requires Assembly-CSharp to be referenced).</returns>
+        public static global::Ragdoll SpawnRagdoll(
```
iRebbok, line 308: What's point of this method? It has 11 arguments and 4 of which are just unpacked arguments from `PlayerStats.HitInfo`. If the game devs will decide to add more than 4 arguments for the `PlayerStats.HitInfo`'s constructor, it'll break this method.

Exiled.API/Features/Map.cs
```diff
@@ -229,6 +253,189 @@ public static Room FindParentRoom(GameObject objectInRoom)
...
+                Vector3 position,
+                Quaternion rotation = default,
+                Vector3 velocity = default,
+                bool allowRecall = false)
+        {
+            if (role.model_ragdoll == null)
+                return null;
+
+            GameObject gameObject = Object.Instantiate(role.model_ragdoll, position + role.ragdoll_offset.position, Quaternion.Euler(rotation.eulerAngles + role.ragdoll_offset.rotation));
+            Mirror.NetworkServer.Spawn(gameObject);
```
iRebbok, line 431: Spawn the gameobject after you've assigned variables, don't serialize the variables twice as you call this method optimized

iRebbok (changes requested):

Exiled.API/Features/Map.cs
RogerFK: You're right, but I was thinking about Galaxy's own words of "EXILED doesn't bring any bloat unless a plugin requires it" and stuff. I know it's just a few bytes, but this implementation followed that mindset

Exiled.API/Features/Map.cs
RogerFK: I couldn't test it, and I'm also wondering what will happen. Either reject it because it requires more testing or just say this is almost absolutely untested and might break between rounds; either way the overload Map::SpawnRagdoll that uses the ev.Player should work flawlessly.

Exiled.API/Features/Map.cs
RogerFK: Right, didn't consider that detail, maybe I should have used the `new PlayerStats.HitInfo() {}` method so it's forwards-compatible *(if that word even exist)*.

Either way, the point of this method is to avoid having to set the damage and other unused stuff, so you don't have to create the entire object with most of its members ignored anyways. I could have perfectly left the "deathCause" value by itself and it would have had the same purpose.

Exiled.API/Features/Map.cs
RogerFK: An absolute scam on my part, lol. I'll fix everything when I find more time. I only checked this because I was showing a friend that my code is actually just comments, but thanks for the tip, makes sense too.

Exiled.API/Features/Map.cs
iRebbok: Yep, he says so, but it's because we don't create a new Player instance every time we call an event.

Exiled.API/Features/Map.cs
iRebbok: You already made an overload which requires a Player instance, if someone wants spawn a "fully controlled" ragdoll, so they forced to use `SpawnRagdoll(RoleType, string, Vector3, Quaternion, Vector3, PlayerStats.HitInfo, bool, int, string)`, another way looks so useless as a middle level.

Exiled.API/Features/Map.cs
iRebbok: Will test it for you.

iRebbok: Ok, the test was successful. Nothing happens if I use null/invalid id.

RogerFK: Marking it as resolved, many thanks!

Exiled.API/Features/Map.cs
```diff
@@ -229,6 +247,188 @@ public static Room FindParentRoom(GameObject objectInRoom)
...
+                return null;
+            var @default = DefaultRagdollOwner;
+            /* // Uncomment if people complain that changing
+             * // DefaultRagdollOwner.DeathCause changes
+             * // already spawned ragdolls.
+            if (hitInfo == default)
+            {
+                hitInfo = new PlayerStats.HitInfo(hitInfo.Amount, hitInfo.Attacker, DamageTypes.FromWeaponId(hitInfo.Tool), hitInfo.PlayerId);
+            }
+            */
```
iRebbok, line 363:
```suggestion
```

Remove the code and people won't even think it exists 😎

iRebbok (approved):

PR #378, opened 2020-12-28
Configurable Velocity on SpawnRagdoll

TrueAbastien:
This seems like the only efficient way to modify a Ragdoll velocity since HarmonyPatch cannot be stacked.

### DIFF
--- a/Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
+++ b/Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
@@ -24,18 +24,18 @@ namespace Exiled.Events.Patches.Events.Player
     [HarmonyPatch(typeof(RagdollManager), nameof(RagdollManager.SpawnRagdoll))]
     internal static class SpawningRagdoll
     {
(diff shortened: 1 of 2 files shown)


Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
```diff
@@ -24,17 +24,18 @@ namespace Exiled.Events.Patches.Events.Player
     [HarmonyPatch(typeof(RagdollManager), nameof(RagdollManager.SpawnRagdoll))]
     internal static class SpawningRagdoll
     {
-        private static bool Prefix(RagdollManager __instance, ref Vector3 pos, ref Quaternion rot, ref int classId, ref PlayerStats.HitInfo ragdollInfo, ref bool allowRecall, ref string ownerID, ref string ownerNick, ref int playerId)
+        private static bool Prefix(RagdollManager __instance, ref Vector3 pos, ref Quaternion rot, ref Vector3 vel, ref int classId, ref PlayerStats.HitInfo ragdollInfo, ref bool allowRecall, ref string ownerID, ref string ownerNick, ref int playerId)
```
iRebbok, line 27: The variable name in the patch must match the variable name in the original method.

iRebbok (changes requested):

Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
TrueAbastien: Thanks, I did actually forgot that since the position and rotation are shortened.

iRebbok (approved):

PR #384, opened 2021-01-07
Removing those .Where

xRoier:
Yes.


galaxy119:
No

xRoier:
Well, I'm such a pepega.

iRebbok:
👀

iopietro:
It would only make sense if it returns a `List<Player>`, which actually doesn't.

PR #386, opened 2021-01-08
Proper call order for common events

iRebbok:
### New behavior
Joined - called once the player's object has been instantiated
Verified - called once the player has been verified
Destroying - called before the player's object is destroyed
Left - called before the player has left the server

Joined event is called right before Verified event is called, after them ChangingGroup event is called.

This was made to fix the current situation where Left event isn't called and during a silent round restart a plugin doesn't know which player is still on the server.


iRebbok:
`Player::IpAddress` no longer has a setter as well as `Player::UserId`.

PR #387, opened 2021-01-10
Changed "smallest" to "latest"

AerGameChannel:


iRebbok:
That line doesn't mean that it's the latest version of exiled, it means that the smallest version was found, i.e if there are two versions 2.1.17 and 2.1.18 so it'll return 2.1.17.

PR #390, opened 2021-01-15
Fixed door type stuff

Thundermaker300:
let me know if I broke something, first time committing to Exiled in a bit


iRebbok:
Will have a deal with this.

PR #391, opened 2021-01-19
Added new method for Player, "HasItem"

KognitysPlayhouse:
Does exactly as it sounds. Searches the players inventory for a given ItemType and if found returns true.


KognitysPlayhouse:
I did a boo boo

PR #392, opened 2021-01-19
Added new method for Player, "HasItem".

KognitysPlayhouse:
Does exactly what it sounds like, searches the player's inventory and if the given item is found returns true. Otherwise false.

(but it actually works this time)

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1073,10 +1073,10 @@ namespace Exiled.API.Features
         }
 
         /// <summary>
-        /// Returns true if item is found otherwise returns false.
+        /// Indicates whether or not the player has an item.
         /// </summary>
         /// <param name="targetItem">The item to search for.</param>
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Player.cs
```diff
@@ -1072,6 +1072,24 @@ public void DropItem(Inventory.SyncItemInfo item)
             Inventory.items.Remove(item);
         }
 
+        /// <summary>
+        /// Returns true if item is found otherwise returns false.
+        /// </summary>
+        /// <param name="targetItem">The item to search for.</param>
+        /// <returns>Returns true or false.</returns>
```
iRebbok, line 1079:
```suggestion
        /// Indicates whether or not the player has an item.
        /// </summary>
        /// <param name="targetItem">The item to search for.</param>
        /// <returns>true, if the player has it; otherwise, false</returns>
```

iRebbok (approved):

PR #398, opened 2021-02-07
Added Map.ChangeUnitColor()

Jesus-QC:
Credits to sanyae

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -532,11 +532,11 @@ namespace Exiled.API.Features
                 if (Unit.Contains("<color"))
                 {
                     System.Text.RegularExpressions.Regex regex = new System.Text.RegularExpressions.Regex("<color=(.*)>(.*)</color>");
-                    var v = regex.Match(Unit);
+                    var v = regex.Match(unit);
                     string s = v.Groups[1].ToString();
-                    ply.ReferenceHub.characterClassManager.NetworkCurUnitName = Unit.Replace(s, color);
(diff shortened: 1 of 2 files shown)


iRebbok:
Looks a little strange. What if a dev wants to change the color twice? Does it even have to be in the API?

Jesus-QC:
Well.. Amathor asked me to put this in here but you are right, I will have to check if the unit has already changed the color

Jesus-QC:
greeny green

iRebbok:
@galaxy119 Your move

galaxy119:
Looks solid to me, this has been possible in the game for several versions (since MP1 iirc). I'm fairly confident this won't be something removed from the game.

Exiled.API/Features/Map.cs
```diff
@@ -514,6 +515,34 @@ public static DoorVariant GetDoorByName(string doorName)
...
+                {
+                    System.Text.RegularExpressions.Regex regex = new System.Text.RegularExpressions.Regex("<color=(.*)>(.*)</color>");
+                    var v = regex.Match(unit);
+                    string s = v.Groups[1].ToString();
+                    ply.ReferenceHub.characterClassManager.NetworkCurUnitName = unit.Replace(s, color);
+                }
+                else
+                {
+                    ply.ReferenceHub.characterClassManager.NetworkCurUnitName = $"<color={color}>{unit}</color>";
+                }
```
iRebbok, line 542: Simplify the regex to `<[^>]*?>` and call `Regex::Replace(string, string, string)` instead to remove tags (https://regex101.com/r/9Wsd6l/1)

iRebbok (changes requested):

PR #401, opened 2021-02-12
Medical events mess

iRebbok:
Isn't actually a bug, but will be merged into dev as well as dev-enhancement soon.


PintTheDragon:
Could you explain what exactly what dequipt means? For example, if a player uses pills, when is this event called?

iRebbok:
> Could you explain what exactly what dequipt means? For example, if a player uses pills, when is this event called?

The `DequippedMedicalItem` event is invoked when a player starts dequipping a medical item, before health begins to increase, but the `UsedMedicalItem` event is invoked after a player gains a health increase. For example, the `DequippedMedicalItem` event fires before the `UsedMedicalItem` event does, and `UsedMedicalItem` might not fire if the player's class has changed.
This PR adds the `DequippedMedicalItem` event and displaces the `UsedMedicalItem` one, and also moves down the `UsedMedicalItem` event to be invoked after `DequippedMedicalItem` for consistency. It isn't a breaking change, nothing is removed, just some kind of displacement.

PR #405, opened 2021-02-16
Add custom item to pickup events

PintTheDragon:
This will allow you to more easily interact with custom items in drop and pickup event, from outside of the custom item itself.


PintTheDragon:
Of course Events can't access CustomItems.

PR #408, opened 2021-02-18
Merging fix for Player object being accessible when RH is incomplete.

galaxy119:
Player.Joined shouldn't be used for anything Player object related anyways due to the player's UserId and various other things being null during this event. 
As such, we also shouldn't initialize this object at this time, as it could cause foreach loops over Player.List to throw unexpected nullrefs if the iteration happens to go over a player who has 'joined' but is not yet verified.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -85,11 +85,6 @@ namespace Exiled.API.Features
         public static Dictionary<int, Player> IdsCache { get; } = new Dictionary<int, Player>(20);
 
         /// <summary>
-        /// A dictionary for storing player objects of connected but not yet verified players.
-        /// </summary>
-        internal static Dictionary<ReferenceHub, Player> UnverifiedPlayers = new Dictionary<ReferenceHub, Player>();
-
-        /// <summary>
         /// Gets the encapsulated <see cref="ReferenceHub"/>.
         /// </summary>
         public ReferenceHub ReferenceHub
@@ -848,6 +843,11 @@ namespace Exiled.API.Features
         /// </summary>
         [Obsolete("Removed from the base-game.", true)]
         public bool HasHands => false;
+
+        /// <summary>
+        /// Gets a dictionary for storing player objects of connected but not yet verified players.
+        /// </summary>
+        internal static Dictionary<ReferenceHub, Player> UnverifiedPlayers { get; } = new Dictionary<ReferenceHub, Player>();
 
         /// <summary>
         /// Gets a <see cref="Player"/> <see cref="IEnumerable{T}"/> filtered by team.
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -96,28 +96,34 @@ namespace Exiled.API.Features
                     throw new NullReferenceException("Player's ReferenceHub cannot be null!");
 
                 referenceHub = value;
+                GameObject = value.gameObject;
+                Ammo = value.ammoBox;
+                HintDisplay = value.hints;
+                Inventory = value.inventory;
+                CameraTransform = value.PlayerCameraReference;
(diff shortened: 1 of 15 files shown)


galaxy119:
Leaving as draft until the changes have been tested for un-intentional side effects.

Exiled.API/Features/Player.cs
```diff
@@ -845,6 +835,11 @@ public int Scp330Usages
         [Obsolete("Removed from the base-game.", true)]
         public bool HasHands => false;
 
+        /// <summary>
+        /// Gets a dictionary for storing player objects of connected but not yet verified players.
+        /// </summary>
+        internal static ConditionalWeakTable<ReferenceHub, Player> UnverifiedPlayers { get; } = new ConditionalWeakTable<ReferenceHub, Player>();
```
iRebbok, line 841: Set an initial capacity as we do with the dictionaries above

Exiled.Installer/Exiled.Installer.csproj
```diff
@@ -16,7 +16,7 @@
     <OutputPath>bin\$(Configuration)\</OutputPath>
     
     <AssemblyTitle>Exiled online installer</AssemblyTitle>
-    <Version>3.2.0</Version>
+    <Version>3.3.0</Version>
```
iRebbok, line 19: Why did you bump minor version for a fix?

Exiled.Installer/Program.cs
```diff
@@ -298,16 +298,14 @@ private static bool TryFindRelease(CommandSettings args, IEnumerable<Release> re
...
 
-            var range = args.TargetVersion != null ? new Range($"={args.TargetVersion}") : new Range(">2.0.0");
-
             foreach (var r in releases)
             {
                 release = r;
 
-                if (!range.IsSatisfied(r.TagName))
-                    continue;
+                if (args.TargetVersion != null && (new Version(args.TargetVersion) == new Version(r.TagName)))
```
iRebbok, line 305: Wouldn't it be better to compare strings instead? Anyway, don't create a new instance for `args.TargetVersion`, it calls regex.

iRebbok (changes requested):

Exiled.Installer/Exiled.Installer.csproj
galaxy119: because I forgot to bump the version when I did my initial changes before releasing 2.2.0, Idk why I bumped minor instead of patch though, I will fix that.

Exiled.Installer/Program.cs
galaxy119: Idk, it might be but I know this way works where range.IsSatisfied didn't. I'll fix the extra regex stuff though.

Exiled.API/Features/Player.cs
galaxy119: There is no initial capacity argument for a ConditionalWeakTable

Exiled.Events/Patches/Generic/GhostMode.cs
```diff
@@ -232,7 +241,7 @@ private static bool Prefix(PlayerPositionManager __instance)
             }
             catch (Exception exception)
             {
-                Log.Error($"GhostMode error: {exception}");
+                Log.Error($"GhostMode error: {exception}\n{exception.InnerException}\n{exception.StackTrace}");
```
iRebbok, line 244: Why? You literally print the same exception twice!1 Undo your change.

iRebbok (changes requested):

Exiled.Events/Patches/Generic/GhostMode.cs
galaxy119: Oops, leftover from testing something

iRebbok (approved):
LGTM!

PR #412, opened 2021-02-21
Removing those .Where()

xRoier:
Now yes.


iopietro:
It's actually slower with less than 100k objects: https://dotnetfiddle.net/A6UTxD.

iRebbok:
It does the same, all IEnumerable are compiled into classes, every call creates an object, so it makes no sense to insert new lines of code.

xRoier:
Oh ok, I didn't test the performance but I thought LiNQ was slower anyway.

PR #414, opened 2021-02-21
Fixed CustomInfo color tag

ArseeniK:


galaxy119:
Don't make PR's with mixed features. custom-items is only for custom-items.

PR #420, opened 2021-02-22
Remove the Both mode from InfoSide

iRebbok:
This image describes the reason


iopietro:
This would break configs on every server that has already dowloaded the latest version of Exiled.
We shouldn't remove `Both` tho, but only set the default value of `Mode` to `Badge`.

iopietro (approved):

galaxy119 (approved):
Yea ok this is fine.

PR #421, opened 2021-02-23
Exiled.CustomItems changes

iopietro:
+ Changed all commands to use a prefix, to lower chances to conflict with other plugins without a prefix or using global commands.
+ Moved things around to make the project structure more like other Exiled ones.
+ Renamed and moved methods to fit them more and call them more fluently.
+ Clean up.


galaxy119 (approved):
Looks fine to me, tested it and custom items still work.

BuildBoy12 (approved):
Looks clean to me.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -347,14 +334,18 @@ protected virtual void OnEscaping(EscapingEventArgs ev)
...
+        protected virtual void ShowMessage(Player player)
+        {
+            player.ShowHint(string.Format(Instance.Config.PickedUpBroadcast.Content, Name, Description), Instance.Config.PickedUpBroadcast.Duration);
+        }
 
         /// <summary>
         /// Called when a player is given the item directly via a command or plugin.
         /// </summary>
         /// <param name="player">The <see cref="Player"/> who received the item.</param>
         protected virtual void ItemGiven(Player player)
```
iRebbok, line 346: If you present it as an event, it'd be better to start the name with `On`, like `OnItemGiven`.

iRebbok (approved):

Exiled.CustomItems/API/Features/CustomItem.cs
galaxy119: That's what I'm going to do. Or pietro will, either way.

iopietro: Already removed
Troll

galaxy119 (approved):
LGTM

BuildBoy12 (approved):
Looks fine to me.

iRebbok (approved):

Exiled.CustomItems/API/Features/CustomGrenade.cs
```diff
@@ -156,9 +156,6 @@ private void OnInternalThrowing(ThrowingGrenadeEventArgs ev)
 
             OnThrowing(ev);
 
-            if (!ev.IsAllowed)
```
galaxy119: Why remove this? It could be useful for items that want to handle the throwing on their own, or want to just not spawn a grenade that's thrown.

iopietro: Because it would never allow to throw grenades, since it's set to false 3 lines above.

galaxy119: Oh good point. Should we add a way to have the built in OnThrowing() event tell the CustomGrenade OnThrowing event not to continue, then?

For example with the Custom Item "Rock", I don't want the CustomGrenade OnInternalThrowing() to continue after it calls the virtual OnThrowing method. The only way to do that currently is override SubscribeEvents to not have it subscribe to InternalThrowing.

iopietro: Yeah, we could.

PR #422, opened 2021-02-23
Add Player::Get(Side) overload

iRebbok:


Exiled.API/Features/Player.cs
```diff
@@ -840,6 +840,13 @@ public int Scp330Usages
         /// </summary>
         internal static ConditionalWeakTable<ReferenceHub, Player> UnverifiedPlayers { get; } = new ConditionalWeakTable<ReferenceHub, Player>();
 
+        /// <summary>
+        /// Gets a <see cref="Player"/> <see cref="IEnumerable{T}"/> filtered by side.
+        /// </summary>
+        /// <param name="side">The player's side.</param>
```
BuildBoy12, line 846: Change player's to players' for consistency with other IEnumerable Gets.

BuildBoy12 (approved):

Exiled.API/Features/Player.cs
iRebbok: Sure, I didn't notice it.

BuildBoy12 (approved):

Exiled.API/Features/Player.cs
```diff
@@ -840,6 +840,13 @@ public int Scp330Usages
         /// </summary>
         internal static ConditionalWeakTable<ReferenceHub, Player> UnverifiedPlayers { get; } = new ConditionalWeakTable<ReferenceHub, Player>();
 
+        /// <summary>
+        /// Gets a <see cref="Player"/> <see cref="IEnumerable{T}"/> filtered by side.
+        /// </summary>
+        /// <param name="side">The players' side.</param>
+        /// <returns>Returns the filtered <see cref="IEnumerable{T}"/>.</returns>
+        public static IEnumerable<Player> Get(Side side) => List.Where(player => player.Side == side);
```
galaxy119, line 848: .Where() :vomiting_face:

galaxy119 (approved):

Exiled.API/Features/Player.cs
BuildBoy12: Isn't the performance impact of .Where versus the foreach variable, as shown by https://dotnetfiddle.net/pG4kKi?

iopietro: We're talking about differences of microseconds and the player list will always have less than 100 objects.
Also, that query will only be evaluated when iterated by a foreach, Count(), ToList() or manually.

iopietro (approved):

Exiled.API/Features/Player.cs
iRebbok: `Where(Predicate<T>)` doesn't go through all values and doesn't allocate a new array, it just returns a new instance of 'filter' with a predicate passed by a developer. So there's no reason to replace it with a custom 'filter' implementation.

PR #423, opened 2021-02-23
Add a null check for RA config

iRebbok:
This will let users know the exact reason of a NullReferenceException with the message `Unable to parse permission config: ...`.

### DIFF
--- a/Exiled.Permissions/Extensions/Permissions.cs
+++ b/Exiled.Permissions/Extensions/Permissions.cs
@@ -88,6 +88,14 @@ namespace Exiled.Permissions.Extensions
         /// </summary>
         public static void Reload()
         {
+            if (ServerStatic.PermissionsHandler == null)
+            {
+                Log.Error("Your Remote Admin config is broken. You have to fix it because the game won't even start with a broken config.");
+
+                // Then do nothing.
+                return;
+            }
+
             try
             {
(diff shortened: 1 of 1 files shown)


Exiled.Permissions/Extensions/Permissions.cs
```diff
@@ -88,6 +88,14 @@ public static void Create()
         /// </summary>
         public static void Reload()
         {
+            if (ServerStatic.PermissionsHandler == null)
+            {
+                Log.Error("Your Remote Admin config is broken. You have to fix it because the game won't even start with a broken config.");
+
+                // Then do nothing.
```
iopietro, line 95: Do we need this comment? Ahah

iopietro (approved):

Exiled.Permissions/Extensions/Permissions.cs
BuildBoy12: I mean, it says what it is lmao

BuildBoy12 (approved):

Exiled.Permissions/Extensions/Permissions.cs
iopietro: Should we add `throw new YouAreDumpException("You borked your Remote Admin configs, you monkey!");`?
Troll

iRebbok: I think everything should be fine now.

galaxy119 (approved):
I'd love a ``throw UserErrorException()`` tbh

PR #426, opened 2021-03-03
10.2.2

galaxy119:
Update to new game version

### DIFF
--- a/Exiled.Events/Patches/Events/Player/EnteringPocketDimension.cs
+++ b/Exiled.Events/Patches/Events/Player/EnteringPocketDimension.cs
@@ -49,9 +49,21 @@ namespace Exiled.Events.Patches.Events.Player
                     return false;
                 }
 
+                Vector3 position = ply.transform.position;
+                float num1 = Vector3.Distance(__instance._hub.playerMovementSync.RealModelPosition, position);
+                float num2 = Math.Abs(__instance._hub.playerMovementSync.RealModelPosition.y - position.y);
+                if ((num1 >= 1.8179999589920044 && num2 < 1.0199999809265137) || (num1 >= 2.0999999046325684 && num2 < 1.9500000476837158) || ((num1 >= 2.6500000953674316 && num2 < 2.200000047683716) || (num1 >= 3.200000047683716 && num2 < 3.0)) || num1 >= 3.640000104904175)
+                {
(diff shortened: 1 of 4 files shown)


galaxy119:
Ignore build failures on Main, reference zip won't be updated until SL updates

Exiled.Events/Patches/Events/Player/EnteringPocketDimension.cs
```diff
@@ -49,9 +49,21 @@ private static bool Prefix(Scp106PlayerScript __instance, GameObject ply, int t)
...
                 }
 
+                Vector3 position = ply.transform.position;
+                float num1 = Vector3.Distance(__instance._hub.playerMovementSync.RealModelPosition, position);
+                float num2 = Math.Abs(__instance._hub.playerMovementSync.RealModelPosition.y - position.y);
+                if ((num1 >= 1.8179999589920044 && num2 < 1.0199999809265137) || (num1 >= 2.0999999046325684 && num2 < 1.9500000476837158) || ((num1 >= 2.6500000953674316 && num2 < 2.200000047683716) || (num1 >= 3.200000047683716 && num2 < 3.0)) || num1 >= 3.640000104904175)
+                {
+                    __instance._hub.characterClassManager.TargetConsolePrint(__instance.connectionToClient, $"106 MovePlayer command rejected - too big distance (code: T1). Distance: {num1}, Y Diff: {num2}.", "gray");
+                }
+                else if (Physics.Linecast(__instance._hub.playerMovementSync.RealModelPosition, ply.transform.position, (int)__instance._hub.weaponManager.raycastServerMask))
```
BuildBoy12, line 59: This `ply.transform.position` could be replaced by `position`, other than that looks fine.

iopietro: I think it's copy pasted base-game code, but you're right.

galaxy119: It is, and yes, I could get both Player objects at the start and use them for all of these things, but i figured using base-game code for anything anti-cheat related is probably the better idea just so if something's broken we know 100% it isn't our fault

iopietro: Also, sooner or later, it will be moved into a transpiler.

iopietro (approved):

BuildBoy12 (approved):

Exiled.Events/Patches/Events/Player/EnteringPocketDimension.cs
BuildBoy12: Fair enough.

iRebbok (approved):

Issue #428, opened 2021-03-03
Problems with Map.Rooms caching


Nekonyx:
EXILED clears `Map.Rooms`'s cache on round restart, and if any of the plugins accesses `Map.Rooms` between (or during) `RestartingRound` and `WaitingForPlayers` (when a new scene is created and new rooms are generated), EXILED may cache the previous rooms.

PoC: https://gist.github.com/Nekonyx/661f0cb56782efe96e2696ea03ff3cb5
![image](https://user-images.githubusercontent.com/16444061/109813878-18cd3500-7c58-11eb-93ec-13f4959a66c3.png)

galaxy119:
Fixed in 2.7.0-rc.1

PR #429, opened 2021-03-03
Fix for #428

galaxy119:
By moving the ClearCache() call to when the map is generated, we guaruntee that the cache will always be cleared as close as possible to when the new map is generated, and should prevent issues pointed out in #428

### DIFF
--- a/Exiled.Events/Handlers/Internal/Round.cs
+++ b/Exiled.Events/Handlers/Internal/Round.cs
@@ -35,7 +35,6 @@ namespace Exiled.Events.Handlers.Internal
         {
             MultiAdminFeatures.CallEvent(MultiAdminFeatures.EventType.ROUND_END);
 
-            API.Features.Map.ClearCache();
             API.Features.Scp173.TurnedPlayers.Clear();
             API.Features.Scp096.TurnedPlayers.Clear();
         }
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -28,11 +28,30 @@ namespace Exiled.API.Features
     /// </summary>
     public static class Map
     {
-        private static readonly List<Room> RoomsValue = new List<Room>(250);
-        private static readonly List<DoorVariant> DoorsValue = new List<DoorVariant>(250);
-        private static readonly List<Camera079> CamerasValue = new List<Camera079>(250);
-        private static readonly List<Lift> LiftsValue = new List<Lift>(10);
-        private static readonly List<TeslaGate> TeslasValue = new List<TeslaGate>(10);
+        /// <summary>
+        /// A list of <see cref="Room"/>s on the map.
+        /// </summary>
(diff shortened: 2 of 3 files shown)


Exiled.Events/Handlers/Internal/Round.cs
```diff
@@ -35,7 +35,6 @@ public static void OnRestartingRound()
         {
             MultiAdminFeatures.CallEvent(MultiAdminFeatures.EventType.ROUND_END);
 
-            API.Features.Map.ClearCache();
```
iRebbok, line 38: Horever, there's a reason for this, the game calls `GC.Collect()` after executing this line, keeping the cache does nothing (I guess) but I'm sure we don't want to prevent objects from being collected. I think also clearing the cache on the MapGenerated event is better.

galaxy119: But clearing the cache too early can lead to nullrefs in plugins aswell, if they try to access something after RoundRestarting but before it's restarted, if the list is empty they could perhaps nullref (IE if someone has a Map.Rooms.FirstOrDefault() without a null check)

iopietro: > I don't think postponing the GC.Collect() call a few sec will make a huge difference.
`GC.Collect()` is manually called from the game when the round restarts, so we can't actually postpone that call, but the GC will collect those objects sooner or later anyway, based on their generation.
I don't think that would actually have a big impact on used memory.

iopietro (approved):

Exiled.Events/Handlers/Internal/Round.cs
iRebbok: Calling Map.Rooms after RoundRestarting seems to be incorrect because there are already no players on the server tbh

iopietro: > I'm still sure we should clear the cache after RoundRestarting & MapGenerated.

So, between RoundRestarting and MapGenerated?

iRebbok: Yep

iRebbok: No, after another implementation I think it shold be cleared during RoundRestarting, plugins shouldn't access the cache after RoumdRestarting, it makes no sence, the issue itself is so controversial.

galaxy119: Well one way or another, plugins were accessing Map.Rooms after the cache had been cleared, and therefore causing the rooms in that list to no longer match with actual rooms on the map after OnMapGenerated.

Moving the clearing of the cache and building those lists to 1 frame after the map is generated, solves any and all potential issues that could arise, without any noticable downsides. (yes, the objects in those lists won't be immediately collected by the GC but it's minor, once they are cleared from the lists, the next time GC collects, it'll purge them, so I don't really see the issue)

BuildBoy12 (approved):
LGTM

PR #433, opened 2021-03-04
Adding GetDefaultDurability()

Jesus-QC:
This will save some code to a few plugins.

### DIFF
--- a/Exiled.API/Extensions/Item.cs
+++ b/Exiled.API/Extensions/Item.cs
@@ -316,14 +316,7 @@ namespace Exiled.API.Extensions
         /// <returns>Returns the item durability.</returns>
         public static float GetDefaultDurability(ItemType item)
         {
-            try
-            {
-                return Pickup.Inv.GetItemByID(item)?.durability ?? float.NaN;
-            }
-            catch (Exception)
-            {
-                return float.NaN;
-            }
+            return Pickup.Inv.GetItemByID(item)?.durability ?? float.NaN;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Jesus-QC:
Fuck, this is master branch, sorry for that <3

iRebbok (changes requested):
It's actually better to get the values from the game directly, also they shouldn't be `int` either.

galaxy119:
They should be uint, and there's a base-game function for this somewhere, use that, because these values could change in the future.

Exiled.API/Extensions/Item.cs
```diff
@@ -308,5 +308,15 @@ public static void SetOther(this Player player, Inventory.SyncItemInfo weapon, O
...
         }
+
+        /// <summary>
+        /// Gets the default durability of any item.
+        /// </summary>
+        /// <param name="item">The <see cref="ItemType">item</see> that you want to get durability of.</param>
+        /// <returns>Returns the item durability.</returns>
+        public static float GetDefaultDurability(ItemType item)
+        {
+            return Pickup.Inv.GetItemByID(item).durability;
```
iRebbok, line 319: Make it not throw an exception when `Pickup.Inv.GetItemByID(item)` returned null

iRebbok (changes requested):

iRebbok (approved):

Exiled.API/Extensions/Item.cs
```diff
@@ -316,14 +316,7 @@ public static void SetOther(this Player player, Inventory.SyncItemInfo weapon, O
         /// <returns>Returns the item durability.</returns>
         public static float GetDefaultDurability(ItemType item)
```
BuildBoy12, line 317: Shouldn't it be `this ItemType item` so that it may be used as an extension?

iopietro: Yes, I agree.

BuildBoy12 (approved):

iopietro (approved):

PR #434, opened 2021-03-04
Copy over correct license from SPDX

miniduikboot:
Previous license was misdetected by license detection tools, so I reimported the license from the SPDX project:
https://github.com/spdx/license-list-data/blob/fa9e53f4475cb6cc64ec59d997b43c59d3720577/text/CC-BY-SA-3.0.txt

There ARE significant changes in this license, including but not limited to a few definitions, a few reworded sentences compared to the previous license which was sourced from the CC website. I'm not sure what's up with this revision.


galaxy119 (approved):
Looks fine to me, basically the same license just in a different formatting, to match the one listed on SPDX

PR #438, opened 2021-03-06
Fix for #436

galaxy119:
Allows for custom grenades that are currently in the world as a pickup to explode with their custom effects if they are exploded with a frag grenade.

Also adds an internal OnExploding virtual event for CustomGrenade that does the Check() for you.


BuildBoy12 (approved):
LGTM

iopietro (approved):

PR #439, opened 2021-03-07
Changes

iopietro:
`[Exiled.Events]`
+ Changed OnAddingTarget, OnUpgradingItems and OnGeneratorActivated into transpilers.
+ Some clean up.
+ Marked SendingConsoleCommandEventArgs::Allow as obsolete.
+ Added GeneratorActivatedEventArgs::IsAllowed.

`[Exiled.Example]`
+ Renamed Handlers folder into Events and its classes.
+ Added some events.

### DIFF
--- a/Exiled.Events/EventArgs/AddingTargetEventArgs.cs
+++ b/Exiled.Events/EventArgs/AddingTargetEventArgs.cs
@@ -19,16 +19,18 @@ namespace Exiled.Events.EventArgs
         /// <summary>
         /// Initializes a new instance of the <see cref="AddingTargetEventArgs"/> class.
         /// </summary>
-        /// <param name="scp096"><see cref="Player"/> who is SCP-096.</param>
-        /// <param name="target"><see cref="Player"/> who is the target to be added.</param>
-        /// <param name="ahpToAdd"><see cref="int"/> amount of temporary health to add to <see cref="Scp096"/>.</param>
-        /// <param name="enrageTimeToAdd"><see cref="float"/> amount of time to add to <see cref="Scp096"/>'s enrage timer. Note: This does not affect anything if he doesn't already have any targets before this event is called.</param>
-        public AddingTargetEventArgs(Player scp096, Player target, int ahpToAdd, float enrageTimeToAdd)
+        /// <param name="scp096"><inheritdoc cref="Scp096"/></param>
+        /// <param name="target"><inheritdoc cref="Target"/></param>
+        /// <param name="ahpToAdd"><inheritdoc cref="AhpToAdd"/></param>
+        /// <param name="enrageTimeToAdd"><inheritdoc cref="EnrageTimeToAdd"/></param>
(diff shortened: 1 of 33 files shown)


Exiled.Events/EventArgs/AddingTargetEventArgs.cs
```diff
@@ -41,19 +43,20 @@ public AddingTargetEventArgs(Player scp096, Player target, int ahpToAdd, float e
...
-        public bool IsAllowed { get; set; } = true;
-
         /// <summary>
         /// Gets or sets the amount of AHP to add to SCP-096 if <see cref="IsAllowed"/> is true.
         /// </summary>
         public int AhpToAdd { get; set; }
 
         /// <summary>
         /// Gets or sets how much time is added to SCP-096's enrage timer if <see cref="IsAllowed"/> is true.
+        /// Note: This does not affect anything if he doesn't already have any targets before this event is called.
```
iRebbok, line 53: Use <remark> </remark> block instead

Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
```diff
@@ -66,7 +66,7 @@ private static bool Prefix(FlashGrenade __instance)
             }
             catch (Exception exception)
             {
-                Log.Error($"Exiled.Events.Patches.Events.Map.ExplodingFlashGrenade: {exception}\n{exception.StackTrace}");
+                Log.Error($"{nameof(Handlers.Map.OnExplodingGrenade)} error: {exception}");
```
iRebbok, line 69: Firstly, nameof() won't give us the full name, secondly, it's OnExplodingGrenade, thirdly, the format isn't the same as for other Transpilers/Prefixes

iRebbok (changes requested):

Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
iopietro: True, I'll correct the first and second things.
But for the third one, doesn't `{exception}\n{exception.StackTrace}` just repeat the stack trace twice? Because by calling `exception.ToString()` you already get the stack trace.

iRebbok: It does, so just do it that way https://github.com/Exiled-Team/EXILED/blob/2176b0e8e29dc9b3cac6988aabf6c6fd73d440a5/Exiled.Events/Patches/Events/Player/DequippedMedicalItem.cs#L38

iRebbok (approved):

Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
iopietro: Ok, then we should change those methods that duplicate the stack trace, if any of them are still left.

galaxy119 (approved):
Looks good to me

PR #444, opened 2021-03-08
Changes

iopietro:
`[Exiled.Events]`
+ Changed PreAuthenticating patch into a transpiler.

`[Exiled.Example]`
+ Added Player::OnPreAuthenticating event.


iRebbok (approved):
LGTM, but don't add lots of handler to Exiled.Example, it'll be too big and complicated for newcomers.

iopietro:
> 
> 
> LGTM, but don't add lots of handler to Exiled.Example, it'll be too big and complicated for newcomers.

You're right.

galaxy119 (approved):
lgtm

PR #445, opened 2021-03-08
Added new permission

ThatsDekuu:
Since spawn isn't a List's parent, the permission should be customitems.spawn instead of customitems.list.


iopietro (approved):

iRebbok:
Don't mix changes into two PR's, you've already made one, #443. And somehow it was merged into master... I'll cherry-pick #443 into dev instead.

PR #449, opened 2021-03-09
Fix for SCP-096 being able to enter 'pre-windup' with no targets.

galaxy119:
When a target is being added, if ev.IsAllowed is set to false in Scp096.AddingTarget event, and there are no other targets, SCP-096 will still enter the 'pre-windup' enraging animation, even if their targets list is empty.

This patch prevents that behavior, returning the function if _targets HashSet is empty.


iopietro (approved):

iRebbok (approved):
Still, isn't that a base-game bug?

PR #450, opened 2021-03-09
Allows NW to be aware of custom items that may impact gameplay

galaxy119:
Currently, there are no items implemented in the Exiled-Team CustomItems that is not obviously a custom item and not a hack of some kind. (With the possible exception of shotgun, but it creates impact decals for every round it shoots, so it should be easy enough to see it's shooting multiple bullets at a time not some kinda damage hack)

A recent PR into CustomItems features a gun that is essentially server-side aimbot. 
This has raised a concerning thought: How would Northwood be able to determine the difference between someone using this item to kill a room full of people quickly, vs someone using an aimbot hack?

This PR is the answer to that question.

1. Players with a CustomItem in their hand will have their nickname changed to include the item's name, but only visible to spectators, as to not ruin immersion/etc to active players in the game.
2. Gives gmods access to the ``ci list ii {player}`` command, allowing them to see what custom item(s) the player has in their inventory, without needing local plugin permissions set.
3. When a gmod gbans someone, they will be sent a message to their console stating that they banned a player with a potentially 'suspicious' item, and that they should re-review their ban proof footage to determine if the ban may have been given to an innocent player using a custom item.

While 2 is always active, 1 & 3 will only function if CustomItem.ShouldMessageOnGban is true.
This is a public get-only bool, that defaults to false, and must be specifically set to true by the creator of the item. This is to ensure that gmods are not constantly spammed with item names that are not needed to be known (stuff like Letal Injection, SCP-1499, EMP grenades, etc) when they use the gban command, and so that only sensitive items will overwrite the player's nickname to spectators.

### DIFF
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -92,6 +92,11 @@ namespace Exiled.CustomItems.API.Features
         public HashSet<Pickup> Spawned { get; } = new HashSet<Pickup>();
 
         /// <summary>
+        /// Gets a value indicating whether whether or not this item causes things to happen that may be considered hacks, and thus be shown to global moderators as being present in a player's inventory when they gban them.
+        /// </summary>
+        public bool ShouldMessageOnGban { get; } = false;
+
+        /// <summary>
         /// Gets a <see cref="CustomItem"/> with a specific ID.
         /// </summary>
         /// <param name="id">The <see cref="CustomItem"/> ID.</param>
@@ -594,6 +599,8 @@ namespace Exiled.CustomItems.API.Features
                 ev.Player.RemoveItem(item);
 
                 Spawn(ev.Player, item);
+
+                Exiled.API.Extensions.MirrorExtensions.ResyncSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync));
             }
         }
 
@@ -614,6 +621,8 @@ namespace Exiled.CustomItems.API.Features
                 InsideInventories.Remove(item.uniq);
 
                 Spawn(ev.Target, item);
+
+                Exiled.API.Extensions.MirrorExtensions.ResyncSyncVar(ev.Target.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync));
             }
         }
 
@@ -634,6 +643,8 @@ namespace Exiled.CustomItems.API.Features
                 InsideInventories.Remove(item.uniq);
 
                 Spawn(ev.NewRole.GetRandomSpawnPoint(), item);
+
+                Exiled.API.Extensions.MirrorExtensions.ResyncSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync));
             }
         }
 
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -715,8 +715,14 @@ namespace Exiled.CustomItems.API.Features
                 return;
             }
 
-            foreach (Player player in Player.Get(RoleType.Spectator))
-                player.SendFakeSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync), $"{ev.Player.Nickname} (CustomItem: {Name})");
+            if (ShouldMessageOnGban)
+            {
+                foreach (Player player in Player.Get(RoleType.Spectator))
+                {
+                    player.SendFakeSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync), $"{ev.Player.Nickname} (CustomItem: {Name})");
+                }
+            }
+
             OnChanging(ev);
         }
 
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -94,6 +94,7 @@ namespace Exiled.CustomItems.API.Features
         /// <summary>
         /// Gets a value indicating whether whether or not this item causes things to happen that may be considered hacks, and thus be shown to global moderators as being present in a player's inventory when they gban them.
         /// </summary>
+        [YamlIgnore]
         public bool ShouldMessageOnGban { get; } = false;
 
         /// <summary>
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -710,8 +710,13 @@ namespace Exiled.CustomItems.API.Features
         private void OnInternalChanging(ChangingItemEventArgs ev)
         {
             if (!Check(ev.NewItem))
+            {
+                Exiled.API.Extensions.MirrorExtensions.ResyncSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync));
                 return;
+            }
 
+            foreach (Player player in Player.Get(RoleType.Spectator))
+                player.SendFakeSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync), $"{ev.Player.Nickname} (CustomItem: {Name})");
             OnChanging(ev);
         }
 
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -95,7 +95,7 @@ namespace Exiled.CustomItems.API.Features
         /// Gets a value indicating whether whether or not this item causes things to happen that may be considered hacks, and thus be shown to global moderators as being present in a player's inventory when they gban them.
         /// </summary>
         [YamlIgnore]
-        public bool ShouldMessageOnGban { get; } = false;
+        public virtual bool ShouldMessageOnGban { get; } = false;
 
         /// <summary>
         /// Gets a <see cref="CustomItem"/> with a specific ID.
--- a/Exiled.CustomItems/CustomItems.cs
+++ b/Exiled.CustomItems/CustomItems.cs
@@ -27,7 +27,7 @@ namespace Exiled.CustomItems
 
         private RoundHandler roundHandler;
         private ServerHandler serverHandler;
-        private CoroutineHandle handle;
+        private PlayerHandler playerHandler;
         private Harmony harmony;
 
         private CustomItems()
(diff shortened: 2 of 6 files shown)


Exiled.CustomItems/Events/ServerHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Handles Server events for the CustomItems API.
+    /// </summary>
+    internal sealed class ServerHandler
+    {
+        /// <inheritdoc cref="Events.Handlers.Server.OnSendingRemoteAdminCommand"/>
+        public void OnRemoteAdminCommand(SendingRemoteAdminCommandEventArgs ev)
+        {
+            if (ev.Name.ToLower() == "gban-kick" && ev.Sender.ReferenceHub.queryProcessor._sender.ServerRoles.RaEverywhere)
+            {
+                Player player = Player.Get(string.Join(" ", ev.Arguments));
```
iopietro, line 29: You could use the short form: `Player.Get(...) is Player player`.

Exiled.CustomItems/Events/ServerHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+
+                if (CustomItem.TryGet(player, out IEnumerable<CustomItem> customItems))
+                {
+                    StringBuilder builder = StringBuilderPool.Shared.Rent();
+
+                    foreach (CustomItem item in customItems)
+                    {
+                        if (item.ShouldMessageOnGban)
+                        {
+                            builder.AppendLine(item.Name + " - " + item.Description);
```
iopietro, line 41: You should use `builder.Append(...).Append(...).AppendLine(...)` instead of contatenating strings.

Exiled.CustomItems/CustomItems.cs
```diff
@@ -36,24 +43,53 @@ private CustomItems()
...
             harmony.UnpatchAll();
 
             harmony = null;
             roundHandler = null;
+            serverHandler = null;
 
             base.OnDisabled();
         }
+
+        private IEnumerator<float> ShowCustomItemInNickname()
```
iopietro, line 72: Is this necessary? Could the fake nickname be updated everytime someone changes its role and become a spectator? 
In servers with 50 players, it would iterate 2500 times every 0.5 seconds.

Exiled.CustomItems/Events/ServerHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+                        }
+                    }
+
+                    string itemNames = StringBuilderPool.Shared.ToStringReturn(builder);
+
+                    if (!string.IsNullOrEmpty(itemNames))
+                    {
+                        ev.Sender.SendConsoleMessage(
```
iopietro, line 49: You should also send a broadcast to the global mod, just in case he don't look at the console.

galaxy119: They must have the console open in order to use the gban command.

Exiled.CustomItems/CustomItems.cs
iopietro: So, shouldn't it update when you have a custom "hacky" item in your hand?
It could be done with the `ChangingItem` event.

Exiled.CustomItems/Events/ServerHandler.cs
iopietro: Yeah, the broadcast purpose would have been for extra security, but it's fine if it is not necessary.

Exiled.CustomItems/CustomItems.cs
galaxy119: Since I don't want everyone on the server to see their new name, only spectators, we would still need to regularly check if someone is/isn't a spectator anymore, and change the user's name/reset it.

iopietro: Yeah, we could check it with `ChangingItem` and `ChangingRole`, to remove the custom nickname to non-spectators.

galaxy119: But what happens if someone changes to the custom item, then someone new dies/joins the server. Their fake nickname would not be updated. 

This could be a problem if, for example, someone is using a gun that shoots through walls, and are reported by a player for 'hacking'. 
The joining gmod would not see the player's changed nickname.

Alternatively, if we make it so that anyone triggering ChangingRole with the new role being spectator, the above issue would be resolved, but we would have to foreach over every player when they change into something not spectator and re-sync each player's name.

`
switch (ev.NewRole == RoleType.Spectator)
{
    case true:
            if (CustomItem.TryGet(player, out IEnumerable<Customitem> customItems))
                foreach (CustomItem item in customitems)
                    if (item.Check(player.CurrentItem))
                        ev.Player.SendFakeSyncVar(...)
            break;
    case false:
            // reset player's nickname
        break;
}
`

I'm not sure that's all that much more efficient. It spreads out how often the foreach() is called, but will likely result in it being called *much* more frequently.

iopietro: > But what happens if someone changes to the custom item, then someone new dies/joins the server. Their fake nickname would not be updated.

That's why we should update the custom nickname when you change the custom item and when someone class gets changed to spectator (when you join, you get automatically changed to spectator).

The `ChangingRole` event will only update the custom nick for the joined player and not for all spectators (you don't need a foreach).
The `ChangingItem` event will change the custom nickname for all spectators (you only need a single foreach).

galaxy119: In ChangingRole I need a foreach to go over a list of all the spawned players, check if they have custom items, then foreach over those custom items and check to see if any of them are their current weapon.

I also need to reset their nickname back if the player's NewRole isn't Spectator aswell, otherwise it defeats the purpose of only targeting spectators for the name change.

iopietro: > In ChangingRole I need a foreach to go over a list of all the spawned players, check if they have custom items, then foreach over those custom items and check to see if any of them are their current weapon.

Why? You can simply subscribe that event to `CustomItem` class and then change the nick of players who are using it, you may still need a foreach tho.
Same thing when that player spawns, but in that case you can reset every nicknames without the need to check for custom items.

iRebbok: I'm in favor of @iopietro's event driven solution. Having a coroutine in the framework that iterates through all players every 0.5 seconds isn't a good idea.

galaxy119: Very well, I will remove the coroutine and adapt an event driven solution and re-request your reviews shortly.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -710,8 +710,13 @@ private void OnInternalPickingUp(PickingUpItemEventArgs ev)
         private void OnInternalChanging(ChangingItemEventArgs ev)
         {
             if (!Check(ev.NewItem))
+            {
+                Exiled.API.Extensions.MirrorExtensions.ResyncSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync));
                 return;
+            }
 
+            foreach (Player player in Player.Get(RoleType.Spectator))
```
iopietro, line 718: Shouldn't the nickname be synced only if the item is "hacky"?

Exiled.CustomItems/Events/PlayerHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+    using Exiled.Events.EventArgs;
+
+    /// <summary>
+    /// Handles Player events for the CustomItem API.
+    /// </summary>
+    internal sealed class PlayerHandler
+    {
+        public void OnChangingRole(ChangingRoleEventArgs ev)
+        {
+            switch (ev.NewRole == RoleType.Spectator)
```
iopietro, line 22: Why a switch instead of an if?

Exiled.CustomItems/Events/PlayerHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                case true:
+                {
+                    foreach (Player player in Player.List)
+                    {
+                        if (player == ev.Player)
+                            continue;
+
+                        if (CustomItem.TryGet(player, out CustomItem item))
```
iopietro, line 31: You should only check if the player has this CustomItem in their hand and not any possible CustomItem.
That check should also be done on "hacky" custom items only.

Exiled.CustomItems/Events/PlayerHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+                }
+
+                case false:
+                {
+                    foreach (Player player in Player.List)
+                    {
+                        if (player == ev.Player)
+                            continue;
+
+                        ev.Player.SendFakeSyncVar(player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync), player.Nickname);
```
iopietro, line 47: Couldn't `ResyncSyncVar` be used like above?
Also, this check should only be done on "hacky" custom items.

Exiled.CustomItems/API/Features/CustomItem.cs
galaxy119: quite right

Exiled.CustomItems/Events/PlayerHandler.cs
galaxy119: Uhm. no real reason, does it particularly matter?

Exiled.CustomItems/Events/PlayerHandler.cs
galaxy119: TryGet(this Player) has 2 overloads, one returns an IEnumerable<CustomItem> that checks for any item in their inventory, and one that returns a single CustomItem if it's in their hand.

iopietro: Yes, I know, but then you've to check if that item correspond to the right one (the `this` CustomItem) or another random CustomItem he's using.

Exiled.CustomItems/Events/PlayerHandler.cs
iopietro: Yeah, I was thinking about using it instead of the foreach.

Exiled.CustomItems/Events/PlayerHandler.cs
galaxy119: This event is not inside the CustomItem class, otherwise i'd just
`
if (Check(player.CurrentItem))
`

iopietro: Ah sorry, my bad, I thought it was there.

Exiled.CustomItems/Events/PlayerHandler.cs
galaxy119: But this is in OnChangingRole, I only want the person who is becoming non-spectator to have the name re-synced. ResyncSyncVar would reset the name for all players, so a foreach is needed here.

iopietro: Like in the other comment, I thought it was inside CustomItem class.

Exiled.CustomItems/Events/PlayerHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                foreach (Player player in Player.List)
+                {
+                    if (player == ev.Player)
+                        continue;
+
+                    if (CustomItem.TryGet(player, out CustomItem item))
+                    {
+                        if (item.ShouldMessageOnGban)
+                            ev.Player.SendFakeSyncVar(player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync), $"{ev.Player.Nickname} (CustomItem: {item.Name})");
```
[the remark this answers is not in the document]
galaxy119: We could, instead of making them synced to spectators, to be synced with NW staff, but this feature could also help reduce the chance of another player reporting them as a hacker.

iopietro (approved):

Exiled.CustomItems/Events/PlayerHandler.cs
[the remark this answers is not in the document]
iRebbok: Then it's okay.

BuildBoy12 (approved):

PR #453, opened 2021-03-10
Turn the giant CallCmdShoot prefix patch into a transpiler.

galaxy119:
Currently the patch used to add the events Player.Shooting and Player.Shot is a very large Prefix patch that always returns false.

These types of patches can be troublesome when the base-game method is changed, but those changes fail to get reflected in the patch properly. 

This changes the patch type to a transpiler, so as to not interfere with the normal logic of the anticheat included in the base method, and to allow easier updating of the patch in the future should things change.

### DIFF
--- a/Exiled.Events/Patches/Events/Player/Shoot.cs
+++ b/Exiled.Events/Patches/Events/Player/Shoot.cs
@@ -54,7 +54,7 @@ namespace Exiled.Events.Patches.Events.Player
             {
                 // Player.Get(this.gameObject)
                 new CodeInstruction(OpCodes.Ldarg_0).MoveLabelsFrom(newInstructions[index]),
-                new CodeInstruction(OpCodes.Call, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager._hub))),
+                new CodeInstruction(OpCodes.Ldfld, Field(typeof(WeaponManager), nameof(WeaponManager._hub))),
                 new CodeInstruction(OpCodes.Call, Method(typeof(Player), nameof(Player.Get), new[] { typeof(ReferenceHub) })),
 
                 // GameObject target
@@ -89,7 +89,7 @@ namespace Exiled.Events.Patches.Events.Player
             {
                 // Player.Get(this.gameObject)
                 new CodeInstruction(OpCodes.Ldarg_0).MoveLabelsFrom(newInstructions[index]),
-                new CodeInstruction(OpCodes.Call, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager._hub))),
+                new CodeInstruction(OpCodes.Ldfld, Field(typeof(WeaponManager), nameof(WeaponManager._hub))),
                 new CodeInstruction(OpCodes.Call, Method(typeof(Player), nameof(Player.Get), new[] { typeof(ReferenceHub) })),
 
                 // GameObject target
--- a/Exiled.Events/Patches/Events/Player/Shoot.cs
+++ b/Exiled.Events/Patches/Events/Player/Shoot.cs
@@ -81,9 +81,9 @@ namespace Exiled.Events.Patches.Events.Player
                 new CodeInstruction(OpCodes.Brfalse, returnLabel),
             });
 
-            // Search for the only "PlayerStats.HurtPlayer()"
-            offset = -34;
-            index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Callvirt && (MethodInfo)instruction.operand == Method(typeof(PlayerStats), nameof(PlayerStats.HurtPlayer))) + offset;
+            // Search for the last "Div" call before the only "PlayerStats.HurtPlayer()"
+            offset = +2;
+            index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Div) + offset;
 
             newInstructions.InsertRange(index, new[]
             {
--- a/Exiled.Events/Patches/Events/Player/Shoot.cs
+++ b/Exiled.Events/Patches/Events/Player/Shoot.cs
@@ -8,15 +8,23 @@
 namespace Exiled.Events.Patches.Events.Player
 {
 #pragma warning disable SA1313
-    using System;
+#pragma warning disable SA1118
+    using System.Collections.Generic;
+    using System.Reflection;
+    using System.Reflection.Emit;
+
+    using CustomPlayerEffects;
 
     using Exiled.API.Features;
     using Exiled.Events.EventArgs;
-    using Exiled.Loader;
 
     using HarmonyLib;
 
+    using NorthwoodLib.Pools;
+
     using UnityEngine;
+
+    using static HarmonyLib.AccessTools;
 
 #pragma warning disable SA1512 // Single-line comments should not be followed by blank line
 #pragma warning disable SA1005 // Single line comments should begin with single space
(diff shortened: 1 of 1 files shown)


Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -29,276 +37,101 @@ namespace Exiled.Events.Patches.Events.Player
...
-
-                    __instance.PlaceDecal(false, new Ray(__instance.camera.position, dir), __instance.curWeapon, 0f);
-                    __instance.RpcConfirmShot(false, __instance.curWeapon);
-                    return false;
-                }
-            }
-            catch (Exception e)
+                // Player.Get(this.gameObject)
+                new CodeInstruction(OpCodes.Ldarg_0).MoveLabelsFrom(newInstructions[index]),
+                new CodeInstruction(OpCodes.Callvirt, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager.gameObject))),
```
iRebbok, line 57: Does it work? The |gameObject` propery is from the `Component` and it should be `call` cuz it's `this` above.

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -29,276 +37,101 @@ namespace Exiled.Events.Patches.Events.Player
...
+            // Search for the only "PlayerStats.HurtPlayer()"
+            offset = 34;
+            index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Callvirt && (MethodInfo)instruction.operand == Method(typeof(PlayerStats), nameof(PlayerStats.HurtPlayer))) - offset;
+
+            newInstructions.InsertRange(index, new[]
             {
-                Exiled.API.Features.Log.Error($"{typeof(Shoot).FullName}.{nameof(Prefix)}:\n{e}");
+                // Player.Get(this.gameObject)
+                new CodeInstruction(OpCodes.Ldarg_0).MoveLabelsFrom(newInstructions[index]),
+                new CodeInstruction(OpCodes.Callvirt, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager.gameObject))),
```
iRebbok, line 92: Same

Exiled.Events/Patches/Events/Player/Shoot.cs
galaxy119: It uses callvirt in the base-game IL code, so that's why I used it here. I double checked and the event works as expected with Callvirt.

Exiled.Events/Patches/Events/Player/Shoot.cs
galaxy119: same

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -29,276 +37,101 @@ namespace Exiled.Events.Patches.Events.Player
...
+        private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions, ILGenerator generator)
         {
-            try
+            List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Shared.Rent(instructions);
+
+            // search for "this._hub" before the only "PlayerEffect.ServerDisable()"
+            int offset = 17;
+            int index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Callvirt && (MethodInfo)instruction.operand == Method(typeof(PlayerEffect), nameof(PlayerEffect.ServerDisable))) - offset;
+
+            LocalBuilder ev1 = generator.DeclareLocal(typeof(ShootingEventArgs));
```
iopietro, line 48: You should call them shootingEv and shotEv, to understand them better when looking at the IL code.

Exiled.Events/Patches/Events/Player/Shoot.cs
iopietro: You could even load the `this._hub` field instead of calling the `gameObject` getter.
I looked at `WeaponManager.CallCmdShoot` and the compiler uses `Call` to call `this.gameObject`.

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -29,276 +37,101 @@ namespace Exiled.Events.Patches.Events.Player
...
     {
-        private static bool Prefix(WeaponManager __instance, GameObject target, HitBoxType hitboxType, Vector3 dir, Vector3 sourcePos, Vector3 targetPos)
+        private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions, ILGenerator generator)
         {
-            try
+            List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Shared.Rent(instructions);
+
+            // search for "this._hub" before the only "PlayerEffect.ServerDisable()"
+            int offset = 17;
+            int index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Callvirt && (MethodInfo)instruction.operand == Method(typeof(PlayerEffect), nameof(PlayerEffect.ServerDisable))) - offset;
```
iopietro, line 46: This should always be an addition, only the offset should have the minus or plus sign.

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -29,276 +37,101 @@ namespace Exiled.Events.Patches.Events.Player
...
+
+                // if (!ev.IsAllowed)
+                //    return;
+                new CodeInstruction(OpCodes.Callvirt, PropertyGetter(typeof(ShootingEventArgs), nameof(ShootingEventArgs.IsAllowed))),
+                new CodeInstruction(OpCodes.Brfalse, returnLabel),
+            });
+
+            // Search for the only "PlayerStats.HurtPlayer()"
+            offset = 34;
+            index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Callvirt && (MethodInfo)instruction.operand == Method(typeof(PlayerStats), nameof(PlayerStats.HurtPlayer))) - offset;
```
iopietro, line 86: Same as above.

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -29,276 +37,101 @@ namespace Exiled.Events.Patches.Events.Player
...
+                new CodeInstruction(OpCodes.Call, Method(typeof(Handlers.Player), nameof(Handlers.Player.OnShooting))),
+
+                // if (!ev.IsAllowed)
+                //    return;
+                new CodeInstruction(OpCodes.Callvirt, PropertyGetter(typeof(ShootingEventArgs), nameof(ShootingEventArgs.IsAllowed))),
+                new CodeInstruction(OpCodes.Brfalse, returnLabel),
+            });
+
+            // Search for the only "PlayerStats.HurtPlayer()"
+            offset = 34;
```
iopietro, line 85: This is fine, but is this the nearest unique opcode?

Exiled.Events/Patches/Events/Player/Shoot.cs
galaxy119: will offset = -17
index = ... + offset  work properly? I've never tried to 'add' a negative number to something like that, lol.

Exiled.Events/Patches/Events/Player/Shoot.cs
galaxy119: ldstr "W.11" or even div would technically be closer, but in the IL viewer they are above the IL call I want to insert above, but in the C# code they are below it, in a switch-case, and I wasn't certain how that works. If I got the index of the last `div` call, and offset it by +2, in the IL code it would be right where I want it, but that IL call is in a switch case that's only used in switch-case, so i wasn't sure how it would behave.

Exiled.Events/Patches/Events/Player/Shoot.cs
iopietro: Yes, it will work fine.

Exiled.Events/Patches/Events/Player/Shoot.cs
iopietro: Yeah, that's normal, you can safely use that `div` call.

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -42,20 +42,20 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
 
             Label returnLabel = generator.DefineLabel();
 
             newInstructions.InsertRange(index, new[]
             {
                 // Player.Get(this.gameObject)
                 new CodeInstruction(OpCodes.Ldarg_0).MoveLabelsFrom(newInstructions[index]),
-                new CodeInstruction(OpCodes.Callvirt, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager.gameObject))),
-                new CodeInstruction(OpCodes.Call, Method(typeof(Player), nameof(Player.Get), new[] { typeof(GameObject) })),
+                new CodeInstruction(OpCodes.Call, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager._hub))),
```
iopietro, line 57: You've to use `OpCodes.Ldfld` here.

Exiled.Events/Patches/Events/Player/Shoot.cs
```diff
@@ -82,15 +82,15 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
+            offset = -34;
+            index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Callvirt && (MethodInfo)instruction.operand == Method(typeof(PlayerStats), nameof(PlayerStats.HurtPlayer))) + offset;
 
             newInstructions.InsertRange(index, new[]
             {
                 // Player.Get(this.gameObject)
                 new CodeInstruction(OpCodes.Ldarg_0).MoveLabelsFrom(newInstructions[index]),
-                new CodeInstruction(OpCodes.Callvirt, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager.gameObject))),
-                new CodeInstruction(OpCodes.Call, Method(typeof(Player), nameof(Player.Get), new[] { typeof(GameObject) })),
+                new CodeInstruction(OpCodes.Call, PropertyGetter(typeof(WeaponManager), nameof(WeaponManager._hub))),
```
iopietro, line 92: Same here.

Exiled.Events/Patches/Events/Player/Shoot.cs
galaxy119: Right, I forgot _hub was a field

iopietro (approved):

iRebbok (approved):

BuildBoy12 (approved):

Issue #459, opened 2021-03-13
Add optional translation config


PintTheDragon:
Adding the ability to translate messages can be very helpful to a lot of plugins, but will clutter up the config significantly. This is why I propose there be a separate config for translations, that plugins are not required to use. Before, I'd like to define what might go inside of the translation config, and what can be left out of it:

Messages used inside of plugins should go inside of the translation config. These messages help certain plugin features, but are not a feature themselves.
Custom broadcast/hint messages should be inside of the plugin config. These exist as plugin features on their own, and shouldn't be with the other translations. There are very few of these types of messages, so they won't clutter the config and won't get mixed up with the other translations.

Translations can be implemented as follows:

IPlugin/Plugin is given a new property, of type ITranslation, which is default to null.
A new interface called ITranslation, similar to IConfig is created, although it does not include is_enabled.
Where the config manager loads configs, it also loads translations from a separate translations file, ignoring plugins that have null translations.

This should be an easy-to-implement feature, as most of what is required already exists in the config manager, and will make translating plugins much easier.

iRebbok:
What kind of messages should plugins use from the translation config if not custom broadcasts? `Log.Warn`/`Log.Error`?

galaxy119:
It'd basically be any language localization options the plugin dev wants to support, such as changing broadcasts/hints, console messages, etc.

PintTheDragon:
Not sure why this issue didn't close automatically when the PR got merged.

PR #460, opened 2021-03-13
Add translations

PintTheDragon:
This adds an optional translations config to plugins that can be used by overriding the Translations property with a new instance of a class implementing ITranslations. This PR implements #459.


iopietro:
Imo it shouldn't be coupled inside `ConfigManager`.

Exiled.Loader/ConfigManager.cs
```diff
@@ -114,12 +114,78 @@ public static class ConfigManager
             }
         }
 
+        /// <summary>
+        /// Loads all plugin translations.
+        /// </summary>
+        /// <param name="rawTranslations">The raw translations to be loaded.</param>
+        /// <returns>Returns a dictionary of loaded translations.</returns>
+        public static SortedDictionary<string, ITranslations> LoadSortedTranslations(string rawTranslations)
```
iRebbok, line 122: The `Sorted` was added because of backward compatibility to the name of the config load method. No requirement to do the same with that method.

iRebbok (changes requested):
Looks good to me. But as per Pietro's comment, I also want to see translation management in a different class.

PintTheDragon:
Alright. Everything has been moved to a new file and SortedDictionaries have been changed to Dictionaries.

iRebbok:
I didn't mean to change from `SortedDictionary` to `Dictionary`, I meant to remove the `Sorted` form the method name.

Exiled.Events/Commands/Reload/Configs.cs
```diff
@@ -42,7 +42,7 @@ public bool Execute(ArraySegment<string> arguments, ICommandSender sender, out s
                 return false;
             }
 
-            bool haveBeenReloaded = ConfigManager.Reload();
+            bool haveBeenReloaded = ConfigManager.Reload() && TranslationManager.Reload();
```
iRebbok, line 45: How about separating that into a `reload translations` command?

iRebbok (approved):
Alright.

Exiled.Events/Commands/Reload/Translations.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <summary>
+        /// Gets static instance of the <see cref="Translations"/> command.
+        /// </summary>
+        public static Translations Instance { get; } = new Translations();
+
+        /// <inheritdoc/>
+        public string Command { get; } = "translations";
+
+        /// <inheritdoc/>
+        public string[] Aliases { get; } = new string[] { };
```
BuildBoy12, line 31: Either having an alias or having it just be an Array.Empty would be better.

BuildBoy12 (approved):
LGTM

Exiled.Loader/TranslationManager.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// <summary>
+    /// Used to handle plugin translations.
+    /// </summary>
+    public static class TranslationManager
+    {
+        /// <summary>
+        /// Gets the translation serializer.
+        /// </summary>
+        public static ISerializer Serializer { get; } = new SerializerBuilder()
```
iopietro, line 30: Since they're equal, you should either directly use the `ConfigManager` serializer and deserializer or move both of them in the `Loader` class to decouple them from the `ConfigManager`.

Exiled.Loader/TranslationManager.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <returns>Returns a dictionary of loaded translations.</returns>
+        public static SortedDictionary<string, ITranslations> Load(string rawTranslations)
+        {
+            try
+            {
+                Log.Info("Loading plugin translations...");
+
+                Dictionary<string, object> rawDeserializedTranslations = Deserializer.Deserialize<Dictionary<string, object>>(rawTranslations) ?? new Dictionary<string, object>();
+                SortedDictionary<string, ITranslations> deserializedTranslations = new SortedDictionary<string, ITranslations>(StringComparer.Ordinal);
+
```
iopietro, line 60: Shouldn't we support translations for `Exiled.Loader`?

Exiled.Events/Commands/Reload/Translations.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                response = "You can't reload translations, you don't have \"ee.reloadtranslations\" permission.";
+                return false;
+            }
+
+            bool haveBeenReloaded = TranslationManager.Reload();
+
+            Handlers.Server.OnReloadedTranslations();
+
+            response = "Plugin translations have been reloaded successfully!";
```
iopietro, line 49: The response should be different when `haveBeenReloaded` is false.

Exiled.Events/Commands/Reload/Translations.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            if (!sender.CheckPermission("ee.reloadtranslations"))
+            {
+                response = "You can't reload translations, you don't have \"ee.reloadtranslations\" permission.";
+                return false;
+            }
+
+            bool haveBeenReloaded = TranslationManager.Reload();
+
+            Handlers.Server.OnReloadedTranslations();
```
iopietro, line 47: The event shouldn't be fired if the reload fails.

Exiled.Events/Handlers/Internal/Round.cs
```diff
@@ -25,7 +25,10 @@ public static void OnWaitingForPlayers()
             MultiAdminFeatures.CallEvent(MultiAdminFeatures.EventType.WAITING_FOR_PLAYERS);
 
             if (Events.Instance.Config.ShouldReloadConfigsAtRoundRestart)
```
iopietro, line 27: Shouldn't we create a new config for reloading translations when the round restarts?
Like `ShouldReloadTranslationsAtRoundRestart`.

Exiled.API/Interfaces/IPlugin.cs
```diff
@@ -68,6 +68,11 @@ public interface IPlugin<out TConfig> : IComparable<IPlugin<IConfig>>
         /// </summary>
         TConfig Config { get; }
 
+        /// <summary>
+        /// Gets the plugin translations.
+        /// </summary>
+        ITranslations Translations { get; }
```
iopietro, line 74: By doing this, to get a translation string, you'd be forced to cast `Translations` to your `Translation` class.
Like: `((Translation)Translations).TranslationString`, which is not so intuitive or comfortable to do.

Exiled.API/Interfaces/ITranslations.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Interfaces
+{
+    /// <summary>
+    /// Defines the contract for basic translation features.
+    /// </summary>
+    public interface ITranslations
```
iopietro, line 13: This should be named `ITranslation`, for consistency with `IConfig`.

iopietro (commented):
This system is fine, but in my opinion we should support more than one language per plugin.
I'd like to put translations inside a different folder like `EXILED/Plugins/Translations`, then create a folder for each plugin and a file for each language it supports.
Like
+ EXILED
  + Plugins
      + Languages
          + AdminTools
              + english.yml
              + italian.yml
              + french.yml

So, if a plugin does have a translation in the language you like, you won't need to change anything.
On the other hand, if a plugin doesn't have it, you could just translate it by yourself and then create a PR for that plugin, so other people who speak your language, would be able to use your translation without the need to make their one.

Also, we could even add the `Language` property in `IConfig`.

Exiled.Events/Commands/Reload/Translations.cs
PintTheDragon: This is an issue with all of the reload commands and should be fixed in a separate PR.

Exiled.Events/Commands/Reload/Translations.cs
PintTheDragon: This is an issue with all of the reload commands and should be fixed in a separate PR.

Exiled.Loader/TranslationManager.cs
PintTheDragon: I'm a bit confused about what you mean by this. The highlighted section doesn't make much sense with it either. Could you elaborate on it a bit?

Exiled.API/Interfaces/IPlugin.cs
PintTheDragon: I can't think of too many ways to do this so that translations are made optional. Maybe we can provide another plugin class that inherits from the main one, with a second generic type used in an override of Translations. I'm not too sure if that'll work though. I'll have to think a bit on how to solve this.

PintTheDragon: Alright, I've made a basic implementation of what I described above. I'll need to do some testing on it later, but it should be working. Let me know if you see any issues with it. I'll continue going through the list of requested changes now as well.

PintTheDragon:
I understand the idea for multiple languages, but it doesn't seem like something that would be worth it. Translations are really just meant to be a second config that plugins can put all of their messages and text content into to declutter the main config and let server owners have a bit more control over what the plugins do.

Plus, it would be a hassle for EXILED to implement all of the plugin folders and multiple translations, and the translation, being a required field, would need to go somewhere other than in the main config as adding a new field to that would break every plugin using it. There would also need to be a way to pack the translations into the plugin assembly. Then, the plugin developers would need to figure out how to get their plugin translated, and keep all of their translations up-to-date with the latest changes or else people will end up with half-translated translations on their server because only new translations will get added to be up-to-date with the current translation. Others won't change because server owners need to be able to change the translations.

All of this is something that I'm not going to deal with in the PR. If you want to deal with the whole mess of predefined translations, then go ahead, but it just seems like it would be a bad idea for everyone involved.

Exiled.Loader/TranslationManager.cs
PintTheDragon: I think I understand. We could add translations to the loader, but it only ever sends messages to the console and I'd rather spend time implementing the actual translation feature. As with the multiple translations, you can add it later if you want, but I don't think it would really be worth the effort to change the messages at the start of the console.

iopietro: Exactly that.
`Exiled.Loader` could print some errors/warnings about plugin/dependencies not loading correctly or outdated plugins, so in my opinion it would be useful.

iopietro:
I got your point, maybe we'll add a more complex translation system in EXILED 3.0.0, we'll see.

Exiled.API/Interfaces/IPlugin.cs
iopietro: It's fine, if it works, I'll approve it.

PintTheDragon:
I'm having a strange issue loading a plugin where commands aren't working. Translations are generated correctly although it's difficult to check if they're working as every translation is in a command. Here's a link to my source if anyone wants to take a look at it for issues: https://github.com/SCPStats/Plugin/tree/translations.

iRebbok:
> I'm having a strange issue loading a plugin where commands aren't working. Translations are generated correctly although it's difficult to check if they're working as every translation is in a command. Here's a link to my source if anyone wants to take a look at it for issues: https://github.com/SCPStats/Plugin/tree/translations.

Check which is loaded first, IIRC we load configs (& translations) after we registered plugin commands, if it's true, it's the issue.

PintTheDragon:
I think I found the issue, although I'm unsure exactly what causes it. When loading a plugin, base.OnEnabled lists the version as Exiled.API's version, even though it uses this.Version. When logging this.Version in OnEnabled, the correct version is logged. I'm assuming this is something like the version override is only overriding Plugin<, >'s Version field, while Plugin<> keeps the same Version. OnEnabled is ran on Plugin<> though, so when it uses the Version, it has the default one.

PintTheDragon:
Ok turns out it was because I guess Assembly.GetCallingAssembly() was being used for Plugin<,> instead of the type that overrode Plugin<,>. I've made Assembly in Plugin<> virtual and override it in Plugin<,> with the calling assembly. The only downside is that tinybrains can now override Assembly.

Also, luckily it turns out that C# inheritance is normal and not at all what I described above. The only weird thing is with Assembly.GetCallingAssembly().

Everything seems to be working fine. I'm gonna do some proper testing now that it actually runs though, and I'll fix any errors I find.

PintTheDragon:
Looks good. I had to make a small change to my test plugin to automatically reload commands on translation reload so that command names update, but that's expected. Other than that, everything worked perfectly fine. I'd say it's good to be merged.

Exiled.API/Features/Plugin.cs
```diff
@@ -167,6 +167,9 @@ public Plugin()
             InternalTranslation = new TTranslation();
         }
 
+        /// <inheritdoc/>
+        public override Assembly Assembly { get; } = Assembly.GetCallingAssembly();
```
iopietro, line 171: Can't the assembly be set in the constructor?

PintTheDragon: Maybe. I can try doing it. Rider was yelling at me before though.

PintTheDragon: Just tested, it's working fine in the constructor.

Exiled.API/Features/Plugin.cs
```diff
@@ -32,14 +32,15 @@ public abstract class Plugin<TConfig> : IPlugin<TConfig>
...
+            Assembly = Assembly.GetCallingAssembly();
             Name = Assembly.GetName().Name;
             Prefix = Name.ToSnakeCase();
             Author = Assembly.GetCustomAttribute<AssemblyCompanyAttribute>()?.Company;
             Version = Assembly.GetName().Version;
         }
 
         /// <inheritdoc/>
-        public virtual Assembly Assembly { get; } = Assembly.GetCallingAssembly();
+        public virtual Assembly Assembly { get; protected set; }
```
iopietro, line 43: You don't need a setter to set things from the constructor.

Exiled.API/Features/Plugin.cs
```diff
@@ -164,12 +165,10 @@ public abstract class Plugin<TConfig, TTranslation> : Plugin<TConfig>
         /// </summary>
         public Plugin()
         {
```
iopietro, line 167: You've to call the base constructor as well, the only problem is that by doing this, the `Assembly` property would be overwritten.
So you've to instantiate the name, prefix, author, etc. in this constructor as well.

PintTheDragon: Does the base constructor not run from the class inheriting it? I tried it and got two constructors running, although I didn't test for EXILED.

PintTheDragon (commented):
What the fuck did I press?

Exiled.API/Features/Plugin.cs
iopietro: Yes, you're right, the parameterless constructor of inherited classes is being called anyway.

iopietro (approved):

PintTheDragon:
Ok, everything's finally done now. I'd say it looks good to be merged.

Exiled.Loader/ConfigManager.cs
```diff
@@ -14,38 +14,14 @@ namespace Exiled.Loader
...
 
     /// <summary>
     /// Used to handle plugin configs.
     /// </summary>
     public static class ConfigManager
     {
-        /// <summary>
-        /// Gets the config serializer.
-        /// </summary>
-        public static ISerializer Serializer { get; } = new SerializerBuilder()
```
iopietro, line 32: You shouldn't completely remove Serializer and Deserializer, to not break plugins that use those two properties, you should "redirect" them to the moved Serializer and Deserializer.
Like `public static ISerializer Serializer => Loader.Serializer;`.

PR #461, opened 2021-03-14
Move RoleType.GetRandomSpawnPoint extension to Extensions.Role

steven4547466:
I feel like RoleType.GetRandomSpawnPoint is misplaced with Extensions.Role existing and it took me a minute to find that it was actually in Features.Map. I believe it makes more sense to place it in Extensions.Role.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -401,7 +401,7 @@ namespace Exiled.API.Features
         /// </summary>
         /// <param name="roleType">The <see cref="RoleType"/> to get the spawn point from.</param>
         /// <returns>Returns the spawn point <see cref="Vector3"/>.</returns>
-        [Obsolete("Moved to Exiled.API.Extensions.Role.")]
+        [Obsolete("Moved to Exiled.API.Extensions.Role.GetRandomSpawnPoint(RoleType).")]
         public static Vector3 GetRandomSpawnPoint(this RoleType roleType)
         {
             return Extensions.Role.GetRandomSpawnPoint(roleType);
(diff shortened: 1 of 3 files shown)


steven4547466:
This would create ambiguity for plugins using both `Exiled.API.Features.Map` and `Exiled.API.Extensions`, however, which is why the build fails.

iRebbok:
Then fix the build.

steven4547466:
> Then fix the build.

Without renaming the extension completely, I don't think you can. Unless you prefer me to remove the method completely from Features.Map, which does introduce other problems.

iRebbok:
> Without renaming the extension completely, I don't think you can. Unless you prefer me to remove the method completely from Features.Map, which does introduce other problems.

Don't remove the old extension, just use the new one explicitly (i.e. `Extensions.RoleType.GetRandomSpawnPoint(RoleType)`).

Exiled.API/Features/Map.cs
```diff
@@ -396,6 +396,17 @@ public static void ShowHint(string message, float duration)
         /// </summary>
         public static void ClearBroadcasts() => Server.Broadcast.RpcClearElements();
 
+        /// <summary>
+        /// Gets a random spawn point of a <see cref="RoleType"/>.
+        /// </summary>
+        /// <param name="roleType">The <see cref="RoleType"/> to get the spawn point from.</param>
+        /// <returns>Returns the spawn point <see cref="Vector3"/>.</returns>
+        [Obsolete("Moved to Exiled.API.Extensions.Role.")]
```
iRebbok, line 404: Complete the message.

iRebbok (changes requested):

Exiled.API/Features/Map.cs
steven4547466: I wasn't entirely sure what you meant, but I hope what I added was what you meant because I need to go to sleep, if it isn't, I'll fix it in the morning.

iRebbok (approved):

galaxy119 (approved):

BuildBoy12 (approved):

PR #464, opened 2021-03-18
CI.Spawn Pickup

SebasCapo:
I hate naming my PRs.
- Added out param for CustomItem.TrySpawn() method.
- CustomItem.Spawn() now returns the Pickup component.
With stuff like this the dev can modify different settings of that specific pickup, or even access the GameObject & other components, which is necessary for some implementations.

### DIFF
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -468,7 +468,7 @@ namespace Exiled.CustomItems.API.Features
 
                 spawned++;
 
-                Spawn(spawnPoint.Position.ToVector3);
+                Spawn(spawnPoint.Position.ToVector3, out Pickup pickup);
 
                 Log.Debug($"Spawned {Name} at {spawnPoint.Position} ({spawnPoint.Name})", Instance.Config.Debug);
             }
@@ -697,7 +697,7 @@ namespace Exiled.CustomItems.API.Features
 
                 ev.Player.RemoveItem(item);
 
-                Spawn(ev.Player, item);
+                Spawn(ev.Player, item, out Pickup pickup);
 
                 Exiled.API.Extensions.MirrorExtensions.ResyncSyncVar(ev.Player.ReferenceHub.networkIdentity, typeof(NicknameSync), nameof(NicknameSync.Network_myNickSync));
             }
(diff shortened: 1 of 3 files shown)


galaxy119 (changes requested):
The existing methods should remain with an [Obsolete] attribute attached, and the new ones to be used by the API and in future plugins should be created as new methods, otherwise this becomes an unnecessarily API breaking change.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -203,7 +203,46 @@ public static bool TryGet(Pickup pickup, out CustomItem customItem)
...
+            return true;
+        }
+
+        /// <summary>
+        /// Tries to spawn a specific <see cref="CustomItem"/> at a specific <see cref="Vector3"/> position.
+        /// </summary>
+        /// <param name="id">The ID of the <see cref="CustomItem"/> to spawn.</param>
+        /// <param name="position">The <see cref="Vector3"/> location to spawn the item.</param>
+        /// <returns>Returns a value indicating whether the <see cref="CustomItem"/> was spawned or not.</returns>
+        [Obsolete("Use TrySpawn method with out parameter instead.")]
```
iRebbok, line 245: Generally speaking, the `out parameter` is a bit confusing as it's natively a parameter modifier, so it's better to add the `modifier`.

iRebbok (approved):

galaxy119 (approved):

Exiled.CustomItems/API/Features/CustomWeapon.cs
```diff
@@ -59,7 +59,21 @@ public virtual uint ClipSize
         public override float Durability { get; set; }
 
         /// <inheritdoc/>
-        public override void Spawn(Vector3 position) => Spawned.Add(Item.Spawn(Type, ClipSize, position, default, Modifiers.SightType, Modifiers.BarrelType, Modifiers.OtherType));
+        public override void Spawn(Vector3 position, out Pickup pickup)
```
iopietro, line 62: We should directly return the pickup instead of using an `out` parameter.
The same applies for every `Spawn` overloads.

On the contrary, using an `out` parameter in `TrySpawn` is perfectly fine.

iRebbok: It won't work in our case. They will conflict with deprecated methods that return `void`.

galaxy119: ^ that. Which was why I told him to make them out parameters, so they can be overloads, otherwise these become breaking changes, and i don't particularly see why they should be. We can always revise this when we introduce other breaking changes, which will also be when we remove all of the [Obsolete] methods, so there would be no more conflict.

iRebbok: Yeah. It's like choosing between `static void Method(out int item1, out int item2)` and `static (int, int) Method()`, both are good, but the second one seems to me better, it's more flexible.

iopietro: Ok, if we're forced, then it's fine.

iopietro (approved):

PR #466, opened 2021-03-21
Cleaning a little bit the example plugin

xRoier:
- Deleted the private constructor.
- Moved down ``base.OnEnabled()`` and ``base.OnDisabled()``.

### DIFF
--- a/Exiled.Example/Example.cs
+++ b/Exiled.Example/Example.cs
@@ -42,11 +42,11 @@ namespace Exiled.Example
         public override void OnEnabled()
         {
             RegisterEvents();
-            
+
             Log.Warn($"I correctly read the string config, its value is: {Config.String}");
             Log.Warn($"I correctly read the int config, its value is: {Config.Int}");
             Log.Warn($"I correctly read the float config, its value is: {Config.Float}");
-            
+
             base.OnEnabled();
         }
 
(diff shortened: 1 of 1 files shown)


Exiled.Example/Example.cs
```diff
@@ -41,21 +37,18 @@ private Example()
         /// <inheritdoc/>
         public override void OnEnabled()
         {
-            base.OnEnabled();
-
             RegisterEvents();
-
```
iRebbok, line 47: Revert that like breaker. It looks better, you don't get confused in logs and method calls with a large number of logs.

Exiled.Example/Example.cs
```diff
@@ -41,21 +37,18 @@ private Example()
...
             Log.Warn($"I correctly read the string config, its value is: {Config.String}");
             Log.Warn($"I correctly read the int config, its value is: {Config.Int}");
             Log.Warn($"I correctly read the float config, its value is: {Config.Float}");
+            base.OnEnabled();
```
iRebbok, line 44: Add a line breaker above that line.

iRebbok (changes requested):
The private constructor used here to tell to Exiled to use one from a static property instead. Revert deletion.

iRebbok (approved):

iRebbok:
Ignore CI, it fails due to a certificate issue on www.exiled.host.

iopietro (approved):

PR #468, opened 2021-03-24
Add ChangedRole event

PintTheDragon:
There is an issue with ChangingRole that causes it to run many times, with IsAllowed and IsEscaped set to true if IsAllowed is set to false in Escaping event. The Escaping event is the only way to accurately view escapes, but it is not possible to do the same with role changes. This new event occurs after a role is changed, to allow responding to the event correctly.

### DIFF
--- a/Exiled.Events/EventArgs/ChangedRoleEventArgs.cs
+++ b/Exiled.Events/EventArgs/ChangedRoleEventArgs.cs
@@ -21,9 +21,9 @@ namespace Exiled.Events.EventArgs
         /// Initializes a new instance of the <see cref="ChangedRoleEventArgs"/> class.
         /// </summary>
         /// <param name="player"><inheritdoc cref="Player"/></param>
-        /// <param name="newRole"><inheritdoc cref="NewRole"/></param>
+        /// <param name="oldRole"><inheritdoc cref="OldRole"/></param>
         /// <param name="items"><inheritdoc cref="Items"/></param>
-        /// <param name="shouldPreservePosition"><inheritdoc cref="ShouldPreservePosition"/></param>
+        /// <param name="preservedPosition"><inheritdoc cref="PreservedPosition"/></param>
         /// <param name="isEscaped"><inheritdoc cref="IsEscaped"/></param>
         public ChangedRoleEventArgs(Player player, RoleType oldRole, List<ItemType> items, bool preservedPosition, bool isEscaped)
         {
--- a/Exiled.Events/EventArgs/ChangedRoleEventArgs.cs
+++ b/Exiled.Events/EventArgs/ChangedRoleEventArgs.cs
@@ -1,0 +1,62 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="ChangedRoleEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+    using System.Collections.Generic;
+
+    using Exiled.API.Features;
+
+    /// <summary>
+    /// Contains all informations after player's <see cref="RoleType"/> changes.
+    /// </summary>
+    public class ChangedRoleEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ChangedRoleEventArgs"/> class.
+        /// </summary>
+        /// <param name="player"><inheritdoc cref="Player"/></param>
+        /// <param name="newRole"><inheritdoc cref="NewRole"/></param>
+        /// <param name="items"><inheritdoc cref="Items"/></param>
+        /// <param name="shouldPreservePosition"><inheritdoc cref="ShouldPreservePosition"/></param>
+        /// <param name="isEscaped"><inheritdoc cref="IsEscaped"/></param>
+        public ChangedRoleEventArgs(Player player, RoleType oldRole, List<ItemType> items, bool preservedPosition, bool isEscaped)
+        {
+            Player = player;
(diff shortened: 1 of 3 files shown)


galaxy119:
It would be better to build in a fix for the issue you described, rather than let it continue to be broken and adding in another event unecessarily.

PintTheDragon:
I'm not sure how a fix could be implemented though. We have two events that are both able to cancel the changing role from running. The best way to do it would be to completely remove Escaping event entirely, and I'm not quite sure why it exists because the same information is also included in Changing role. The issue is that plugins are using Escaping event, but could probably be updated in time if the Escaping event was removed because like I said, it can easily be changed to the Changing Role event.

Exiled.Events/Patches/Events/Player/ChangingRole.cs
```diff
@@ -70,12 +70,17 @@ private static bool Prefix(CharacterClassManager __instance, ref RoleType classi
                     classid = escapingEventArgs.NewRole;
                 }
 
+                var oldRole = API.Features.Player.Get(ply).Role;
```
iRebbok, line 73: Get the player once instead, don't call `Player::Get(GameObject)` every time you need a value of a property.

iRebbok (changes requested):

iRebbok (approved):
In my opinion, I like that addition, every event is a gold bar 🙏. But, here's my text wall after I looked at the patch's code: We call `OnDied` when the new role is Spectator, lol? We fake the event args aswell, no matter why, I see it's not well, the player doesn't get hurt, just a plugin updates their role. The feature, you @PintTheDragon mentioned, is not an issue in itself, that's standard behavior: you prevent the client from escaping, it call the RCP again; if a plugin roughly deals with those events, then it'll fail with a StackOverflowException.

PintTheDragon:
Yeah, I also find it quite strange that died event is ran there.

PintTheDragon:
Looks like it's in 56300b1fad52220ceba4c641a13b8d67575a7613. @joker-119 , what's that about lol?

Exiled.Events/Patches/Events/Player/ChangingRole.cs
```diff
@@ -70,12 +72,17 @@ private static bool Prefix(CharacterClassManager __instance, ref RoleType classi
                     classid = escapingEventArgs.NewRole;
                 }
 
+                var oldRole = player.Role;
+
                 ply.GetComponent<CharacterClassManager>().SetClassIDAdv(classid, lite, escape);
                 ply.GetComponent<PlayerStats>().SetHPAmount(__instance.Classes.SafeGet(classid).maxHP);
                 ply.GetComponent<FirstPersonController>().ResetStamina();
```
iopietro, line 80: Just define one single `changedRoleEventArgs` variable here.

Exiled.Events/EventArgs/ChangedRoleEventArgs.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether the player escaped or not.
+        /// </summary>
+        public bool IsEscaped { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the position has been preserved after changing the role.
+        /// </summary>
+        public bool PreservedPosition { get; }
```
iopietro, line 60: This (and `hasPreservedPosition` argument) should be named `HasPreservedPosition` , to be more consistent to `IsEscaped` name.

Exiled.Events/Patches/Events/Player/ChangingRole.cs
```diff
@@ -160,8 +159,7 @@ private static bool Prefix(CharacterClassManager __instance, ref RoleType classi
 
                 ListPool<Inventory.SyncItemInfo>.Shared.Return(list);
 
-                var changedRoleEventArgs1 = new ChangedRoleEventArgs(player, oldRole, startItemsList, false, escape);
-                Player.OnChangedRole(changedRoleEventArgs1);
+                Player.OnChangedRole(changedRoleEventArgs);
```
iopietro, line 162: Remember to set IsEscaped to false.
You sholdn't instantiate `ChangedRoleEventArgs` immediately.
You should instantiate a new `ChangedRoleEventArgs` here with IsEscaped set to false and above inside the `if (lite)` with IsEscaped set to true, we'll only instantiate one object anyway.

Exiled.Events/Patches/Events/Player/ChangingRole.cs
```diff
@@ -72,15 +72,14 @@ private static bool Prefix(CharacterClassManager __instance, ref RoleType classi
                     classid = escapingEventArgs.NewRole;
                 }
 
-                var oldRole = player.Role;
+                var changedRoleEventArgs = new ChangedRoleEventArgs(player, player.Role, startItemsList, lite, escape);
```
iopietro, line 75: > You sholdn't instantiate ChangedRoleEventArgs immediately.

I'm referring to this line.

Exiled.Events/Patches/Events/Player/ChangingRole.cs
PintTheDragon: Lite doesn't set isescaped to true. It set's has position override to true, or whatever it is. The only difference between the two initializations is that if lite is true, the it's set to true, otherwise it's set to false, which is what's being done here.

PintTheDragon: I don't think there's a real issue with initializing it immediately. It saves 2 lines that would be used for the initialization before running the event, and it's never going to go to waste. Plus, nothing inside of the events is modified after the initialization.

iopietro: It's fine then, I misunderstood one of your previous commit.

iopietro (approved):

PR #470, opened 2021-03-29
Fix GetGroup

PintTheDragon:
SL clones UserGroups for no real reason and it breaks Exiled.Permissions and any other plugins that need to get the key/name of a UserGroup. This adds a simple patch on the only method that clones UserGroups to not clone the UserGroup.


iRebbok:
Why? If you need, you get any from `PermissionHandler::_groups`, so you modified the method for no reason, which might break in-game stuff in the future. Exiled.Permissons doesn't even use that method, we directly use `PermissionsHandler._groups`.

PintTheDragon:
I think your misunderstanding the point of these changes. The setgroup command ingame uses the patched function to set a user's group. The issue is that because it's a clone, it doesn't appear in serverRoles, and do Exiled.Permissions doesn't give the person permission to do things they should have access to. The same goes for other plugins as well. Setgroup makes no modifications to the UserGroup, so there's no real reason for it to be cloned. All it does is break plugins.

PintTheDragon:
Essentially, when you setgroup someone, they don't get EXILED permissions, because there isn't any way to find the name of the group that's set. This fixes that.

iopietro:
Does this fix the bug that removes every EXILED permissions when you run `pm reload`?

PintTheDragon:
I don't think so, since when I looked only setgroup uses the patched method. I'll take a look at it and fix it if it is a similar issue.

iopietro:
Nice, it would be cool if you fix that bug aswell.

PintTheDragon:
I'm not quite sure what the issue with pm is, but I looked a bit and it seems unrelated to what this PR fixes.

iRebbok:
> Essentially, when you setgroup someone, they don't get EXILED permissions, because there isn't any way to find the name of the group that's set. This fixes that.

But still you can compare objects themselves, not their referenced, I don't see the reason for the patch.

iRebbok:
> Does this fix the bug that removes every EXILED permissions when you run `pm reload`?

Just took a closer look at the `pm reload` command, it refreshes the `ServerStatic::PermissionHandler`, so all the `Group` entities are refreshed aswell. Speaking of that, the solution I mentioned above works well in this case.

iopietro:
> Speaking of that, the solution I mentioned above works well in this case.

Do you mean comparing objects instead of their reference?

iRebbok:
> > Speaking of that, the solution I mentioned above works well in this case.
> 
> Do you mean comparing objects instead of their reference?

Sure.

iRebbok:
Lemme know if I missed something in the projects. I don't really remember where we compare `UserGroup`'s.

iopietro (approved):

Exiled.API/Extensions/UserGroupExtension.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// Contains a useful extension to compare two <see cref="UserGroup"/>'s.
+    /// </summary>
+    public static class UserGroupExtension
+    {
+        /// <summary>
+        /// Compares two <see cref="UserGroup"/>'s for equality.
+        /// </summary>
+        /// <param name="this">The fist <see cref="UserGroup"/>.</param>
+        /// <param name="other">The second <see cref="UserGroup"/>.</param>
+        /// <returns><c>true</c> if they are equal; otherwise, false.</returns>
```
iRebbok, line 20:
```suggestion
        /// <returns><c>true</c> if they are equal; otherwise, <c>false<c>.</returns>
```

BuildBoy12 (approved):

PR #471, opened 2021-03-30
MirrorExtesnions::ChangeRunningSpeed & MirrorExtesnions:ChangeWalkingSpeed methods

Michal78900:
So I decided to add `canBeSent` to all of Log messages, because using only Debug makes thing harder to read in console if every message is green.

The second thing I added are methods to Change Player's both walking and running speed. I know it can cause issus with AC at higher speeds so may I should cap it at some max multiplier value.


iRebbok:
Firstly, that's the wrong way, you should never do it, if you need to do so, you add an IF statement to your code; secondly, this will break all plugins calling those methods (all plugins do it while enabling/disabling and etc, lol). I'm literally disagree with your logic, if you need to read a lot of logs, you do it from a log file, not the console.

Michal78900:
But `canLog` is by default set to `true`. Even with that will it break anything?

Michal78900:
If it does I'm sorry I didn't know that addind one if statement breaks entire framework. What about 2 methods that I added?

iRebbok:
It does, the IL code for those methods has changed, it requires emitted an additional `ldc.i4.1` to call.

iopietro (approved):

Exiled.API/Extensions/MirrorExtensions.cs
```diff
@@ -152,6 +154,34 @@ public static void ChangeAppearance(this Player player, RoleType type)
         /// <param name="makeNoise">Same on <see cref="Cassie.Message(string, bool, bool)"/>'s isNoisy.</param>
         public static void PlayCassieAnnouncement(this Player player, string words, bool makeHold = false, bool makeNoise = true) => SendFakeTargetRpc(player, RespawnEffectsController.AllControllers.First().netIdentity, typeof(RespawnEffectsController), nameof(RespawnEffectsController.RpcCassieAnnouncement), words, makeHold, makeNoise);
 
+        /// <summary>
+        /// Change <see cref="Player"/> walking speed. Negative values will invert player's controls.
```
iRebbok, line 158:
```suggestion
        /// Changes the <see cref="Player"/>'s walking speed. Negative values will invert the player's controls.
```

Exiled.API/Extensions/MirrorExtensions.cs
```diff
@@ -152,6 +154,34 @@ public static void ChangeAppearance(this Player player, RoleType type)
...
+        public static void ChangeWalkingSpeed(this Player player, float multiplier, bool useCap = true)
+        {
+            if (useCap)
+                multiplier = Mathf.Clamp(multiplier, -2f, 2f);
+
+            SendFakeSyncVar(player, ServerConfigSynchronizer.Singleton.netIdentity, typeof(ServerConfigSynchronizer), nameof(ServerConfigSynchronizer.Singleton.NetworkHumanWalkSpeedMultiplier), multiplier);
+        }
+
+        /// <summary>
+        /// Change <see cref="Player"/> running speed. Negative values will invert player's controls.
```
iRebbok, line 172:
```suggestion
        /// Changes the <see cref="Player"/>'s running speed. Negative values will invert the player's controls.
```

iRebbok (approved):

PR #472, opened 2021-04-01
Fix a typo

PintTheDragon:
I noticed there was a typo in the SEXILED source. Unfortunately, this typo seems to also extend to the repository name, which needs to be changed after merge. I'd like to get this merged as soon as possible so that as few plugins need to be recompiled as possible, so I've skipped some stylecop steps that can be safely changed later. Finally, the welcome message has been updated to support all consoles.

Also, sorry about all the commits. I've been having an issue building on my computer and the CI shows only some of the errors every build.


Parkeymon:
this fixes major typos, merge it now

iopietro (approved):
Thank you for fixing so many typos! I don't know how we haven't noticed them for so long!

PR #475, opened 2021-04-03
Added Player.HasHint and Player.Zone

Raul125:
A new feature that I think would be good for exiled.


Raul125:
Closed, because i improved the code.

PR #476, opened 2021-04-03
Added Player.HasHint and Player.Zone

Raul125:
A new feature that I think would be good for exiled. (created another PR because i improved the code ._.)


iopietro:
The branch you're using is wrong, it must be `dev` and not `master`.

Raul125:
> The branch you're using is wrong, it must be `dev` and not `master`.

Whoops

PR #479, opened 2021-04-04
mini fix ShouldReloadConfigsAtRoundRestart config

sanyae2439:
* add Description
* set default

### DIFF
--- a/Exiled.Events/Config.cs
+++ b/Exiled.Events/Config.cs
@@ -51,7 +51,8 @@ namespace Exiled.Events
         /// <summary>
         /// Gets or sets a value indicating whether configs has to be reloaded every time a round restarts.
         /// </summary>
-        public bool ShouldReloadConfigsAtRoundRestart { get; set; }
+        [Description("Indicates whether configs has to be reloaded every time a round restarts")]
+        public bool ShouldReloadConfigsAtRoundRestart { get; set; } = true;
 
         /// <summary>
         /// Gets a value indicating whether bans should be logged or not.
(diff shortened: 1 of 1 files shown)


Exiled.Events/Config.cs
```diff
@@ -51,7 +51,8 @@ public sealed class Config : IConfig
         /// <summary>
         /// Gets or sets a value indicating whether configs has to be reloaded every time a round restarts.
         /// </summary>
-        public bool ShouldReloadConfigsAtRoundRestart { get; set; }
+        [Description("Indicates whether configs has to be reloaded every time a round restarts")]
+        public bool ShouldReloadConfigsAtRoundRestart { get; set; } = false;
```
iopietro, line 55: There's no need to explicitly set it to false, it's already set to false by default if you don't specify it.

Exiled.Events/Config.cs
```diff
@@ -51,7 +51,8 @@ public sealed class Config : IConfig
         /// <summary>
         /// Gets or sets a value indicating whether configs has to be reloaded every time a round restarts.
         /// </summary>
-        public bool ShouldReloadConfigsAtRoundRestart { get; set; }
+        [Description("Indicates whether configs has to be reloaded every time a round restarts")]
```
iRebbok, line 54:
```suggestion
        [Description("Indicates whether configs have to be reloaded every round restart")]
```

Exiled.Events/Config.cs
sanyae2439: other configs are well specify, so i think this should also be unified.

iopietro (approved):

iRebbok (approved):

PR #480, opened 2021-04-05
Add player.HasHint, player.Zone and player.ArtificialHealthDecay

Raul125:
A new feature that I think would be good for exiled. (now with dev branch, my 🧠 is small .-. )


Exiled.API/Features/Player.cs
```diff
@@ -781,6 +786,11 @@ public string GroupName
         /// </summary>
         public Room CurrentRoom => Map.FindParentRoom(GameObject);
 
+        /// <summary>
+        /// Gets the current zone the player is in.
+        /// </summary>
+        public ZoneType Zone => Map.FindParentRoom(GameObject).Zone;
```
iopietro, line 792: You should just call `CurrentRoom.Zone`.

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+#pragma warning disable SA1600 // Elements should be documented
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+        // Creating a list for coroutine check
+        private static List<Player> plyHaveCoroutine = new List<Player>();
+
+        public static void Postfix(HintDisplay __instance, Hint hint)
+        {
+            // Try to get the player, if it doesn't exist, just return
+            Player player = Player.Get(__instance.gameObject);
+            if (player == null)
```
iopietro, line 32: You should use a shorter form: `if (!(Player.Get(__instance.gameObject) is Player player))`

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
```diff
@@ -0,0 +1,69 @@
+// -----------------------------------------------------------------------
+// <copyright file="PlayerHasHintPatch.cs" company="Exiled Team">
```
iopietro, line 2: You don't need to use the `Patch` suffix.

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            // Set HasHint to true, and then delays to false for the duration of the hint
+            player.HasHint = true;
+
+            // Set HintDisplayTime to the duration of the last hint
+            // This prevents to set HasHint to false before the hint has finished
+            player.HintRemainingTime = hint.DurationScalar;
+
+            // If not contains, run coroutine
+            if (!plyHaveCoroutine.Contains(player))
```
iopietro, line 43: Hints don't queue up like broadcasts, they overwrite eachother.
This check would fail when a plugin shows a hint for x seconds, then another one immediately after with a duration lower or higher than x.

So, if that list already contains a coroutine, you should kill it and run a new one.

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// Patches <see cref="HintDisplay.Show(Hint)"/>.
+    /// </summary>
+    [HarmonyPatch(typeof(HintDisplay), nameof(HintDisplay.Show))]
+    public static class PlayerHasHintPatch
+    {
+#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
+#pragma warning disable SA1600 // Elements should be documented
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+        // Creating a list for coroutine check
+        private static List<Player> plyHaveCoroutine = new List<Player>();
```
iopietro, line 26: You should call it `playerHasHintCoroutines` and use a `HashSet` instead of a `List`, since you call both `Contains` and `Remove`.

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public static IEnumerator<float> HasHintToFalse(Player player)
+        {
+            // if player hint remaining time is less than 0, stop the coroutine
+            while (player.HintRemainingTime > 0)
+            {
+                yield return Timing.WaitForSeconds(1f);
+                player.HintRemainingTime -= 1;
+                if (player.HintRemainingTime == 0 || player.HintRemainingTime < 0)
```
iopietro, line 57: You should use `if (player.HintRemainingTime <= 0)`.

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        public static IEnumerator<float> HasHintToFalse(Player player)
+        {
+            // if player hint remaining time is less than 0, stop the coroutine
+            while (player.HintRemainingTime > 0)
+            {
+                yield return Timing.WaitForSeconds(1f);
+                player.HintRemainingTime -= 1;
```
iopietro, line 56: Here you should check if `player.GameObject` is null and break the coroutine consequently, to not throw exceptions if a player quits before viewing the hint for the whole duration.

Exiled.API/Features/Player.cs
```diff
@@ -116,6 +116,11 @@ private set
         /// </summary>
         public AmmoBox Ammo { get; private set; }
 
+        /// <summary>
+        /// Gets a value indicating whether gets if the player is viewing a hint.
```
iRebbok, line 120:
```suggestion
        /// Gets a value indicating whether or not the player is viewing a hint.
```

Exiled.Events/Patches/Generic/PlayerHasHintPatch.cs
Raul125: The check would not fail because I was using HintRemainingTime (my 🧠 went brrrr) and that's neither effective nor optimized, or i think so, but when you told me this I came up with another more effective idea.

Exiled.Events/Patches/Generic/PlayerHasHint.cs
```diff
@@ -17,50 +17,55 @@ namespace Exiled.Events.Patches.Generic
...
+            // If it is false, then to true
+            if (!player.HasHint)
+                player.HasHint = true;
+
+            // I have to create a variable to be able to change the value of coroutine tag
+            CoroutineHandle coroutine = Timing.RunCoroutine(HasHintToFalse(player, hint.DurationScalar));
+
+            // Change the value of coroutine tag to the userid of the player
+            // Add HasHint part to ensure the tag is unique
+            coroutine.Tag = player.UserId + "HasHint";
```
iopietro, line 55: This is fine, but you could even use a `Dictionary<Player, CoroutineHandle>` to simplify the code and then, above, you could just use `playerHasHintCoroutines.TryGetValue(...)` to get the coroutine and kill it.

Raul125: Done, now using Dictionary, and i have added Player.ArtificialHealthDecay.

Exiled.Events/Patches/Generic/PlayerHasHint.cs
```diff
@@ -23,44 +23,43 @@ public static class PlayerHasHint
...
+            if (playerHasHintCoroutines.TryGetValue(player, out CoroutineHandle oldcoroutine))
             {
-                // Kill player coroutine
-                Timing.KillCoroutines(player.UserId + "HasHint");
+                // Kill the coroutine
+                Timing.KillCoroutines(oldcoroutine);
+            }
+
+            // If not contains add it and run the couroutine
+            if (!playerHasHintCoroutines.ContainsKey(player))
```
iopietro, line 42: You don't need to check it twice with `TryGetValue` and `ContainsKey?`

iopietro (approved):

joker-119 (approved):
I'd prefer if the patch was a transpiler, but otherwise looks fine to me.

PR #498, opened 2021-05-18
Fixed not setting correctly item in player's hand

grzes0071:
Fixed issue where after setting item and then getting it, it returned item that was previously in hand


joker-119 (approved):
lgtm

BuildBoy12 (approved):

iopietro (approved):

PR #505, opened 2021-06-04
Exiled.Events - Cassie

Jesus-QC:
Adds C.A.S.S.I.E. related events


Jesus-QC:
# Cassie - Exiled Events

This will help devs to use cassie related events, one example could be a TTS that replaces cassie, this would get the message, play the message with the TTS and returning ev.IsAllowed as false.

Jesus-QC:
I don't see any error, idk why the build is failing :(

Jesus-QC:
Oh, i see, lol
image

Exiled.Events/Patches/Cassie/SendingMessage.cs
```diff
@@ -0,0 +1,42 @@
...
+                var ev = new SendingMessageEventArgs(words, makeHold, makeNoise);
+                Handlers.Cassie.OnSendingMessage(ev);
+
+                words = ev.Words;
+                makeHold = ev.MakeHold;
+                makeNoise = ev.MakeNoise;
+
+                return ev.IsAllowed;
+            }
+            catch (Exception e)
```
iopietro, line 35: Since there's no base-game code, this try-catch is not needed, there's already one inside [InvokeSafely](https://github.com/Exiled-Team/EXILED/blob/cd4b256834bb78fe4d1ce6f5d1a8fb2dbd2546ee/Exiled.Events/Extensions/Event.cs#L35).

iopietro (changes requested):

iopietro (approved):

Exiled.Events/EventArgs/SendingMessageEventArgs.cs
```diff
@@ -0,0 +1,52 @@
...
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+
+    /// <summary>
+    /// Contains all the information after sending a C.A.S.S.I.E. message.
+    /// </summary>
+    public class SendingMessageEventArgs : EventArgs
```
iRebbok, line 15: Wouldn't a closer name be better? E.g., if we decide to add an event for the admin chat (messages starting with `@` in RA)...

Jesus-QC: Renamed it to SendingCassieMessage

iRebbok:
To ensure you approve the changes.

iopietro (approved):
I agree with you.

BuildBoy12 (approved):

PR #507, opened 2021-06-07
ChangingMicroHIDState and UsingMicroHIDEnergy events

Michal78900:
I've added 2 new events:
`ChangingMicroHIDState` - it is called when the MicroHID changes it's state, for example from idle to rolling
`UsingMicroHIDEnergy` - it is called when the MicroHID uses it's energy

### DIFF
--- a/Exiled.Events/EventArgs/ChangingMicroHIDStateEventArgs.cs
+++ b/Exiled.Events/EventArgs/ChangingMicroHIDStateEventArgs.cs
@@ -1,0 +1,54 @@
+// -----------------------------------------------------------------------
+// <copyright file="ChangingMicroHIDStateEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+
+    using Exiled.API.Features;
+
+    /// <summary>
+    /// Contains all information before MicroHID state is changed.
+    /// </summary>
+    public class ChangingMicroHIDStateEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ChangingMicroHIDStateEventArgs"/> class.
+        /// </summary>
+        /// <param name="player"><inheritdoc cref="Player"/></param>
+        /// <param name="oldState"><inheritdoc cref="OldState"/></param>
+        /// <param name="newState"><inheritdoc cref="NewState"/></param>
+        /// <param name="isAllowed"><inheritdoc cref="IsAllowed"/></param>
+        public ChangingMicroHIDStateEventArgs(Player player, MicroHID.MicroHidState oldState, MicroHID.MicroHidState newState, bool isAllowed = true)
+        {
+            Player = player;
+            OldState = oldState;
+            NewState = newState;
+            IsAllowed = isAllowed;
+        }
+
+        /// <summary>
+        /// Gets the player who's using the MicroHID.
+        /// </summary>
+        public Player Player { get; }
+
+        /// <summary>
+        /// Gets the old MicroHID state.
+        /// </summary>
+        public MicroHID.MicroHidState OldState { get; }
+
+        /// <summary>
+        /// Gets or sets the new MicroHID state.
+        /// </summary>
+        public MicroHID.MicroHidState NewState { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the MicroHID state can be changed or not.
+        /// </summary>
+        public bool IsAllowed { get; set; }
+    }
+}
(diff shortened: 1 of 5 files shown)


Exiled.Events/EventArgs/ChangingMicroHIDStateEventArgs.cs
```diff
@@ -0,0 +1,54 @@
...
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+
+    using Exiled.API.Features;
+
+    /// <summary>
+    /// Contains all informations before MicroHID state is changed.
```
PintTheDragon, line 15: Change `informations` to `information`.

Exiled.Events/EventArgs/UsingMicroHIDEnergyEventArgs.cs
```diff
@@ -0,0 +1,61 @@
...
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+
+    using Exiled.API.Features;
+
+    /// <summary>
+    /// Contains all informations before MicroHID energy is changed.
```
PintTheDragon, line 15: Same as above, `informations` to `informations`.

Exiled.Events/Handlers/Player.cs
```diff
@@ -310,6 +310,16 @@ public static class Player
         /// </summary>
         public static event CustomEventHandler<ChangingRadioPresetEventArgs> ChangingRadioPreset;
 
+        /// <summary>
+        /// Invoked before a user's MicroHID state is changed.
```
PintTheDragon, line 314: Change `user` to `player`.

Exiled.Events/Handlers/Player.cs
```diff
@@ -310,6 +310,16 @@ public static class Player
...
+        /// </summary>
+        public static event CustomEventHandler<ChangingMicroHIDStateEventArgs> ChangingMicroHIDState;
+
+        /// <summary>
+        /// Invoked before a user's MicroHID energy is changed.
```
PintTheDragon, line 319: Same as above.

Exiled.Events/Handlers/Player.cs
```diff
@@ -651,5 +661,17 @@ public static class Player
         /// </summary>
         /// <param name="ev">The <see cref="ChangingRadioPresetEventArgs"/> instance.</param>
         public static void OnChangingRadioPreset(ChangingRadioPresetEventArgs ev) => ChangingRadioPreset.InvokeSafely(ev);
+
+        /// <summary>
+        /// Called before a user's MicroHID state is changed.
```
PintTheDragon, line 666: Same as above.

Exiled.Events/Handlers/Player.cs
```diff
@@ -651,5 +661,17 @@ public static class Player
...
+        /// </summary>
+        /// <param name="ev">The <see cref="ChangingRadioPresetEventArgs"/> instance.</param>
+        public static void OnChangingMicroHIDState(ChangingMicroHIDStateEventArgs ev) => ChangingMicroHIDState.InvokeSafely(ev);
+
+        /// <summary>
+        /// Called before a user's MicroHID energy is changed.
```
PintTheDragon, line 672: Same as above.

PintTheDragon (changes requested):
Looks good, although there are a few naming issues.

iopietro (approved):

Exiled.Events/Patches/Events/Player/ChangingMicroHIDState.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Adds the <see cref="Handlers.Player.OnChangingMicroHIDState"/> event.
+    /// </summary>
+    [HarmonyPatch(typeof(MicroHID), nameof(MicroHID.NetworkCurrentHidState), MethodType.Setter)]
+    internal static class ChangingMicroHIDState
+    {
+        private static bool Prefix(MicroHID __instance, ref MicroHID.MicroHidState value)
+        {
+            // NetworkCurrentHid state is set each frame, so this is to prevent calling the method each frame.
+            if (__instance.CurrentHidState == value)
+                return true;
```
iRebbok, line 28: In .NET property setters not ever called when assigning the same value 👀

Michal78900: ~~Northwood moment~~
Without this, the method is called each frame so maybe it isn't a property fault as you said, but something else.
Anyways without this `if` the event doesn't work as it should.

Michal78900:
Shit I messed up and deleted this PR

Michal78900:
Nevermind, you saw nothing :trollface:

Exiled.Events/Patches/Events/Player/ChangingMicroHIDState.cs
iRebbok: 👌 then keep it up

iopietro: > In .NET property setters not ever called when assigning the same value 👀
Aren't they? I tested it pretty quickly and they seem to be called even when assigning to them the same value.

iRebbok: > Aren't they? To be 100% sure I tested it and they get called even when assigning to them the same value.

Dunno, that was an old thread

BuildBoy12 (approved):

PR #508, opened 2021-06-12
Added NetworkforceStopInputs

NaoUnderscore:
I don't know how to rename it, I think it might be useful tho.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -908,7 +908,7 @@ namespace Exiled.API.Features
         /// Gets player's items.
         /// </summary>
         public Inventory.SyncListItemInfo Items => Inventory.items;
-        
+
         /// <summary>
         /// Gets or sets a value indicating whether or not the player can send inputs.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Player.cs
```diff
@@ -908,7 +908,16 @@ public int Scp330Usages
         /// Gets player's items.
         /// </summary>
         public Inventory.SyncListItemInfo Items => Inventory.items;
-
+        
+        /// <summary>
+        /// Gets or sets a value indicating whether or not the player can send inputs.
+        /// </summary>
+        public bool NetworkforceStopInputs
```
PintTheDragon, line 915: This name's kinda gross. Maybe go with something more descriptive like `CanSendInputs`.

PintTheDragon (changes requested):

Exiled.API/Features/Player.cs
Nekonyx: Or `CanMove`

NaoUnderscore: Yeah I think you're right. `CanSendInputs` is fine.

iopietro (changes requested):
Fix build errors.

NaoUnderscore:
I’ll fix it today.

BuildBoy12 (approved):

iopietro (approved):

PR #512, opened 2021-06-18
Updated Player::CurrentRoom for tinybrain people

Monaldcry7788:
*Done with the phone, I may have been wrong*

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1472,6 +1472,16 @@ namespace Exiled.API.Features
         public void RemoveHands()
         {
         }
+        
+        /// <summary>
+        /// Teleport the player to a room
+        /// </summary>
+        /// <param name="roomType">The room when player will be teleported</param>
+        public void SetPlayerRoom(RoomType roomType)
+        {
+            var room = Map.Rooms.Where(x => x.Type == roomType).FirstOrDefault();
+            ReferenceHub.playerMovementSync.OverridePosition(room.Position + Vector3.up * 1.5f, 0f);
+        }
 
         /// <inheritdoc/>
         public override string ToString() => $"{Id} {Nickname} {UserId} {Role} {Team}";
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -782,7 +782,22 @@ namespace Exiled.API.Features
         public Room CurrentRoom
         {
             get => Map.FindParentRoom(GameObject);
-            set => ReferenceHub.playerMovementSync.OverridePosition(value.Position + Vector3.up * 1.5f, 0f);
+            set
+            {
+                PlayerMovementSync.FindSafePosition(value.Position, out Vector3 safePos);
+                switch (value.Type)
+                {
+                    case RoomType.HczTesla:
+                        ReferenceHub.playerMovementSync.ForceSafePosition(safePos + (Vector3.back * 4f));
+                        break;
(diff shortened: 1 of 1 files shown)


iopietro:
Just add a setter to [Player::CurrentRoom](https://github.com/Exiled-Team/EXILED/blob/cd4b256834bb78fe4d1ce6f5d1a8fb2dbd2546ee/Exiled.API/Features/Player.cs#L810).

iRebbok:
> Just add a setter to [Player::CurrentRoom](https://github.com/Exiled-Team/EXILED/blob/cd4b256834bb78fe4d1ce6f5d1a8fb2dbd2546ee/Exiled.API/Features/Player.cs#L810).

It'll force devs to find an instance of `Room`, could a new property called `CurrentRoomType` or something be good to address this?

Monaldcry7788 (commented):
Updated Player::CurrentRoom for sets the current player's room

Remuchi:
Setting player position to room.gameobject.position will not work properly for some rooms. For example, 939 chamber (you will spawn in air, fall down and die), 079 room (you spawn between walls), collapsed tunnels in entrance zone etc 
Don't know about surface

iopietro:
> It'll force devs to find an instance of Room, could a new property called CurrentRoomType or something be good to address this?
Yes, you're right, what do you think about having bot `CurrentRoom` and `CurrentRoomType`? 
In my opinion the first one could still be useful for devs who search for rooms directly from `Map.Rooms`.

> Setting player position to room.gameobject.position will not work properly for some rooms. For example, 939 chamber (you will spawn in air, fall down and die), 079 room (you spawn between walls), collapsed tunnels in entrance zone etc
Don't know about surface.

@Monaldcry7788 Because of that, you should use `PlayerMovementSync::FindSafePosition`, which is a base game method.

iRebbok:
> Yes, you're right, what do you think about having bot CurrentRoom and CurrentRoomType?
In my opinion the first one could still be useful for devs who search for rooms directly from Map.Rooms.

Sure!

Monaldcry7788:
Updated

4310V343k:
Chill, chill! It's only one line!

<details>
<summary></summary>



</details>

Monaldcry7788:
Removed white line at L1194

Exiled.API/Features/Player.cs
```diff
@@ -777,9 +777,28 @@ public string GroupName
...
+        /// Gets or sets the current room the player is in.
         /// </summary>
-        public Room CurrentRoom => Map.FindParentRoom(GameObject);
+        public Room CurrentRoom
+        {
+            get => Map.FindParentRoom(GameObject);
+            set
+            {
+                PlayerMovementSync.FindSafePosition(value.Position, out Vector3 safePos);
+                switch (value.Type)
```
joker-119, line 788: There needs to be more problematic rooms added to this check, for example: 173's room, 079's room, 939's room, etc. You also need to take into account that these rooms can be facing 4 different directions along the x axis, Vector3.back may not always be the safest direction for a particular room.

joker-119 (changes requested):
More rooms need to be added to the switch check, for ex, 173's room will put you in a wall, 079's put you inside the cage, 939's puts you in the middle of the air, etc.

Also you need to keep in mind that all rooms can be facing 4 different directions along the x axis, Vector3.back (negative z) may not always be the direction of 'safety', I'd recommend using the room's transform instead of the static Vector3

PR #515, opened 2021-06-28
Add ConfigVector3

PintTheDragon:
Because Vector3s don't actually work in config. This just wraps around it and generates a value when requested. Can be used like:
```csharp
public ConfigVector3 Position {get; set;} = new ConfigVector3(1, 1, 1);
```
Then just
```csharp
Config.Position.Value
```

### DIFF
--- a/Exiled.API/Features/ConfigVector3.cs
+++ b/Exiled.API/Features/ConfigVector3.cs
@@ -1,0 +1,76 @@
+// -----------------------------------------------------------------------
+// <copyright file="ConfigVector3.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Features
+{
+    using UnityEngine;
+
+    using YamlDotNet.Serialization;
+
+    /// <summary>
+    /// A helper struct used for storing Vector3s in config.
+    /// </summary>
+    public struct ConfigVector3 
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConfigVector3"/> struct.
+        /// </summary>
+        /// <param name="x">The x coordinate.</param>
+        /// <param name="y">The y coordinate.</param>
+        /// <param name="z">The z coordinate.</param>
+        public Badge(float x, float y, float z)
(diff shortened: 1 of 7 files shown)


PintTheDragon:
Someone decided to hide this in CustomItems API for some reason so I'm just gonna move it.

PintTheDragon:
Now normally I wouldn't do this, but since there commits are going to be squashed...

iRebbok:
Why don't you like this boy? 
https://github.com/Exiled-Team/EXILED/blob/c90154dd9e4ae46a1de6ae2647a57e5ac65e4d8e/Exiled.CustomItems/API/Features/Vector.cs#L17-L19
It's been in our codebase since the first release, and it works pretty well...

PintTheDragon:
It isn't too related to CustomItems and is a bit hidden. Unless someone is doing an item, they probably won't see it.

My other problem is that Vector, which is a wrapper for config, has a lot of overrides for vector math, even though it's purpose is for putting vectors in config. Now Joker said this might be to save those ms because it's value isn't cached, which is another issue with it. It recreates the Vector3 every time, even though Vector3 is a struct and can be safely cached.

iRebbok:
Are you speaking about saving some CPU ticks? Now think how your change appears on the heap and the same ticks. Caching a struct allocates it on the heap using a class-wrapper that's also exists on the heap isn't a way I might approve. The purpose of using structs is an allocation on the stack which is faster than the heap. Considering that the game uses an old version of Unity (actually 2019.4), and that it's an old version of Mono, your change increases GC pauses.

Back on topic, didn't you think it could be a struct, or it could have overridden operators, or you could just move `Vector` to `Exiled.API.Features`? I believe someone else will come and tell his dev story about how it'd be cool to have all those overrides.

joker-119:
When the Vector struct was first created, it was specifically for CustomItems, which is why it was in it's API and not the general EXILED one, where it should have been.

I originally made it a struct because stack allocation better than heap allocation, but as a struct it can't internally cache the result of the 'new Vector3()' required to convert it to a Vector3.

I won't pretend to know what the performance difference between keeping it a struct but having to make a new Vector3 everytime .ToVector3 is called, vs having it on the heap but being able to cache the Vector3 it creates.

PintTheDragon:
Closing in favor of #519.

PR #516, opened 2021-06-29
Adds FirstAvailable as an InfoSide for CreditTags

0b10000:
This PR adds FirstAvailable as an InfoSide for CreditTags. Basically it finds the first unused spot to put the credit tag and puts it there. This avoids the downfalls of the old Both option, as it only puts it in one spot.

I did make this the default config option, please let me know if you have an objection to that.


0b10000:
Changed to a draft until I fix an issue with it overriding hidden by default tags

PR #517, opened 2021-06-30
Add FirstAvailable as an InfoSide for CreditTags

0b10000:
This PR adds FirstAvailable as an InfoSide for CreditTags. It finds the first unused spot to put the credit tag and then puts it there. 

I did change the default config option to FirstAvailable, as I believe it provides the most flexibility without any config changes by the server owner.

### DIFF
--- a/Exiled.CreditTags/Enums/InfoSide.cs
+++ b/Exiled.CreditTags/Enums/InfoSide.cs
@@ -20,6 +20,11 @@ namespace Exiled.CreditTags.Enums
         CustomPlayerInfo,
 
         /// <summary>
+        /// Uses Badge if available, otherwise uses CustomPlayerInfo if available
+        /// </summary>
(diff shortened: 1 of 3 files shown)


joker-119 (approved):

Exiled.CreditTags/Enums/InfoSide.cs
```diff
@@ -19,6 +19,11 @@ public enum InfoSide
         /// </summary>
         CustomPlayerInfo,
 
+        /// <summary>
+        /// Uses Badge if available, otherwise uses CustomPlayerInfo if available
```
iRebbok, line 23:
```suggestion
        /// Uses Badge if available, otherwise uses CustomPlayerInfo if available.
```

iRebbok (approved):

BuildBoy12 (approved):

PR #525, opened 2021-07-04
Converted DamagingWindow into a transpiler

xRoier:
After a whole day with headache learning .NET Framework based IL I made my first transpiler 😎


joker-119 (approved):
LGTM

I can't currently look at the game-code to determine if the index/offset are correct, but I'm assuming roier has tested this change and that it works.

iRebbok (approved):

PR #527, opened 2021-07-06
Add LockingDown Event

Remuchi:
Adds a SCP-079 Lockdown event (when scp-079 turns off the lights in room and locks the doors in it)

### DIFF
--- a/Exiled.Events/EventArgs/LockdowningEventArgs.cs
+++ b/Exiled.Events/EventArgs/LockdowningEventArgs.cs
@@ -1,0 +1,54 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="LockdowningEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+    using Exiled.API.Features;
+    using UnityEngine;
+
+    /// <summary>
+    /// Contains all informations before SCP-079 lockdowns a room.
+    /// </summary>
+    public class LockdowningEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LockdowningEventArgs"/> class.
+        /// </summary>
+        /// <param name="player"><inheritdoc cref="Player"/></param>
+        /// <param name="roomGameObject"><inheritdoc cref="RoomGameObject"/></param>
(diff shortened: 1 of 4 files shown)


Exiled.Events/Handlers/Scp079.cs
```diff
@@ -126,6 +131,12 @@ public static class Scp079
         /// <param name="ev">The <see cref="ElevatorTeleportingEventArgs"/> instance.</param>
         public static void OnElevatorTeleporting(ElevatorTeleportingEventArgs ev) => ElevatorTeleporting.InvokeSafely(ev);
 
+        /// <summary>
+        /// Called before SCP-079 lockdowns a room.
+        /// </summary>
+        /// <param name="ev">The <see cref="LockdowningEventArgs"/> instance.</param>
+        public static void OnLockdowning(LockdowningEventArgs ev) => Lockdowning?.Invoke(ev);
```
Michal78900, line 138: Shouldn't this be `Lockdowning.InvokeSafely(ev);`?

Exiled.Events/EventArgs/LockdowningEventArgs.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Gets the player who's controlling SCP-079.
+        /// </summary>
+        public Player Player { get; }
+
+        /// <summary>
+        /// Gets the gameobject of the room that will be lockdowned.
+        /// </summary>
+        public GameObject RoomGameObject { get; }
```
Michal78900, line 42: I think it would be better to make this also a setter (being able to lockdown completely different room, some dev would definetly use that) and getting an actual `Room` object instead of `GameObject`.
I'm still learning IL code, so I can't help you with that, but I think it would be possible.

Remuchi: The only way to get room from gameobject as far as I know is Map.Rooms.First(room => room.gameobject = gameobject) although I don't think this is a good way of doing it

joker-119:
While probably just a nitpick, I think the event should be called LockingDown not Lockdowning

Exiled.Events/EventArgs/LockdowningEventArgs.cs
Michal78900: I think that gameObject.GetComponentInParent<Room>() would also work, but I understand the problem that it may be inefficient.

joker-119 (approved):

iopietro:
I agree with @joker-119, `LockingDown` is better.

iopietro (approved):

PR #530, opened 2021-07-07
Fixed Player.CanSendInputs being inverted

Michal78900:
Because the method uses ReferenceHub.fpc.Networkforce**Stop**Inputs; setting `player.CanSendInputs` to `true` actually means forcing stopping inputs and vice versa.

Yes, the fix is literally adding two "!!".


joker-119 (approved):
lgtm

BuildBoy12 (approved):

iopietro (approved):

PR #513, opened 2021-06-23
graceful exceptions

iRebbok:
Another bulky non-tested PR...

I know many of us (probably everyone except @sanyae2439 😅) don't fuck up with mono & dnSpy to make debugging a dll easier.
This PR doesn't include a debugger but makes all possible exceptions graceful (see https://github.com/benaadams/Ben.Demystifier#make-error-logs-more-productive).

Not sure if this will help anyone except server owners with erroring plugins by reducing the size of the logs (probably).

PR #531, opened 2021-07-08
Sending a helpful message, when an exception happens

Michal78900:
# Changes:
- Added sending helpful messages to a user, when he gots an exception which can led to solving a problem without needing to ping developers. (right know this PR supports 4 exceptions)
- Removed try-catch from all event patches, because InvokeSafely() has one and it was unnecessary to have them.
- Fixed `Exiled.Loader` bug with checking only the Major version

#513 PR was added, but it will be helpful probably only for developers.

This is still WIP, but I think draft pull request is for that, am I right?

### DIFF
--- a/Exiled.API/Features/ExceptionHandler.cs
+++ b/Exiled.API/Features/ExceptionHandler.cs
@@ -9,7 +9,8 @@ namespace Exiled.API.Features
 {
     using System;
     using System.IO;
-    using System.Linq;
+
+    using YamlDotNet.Core;
 
     /// <summary>
     /// .
@@ -25,21 +26,21 @@ namespace Exiled.API.Features
         {
             string message = null;
 
-            switch (exception.GetType().Name)
+            switch (exception)
             {
-                case "MissingMethodException":
+                case MissingMethodException _:
                     message = "This error can be fixed by updating Exiled or plugin to the latest version.";
                     break;
 
-                case "FileNotFoundException":
-                    message = $"The \"{(exception as FileNotFoundException).FileName.Split(new char[] { ',' })[0]}.dll\" file is either missing or it outdated.";
+                case FileNotFoundException fileNotFound:
+                    message = $"The \"{fileNotFound.FileName.Split(new char[] { ',' })[0]}.dll\" file is either missing or is outdated.";
                     break;
 
-                case "YamlException":
+                case YamlException _:
                     message = $"Check the line stated above. If you don't know what is the issue parse the .yml file through YAML Validator for example through this one: https://codebeautify.org/yaml-validator";
                     break;
 
-                case "BadImageFormatException":
+                case BadImageFormatException _:
                     message = $"The one or more .dll files are corrupt. Please reinstall them.";
                     break;
 
@@ -54,10 +55,9 @@ namespace Exiled.API.Features
             }
             else
             {
-                #if DEBUG
+#if DEBUG
                 Log.Send(exception.GetType().Name, Discord.LogLevel.Warn, ConsoleColor.DarkYellow);
-                #endif
-
+#endif
                 return false;
             }
         }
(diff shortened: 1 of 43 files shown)


Exiled.API/Features/ExceptionHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+                case MissingMethodException _:
+                    message = "This error can be fixed by updating Exiled or plugin to the latest version.";
+                    break;
+
+                case FileNotFoundException fileNotFound:
+                    message = $"The \"{fileNotFound.FileName.Split(new char[] { ',' })[0]}.dll\" file is either missing or is outdated.";
+                    break;
+
+                case YamlException _:
+                    message = $"Check the line stated above. If you don't know what is the issue parse the .yml file through YAML Validator for example through this one: https://codebeautify.org/yaml-validator";
```
joker-119, line 40: We shouldn't recommend the site to use, since it is not a site we control, we cannot guarantee it will exist, or function properly, or that it doesn't/won't have potentially malicious content (the last thing we want to do is send server owners to a website that's been compromised and give them a virus). Instead, simply state that one should use a YAML validator.

Exiled.API/Features/ExceptionHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+                    break;
+
+                case BadImageFormatException _:
+                    message = $"The one or more .dll files are corrupt. Please reinstall them.";
```
joker-119, line 44: "The dll file stated above is likely corrupted, this can usually be fixed by re-downloading it."

Exiled.Loader/Loader.cs
```diff
@@ -347,23 +352,22 @@ private static bool CheckPluginRequiredExiledVersion(IPlugin<IConfig> plugin)
...
                 {
-                    Log.Error($"You're running an older version of Exiled ({Version.ToString(3)})! {plugin.Name} won't be loaded! " +
-                              $"Required version to load it: {plugin.RequiredExiledVersion.ToString(3)}");
+                    Log.Error($"You're running an older version of Exiled ({Version})! {plugin.Name} won't be loaded! " +
+                              $"Required version to load it: {plugin.RequiredExiledVersion}");
 
                     return true;
                 }
-                else if (requiredVersion.Major < actualVersion.Major && !Config.ShouldLoadOutdatedPlugins)
+                else if (requiredVersion < actualVersion && !Config.ShouldLoadOutdatedPlugins)
```
joker-119, line 367: This check can remain a .Major check only, as older plugins requiring older versions of EXILED are *typically* save to load so long as the EXILED major version is the same. The only real issues from mismatched EXILED versions come if the plugin requires a newer version of EXILED than is on the server, as things get added only change the .Minor, if something is removed completely, the .Major version will always have to change.

Exiled.Loader/Loader.cs
```diff
@@ -347,23 +352,22 @@ private static bool CheckPluginRequiredExiledVersion(IPlugin<IConfig> plugin)
...
                 {
-                    Log.Error($"You're running an older version of {plugin.Name} ({plugin.Version.ToString(3)})! " +
-                              $"Its Required Major version is {requiredVersion.Major}, but excepted {actualVersion.Major}. ");
+                    Log.Error($"You're running an older version of {plugin.Name} ({plugin.Version})! " +
```
joker-119, line 369: This error is painfully in need of some grammar checking, it should be:

"{plugin.Name} requires EXILED version {requiredVersion}, however your current version ({actualVersion}) is significantly newer. This may cause issues with the plugin, so it will not be loaded. To change this, set should_load_outdated_plugins to true."

Exiled.Loader/Loader.cs
```diff
@@ -404,6 +408,7 @@ private static void LoadDependencies()
             catch (Exception exception)
             {
                 Log.Error($"An error has occurred while loading dependencies! {exception}");
+                ExceptionHandler.TrySendingHelpMessage(exception);
```
joker-119, line 411: This needs to be removed. This method is called before Exiled.API.dll is loaded, this this method will not be accessible here, and will cause a MissingMethodException when trying to call the method, which will break dependency and plugin loading completely.

joker-119 (changes requested):

joker-119:
Ty <3, looks good so far :D

PR #532, opened 2021-07-09
Added ForceCuff check in Player.IsCuffed

Lufou:
Added the ForceCuff (RA) check in Player.IsCuffed bool

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -376,7 +376,7 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets a value indicating whether or not the player is cuffed.
         /// </summary>
-        public bool IsCuffed => CufferId != -1;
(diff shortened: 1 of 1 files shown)


iopietro (approved):

joker-119 (approved):

joker-119 (changes requested):
This PR should be made to the dev branch, not master.

joker-119 (approved):

PintTheDragon:
Keep in mind that this can introduce breaking changes. For example, SCPStats uses this to determine if a player is cuffed by a player (https://github.com/SCPStats/Plugin/blob/8ca04cbb4817bc09599433c181688bddec7e66db/SCPStats/EventHandler.cs#L283). @joker-119, will this be included in 2.11? If so, I'll need to update my code to work with it.

PR #538, opened 2021-07-10
Fixed description of Shot event

TeamEXAngus:
It's wrong and it was a pain in the ass for me


joker-119 (approved):
PR was made against wrong branch, only actual edit is Exiled.Events/Handlers/Player.cs L#179

Looks fine, I guess?

iopietro:
This event isn't fired only when a player gets shot, but every time a bullet hits something.

TeamEXAngus:
> This event isn't fired only when a player gets shot, but every time a bullet hits something.

When I was testing this event, it only triggered when players got shot, also this is what I got told in dev talk on the discord

iopietro:
Mmh ok, then it's strange.

BuildBoy12 (approved):

PR #544, opened 2021-07-12
Added Map.SpawnGrenade() method

Michal78900:
For some reason, this still wasn't a thing. The same method exists in `Exiled.CustomItems` but it isn't accessible, because you can use it only from a `CustomGrenade` class.

I wanted to add `explodeOnCollision` but because `Exiled.CustomItems` is inaccessible from `Exiled.API` I can't use a collider component from it.

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -511,6 +511,9 @@ namespace Exiled.API.Features
         /// <returns>The <see cref="Grenade"/> being spawned.</returns>
         public static Grenade SpawnGrenade(Vector3 position, GrenadeType grenadeType = GrenadeType.FragGrenade, float fuseTime = 3f, Vector3? velocity = null, bool explodeOnCollision = false, Player player = null)
         {
+            if (!Enum.IsDefined(typeof(GrenadeType), grenadeType))
+                return null;
+
             if (player == null)
                 player = Server.Host;
 
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -508,7 +508,7 @@ namespace Exiled.API.Features
         /// <param name="grenadeType">The <see cref="GrenadeType"/> of the grenade to spawn.</param>
         /// <param name="player">The <see cref="Player"/> to count as the thrower of the grenade.</param>
         /// <returns>The <see cref="Grenade"/> being spawned.</returns>
-        public static Grenade Spawn(Vector3 position, Vector3 velocity, float fuseTime = 3f, GrenadeType grenadeType = GrenadeType.FragGrenade, Player player = null)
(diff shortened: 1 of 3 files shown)


joker-119:
It would be possible to move the collision component from Exiled.CustomItems to Exiled.API, to allow you to do that, since Exiled.CustomItems relies on the API already, moving it wouldn't be a bad thing afaik, since Exiled.API will always be loaded first before anything else.

Exiled.API/Features/Map.cs
```diff
@@ -493,6 +499,33 @@ public static void ChangeUnitColor(int index, string color)
...
+        /// </summary>
+        /// <param name="position">The <see cref="Vector3"/> to spawn the grenade at.</param>
+        /// <param name="velocity">The <see cref="Vector3"/> directional velocity the grenade should move at.</param>
+        /// <param name="fuseTime">The <see cref="float"/> fuse time of the grenade.</param>
+        /// <param name="grenadeType">The <see cref="GrenadeType"/> of the grenade to spawn.</param>
+        /// <param name="player">The <see cref="Player"/> to count as the thrower of the grenade.</param>
+        /// <returns>The <see cref="Grenade"/> being spawned.</returns>
+        public static Grenade SpawnGrenade(Vector3 position, Vector3? velocity = null, float fuseTime = 3f, GrenadeType grenadeType = GrenadeType.FragGrenade, Player player = null)
+        {
+            if (player == null)
```
iopietro, line 513: I would add a check here, to verify whether the grenadeType is valid or not.

iopietro (changes requested):

Exiled.API/Features/Map.cs
Michal78900: How can it be invalid? All 3 possible values are valid and are working.

Michal78900: I don't think you can set it to `null`

Michal78900:
> It would be possible to move the collision component from Exiled.CustomItems to Exiled.API, to allow you to do that, since Exiled.CustomItems relies on the API already, moving it wouldn't be a bad thing afaik, since Exiled.API will always be loaded first before anything else.

I'll probably do that

Exiled.API/Features/Map.cs
iopietro: Using [this](https://docs.microsoft.com/en-US/dotnet/api/system.enum.tryparse?view=net-5.0).

Michal78900: I'm sorry, but I really don't know what you mean. Like, checking the `grenadeType` via `Enum.TryParse()`?? How does parsing the already parsed thing make any sense? I'm probably misunderstood what you meant.

iopietro: You've to parse it and check if it's valid because someone could just do something like `(GrenadeType)999` and it would be counted as a valid enum, even if it doesn't exist.

iopietro (approved):

joker-119 (approved):

PR #546, opened 2021-07-13
Fixed serialization/deserialization of vector configs.

iopietro:


joker-119:
What does this actually fix? I haven't noticed an issue with them since the last change to allow for them.

PintTheDragon:
Well, it isn't very clear, but essentially:
Serializer serializes floats in US format.
Deserializer deserializes into the current culture.

So, it gets written as 2.1, but it expects 2,1. This fixes that.

iopietro:
Like @PintTheDragon said, it uses the default serialization and deserialization of YamlDotNet, which is based on the US format.

BuildBoy12 (approved):

Issue #547, opened 2021-07-15
EXILED's banned event gives a null issuer if the issuer's name contains unsafe characters


PintTheDragon:
Basically, SL stores the ban issuer nickname for a ban, which is fine for getting a player from it, but it also replaces unsafe characters. If a player has a nickname containing unsafe characters, the issuer will be null because the nickname has been changed.

My suggestion for fixing this is to turn the Banned patch into a transpiler, doing something along the lines of this:
```
var oldIssuer = ban.Issuer;
... original method code ...
... current Banned patch postfix code, using oldIssuer instead of ban.Issuer ...
```

PintTheDragon:
Never mind fuck transpilers I'm writing a prefix.

PintTheDragon:
Closing, as a pr (#549) has been merged to fix this

PR #548, opened 2021-07-15
Make the Banned event work with unsafe issuer nicknames

PintTheDragon:
More info on #547, but essentially the issuer is null if their nickname has unsafe characters.


PintTheDragon:
Ah shit ok I'll just redo the pr.

PR #554, opened 2021-07-19
Added Ragdoll class

NaoUnderscore:
> Added `Ragdoll` _class_.

> Removed `Ragdoll` _extensions_.

> Moved **Ragdoll** _methods_ from `Map` _class_ to `Ragdoll` _class_.

> Moved **Ragdoll** _property_ from `Map` _class_ to `Ragdoll` _class_.


Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
```diff
@@ -43,6 +43,11 @@ private static bool Prefix(RagdollManager __instance, ref Vector3 pos, ref Quate
                 ownerNick = ev.PlayerNickname;
                 playerId = ev.PlayerId;
 
+                Role role = __instance.hub.characterClassManager.Classes.SafeGet(classId);
+                GameObject gameObject = UnityEngine.Object.Instantiate(role.model_ragdoll, pos + role.ragdoll_offset.position, Quaternion.Euler(rot.eulerAngles + role.ragdoll_offset.rotation));
```
joker-119, line 47: Doing this will Firstly, cause the server to instantiate a new ragdoll gameobject even if the event's ev.IsAllowed is set to false.

Secondly, it will cause this event to always instantiate 2 ragdoll game objects, however only one will be spawned into the map. Additionally, the one that is properly spawned in, won't be on your Map.Ragdolls list.

This event needs to be reworked significantly if you are going to accomplish what you think it is you need to be doing here.

Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
```diff
@@ -43,6 +43,11 @@ private static bool Prefix(RagdollManager __instance, ref Vector3 pos, ref Quate
...
+
+                Exiled.API.Features.Map.RagdollsValue.Add(new API.Features.Ragdoll(gameObject.GetComponent<Ragdoll>()));
```
joker-119, line 49: You create a new ragdoll, but never gave it the ragdoll component, therefore GetComponent<Ragdoll>() will return null.

joker-119 (changes requested):

Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
NaoUnderscore: Alright. I'm gonna do that.

Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
NaoUnderscore: I'm gonna rework this event prob.

Exiled.Events/Patches/Events/Player/SpawningRagdoll.cs
```diff
@@ -43,12 +43,22 @@ private static bool Prefix(RagdollManager __instance, ref Vector3 pos, ref Quate
...
                 playerId = ev.PlayerId;
 
+                if (!ev.IsAllowed)
+                    return false;
+
                 Role role = __instance.hub.characterClassManager.Classes.SafeGet(classId);
+                if (role.model_ragdoll == null)
+                    return false;
                 GameObject gameObject = UnityEngine.Object.Instantiate(role.model_ragdoll, pos + role.ragdoll_offset.position, Quaternion.Euler(rot.eulerAngles + role.ragdoll_offset.rotation));
+                Ragdoll component = gameObject.GetComponent<Ragdoll>();
```
joker-119, line 53: You are still instantiating a new GameObject, and then running a GetComponent<> without that componend ever being added anywhere.

Have you tested this and made sure that this GetComponent<> isn't null?

NaoUnderscore: I've tested it and everything works. Tried to enable/disable/get ragdoll informations with commands in game.
I'll add a null-check today/tomorrow, but since I spawn the ragdoll using the API.Features.Ragdoll class, the GameObject can't be null, or it wouldn't be spawned.

joker-119 (approved):

Exiled.API/Features/Ragdoll.cs
```diff
@@ -0,0 +1,337 @@
...
+    using System.Linq;
+
+    using UnityEngine;
+
+    /// <summary>
+    /// A set of tools to handle the ragdolls more easily.
+    /// </summary>
+    public class Ragdoll
+    {
+        private readonly global::Ragdoll ragdoll;
```
PintTheDragon, line 19: Nice on all of the properties/methods, but if any new features are added and EXILED isn't changed, you'll have to either use reflection or getcomponents to access the underlying ragdoll class.

I believe that, as a rule, we should definitely try to abstract away most of the basegame interactions, but we should not prevent them. Keeping the underlying class private serves no real purpose, other than to make it harder to interact with basegame code, which we should NOT be doing. Every one of these wrapper classes should try its best to fully support what it's wrapping, but it shouldn't prevent plugins from accessing it.

PintTheDragon (changes requested):

Exiled.API/Features/Ragdoll.cs
NaoUnderscore: Could a `Player.Ragdoll`make sense? You'd be able to access it easily.

PintTheDragon: Can't players have more than one Ragdoll? Also, I don't think it makes sense to put access to the ragdoll on the Player class instead of the Ragdoll class.

NaoUnderscore: That's true

joker-119: The classes reference to the base-game equivalent should be a public property, incase anyone ever needs to access it directly for something that the API alone cannot do.

PintTheDragon:
I'll approve once the underlying Ragdoll class becomes publicly accessible.

BuildBoy12 (approved):

PR #555, opened 2021-07-19
Added setter to Player::MoveState

NaoUnderscore:


Michal78900:
How that supposed to work?

iopietro (approved):

joker-119:
I don't think changing this has no lasting effect, as the client re-sends it's current state every frame. Can you verify that you've tested having the server alter this value and that it actually works as expected?

joker-119 (approved):

PR #556, opened 2021-07-19
Added Player.Heal()

NaoUnderscore:


PintTheDragon:
What's the difference between this and `p.Health += 10`? Is it capped at the max health?

NaoUnderscore:
Yes, it is.
You can check through dnSpy. 
SetHPAmount() is not capped at the max health, therefore I used HealHPAmount.

iopietro (approved):

Exiled.API/Features/Player.cs
```diff
@@ -1224,6 +1224,12 @@ public bool HasItem(ItemType targetItem)
         /// <param name="reason">The disconnection reason.</param>
         public void Disconnect(string reason = null) => ServerConsole.Disconnect(GameObject, string.IsNullOrEmpty(reason) ? string.Empty : reason);
 
+        /// <summary>
+        /// Heals the player.
```
joker-119, line 1228: Since this is limited to only healing them up to their max health, the summary should specify this to avoid confusion.

NaoUnderscore: Check it now and lemme know if it's ok.

joker-119 (approved):

PR #557, opened 2021-07-19
Added some properties to Round class

NaoUnderscore:


iopietro (approved):

joker-119:
Aren't these unset until the round ends? Or do I just remember their loop incorrectly?

NaoUnderscore:
private IEnumerator<float> _ProcessServerSideCode() is exectued while (RoundSummary != null) and it keeps updating all these values with local vars.

joker-119 (approved):

iopietro (approved):

PR #564, opened 2021-07-29
Allows Custom Items to be spawned inside lockers.

joker-119:
This PR adds the ability for Custom Items to use a "InsideLocker" dynamic spawn location. This will pick a random locker on the map, and replace the item that would normally spawn inside the locker with the custom item.

### DIFF
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -468,9 +468,38 @@ namespace Exiled.CustomItems.API.Features
 
                 spawned++;
 
-                Spawn(spawnPoint.Position.ToVector3, out _);
+                if (spawnPoint is DynamicSpawnPoint dynamicSpawnPoint && dynamicSpawnPoint.Location == SpawnLocation.InsideLocker)
+                {
+                    for (int i = 0; i < 50; i++)
+                    {
+                        Locker locker =
+                            LockerManager.singleton.lockers[
+                                Instance.Rng.Next(LockerManager.singleton.lockers.Length)];
+                        if (locker._itemsToSpawn == null)
+                        {
+                            Log.Debug($"{nameof(Spawn)}: Invalid locker location. Attempting to find a new one..", Instance.Config.Debug);
+                            continue;
+                        }
 
-                Log.Debug($"Spawned {Name} at {spawnPoint.Position} ({spawnPoint.Name})", Instance.Config.Debug);
+                        LockerChamber chamber = locker.chambers[Instance.Rng.Next(Mathf.Max(0, locker.chambers.Length - 1))];
+
+                        foreach (Locker.ItemToSpawn item in locker._itemsToSpawn.ToList())
+                        {
+                            if (item._pos == chamber.spawnpoint.position)
+                                locker._itemsToSpawn.Remove(item);
+                        }
+
+                        Vector3 position = chamber.spawnpoint.transform.position;
+                        Spawn(position, out _);
+                        Log.Debug($"Spawned {Name} at {position} ({spawnPoint.Name})", Instance.Config.Debug);
+                    }
+                }
+                else
+                {
+                    Spawn(spawnPoint.Position.ToVector3, out _);
+
+                    Log.Debug($"Spawned {Name} at {spawnPoint.Position} ({spawnPoint.Name})", Instance.Config.Debug);
+                }
             }
 
             return spawned;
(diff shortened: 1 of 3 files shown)


Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -468,9 +468,38 @@ public virtual uint Spawn(IEnumerable<SpawnPoint> spawnPoints, uint limit)
 
                 spawned++;
 
-                Spawn(spawnPoint.Position.ToVector3, out _);
+                if (spawnPoint is DynamicSpawnPoint dynamicSpawnPoint && dynamicSpawnPoint.Location == SpawnLocation.InsideLocker)
+                {
+                    for (int i = 0; i < 50; i++)
```
iopietro, line 473: Why looping from `0` to `49`?

Exiled.CustomItems/CustomItems.cs
```diff
@@ -25,6 +25,7 @@ public class CustomItems : Plugin<Config>
     {
         private static readonly CustomItems Singleton = new CustomItems();
 
+        private Random rng = new Random();
```
iopietro, line 28: Use `Exiled.Loader::Random` or `UnityEngine::Random`.

Exiled.CustomItems/CustomItems.cs
```diff
@@ -39,6 +40,11 @@ private CustomItems()
         /// </summary>
         public static CustomItems Instance => Singleton;
 
+        /// <summary>
+        /// Gets a random number generator.
+        /// </summary>
+        internal Random Rng => rng;
```
iopietro, line 46: Same as above.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -468,9 +468,38 @@ public virtual uint Spawn(IEnumerable<SpawnPoint> spawnPoints, uint limit)
...
+                    {
+                        Locker locker =
+                            LockerManager.singleton.lockers[
+                                Instance.Rng.Next(LockerManager.singleton.lockers.Length)];
```
iopietro, line 477: You should directly use `Exiled.Loader::Random` or `UnityEngine::Random`.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -468,9 +468,38 @@ public virtual uint Spawn(IEnumerable<SpawnPoint> spawnPoints, uint limit)
...
+                        {
+                            Log.Debug($"{nameof(Spawn)}: Invalid locker location. Attempting to find a new one..", Instance.Config.Debug);
+                            continue;
+                        }
+
+                        LockerChamber chamber = locker.chambers[Instance.Rng.Next(Mathf.Max(0, locker.chambers.Length - 1))];
+
+                        foreach (Locker.ItemToSpawn item in locker._itemsToSpawn.ToList())
+                        {
+                            if (item._pos == chamber.spawnpoint.position)
```
iopietro, line 488: Why do you prevent normal items from spawning if they're in the same location of the custom ones?

iopietro (changes requested):

Exiled.CustomItems/API/Features/CustomItem.cs
joker-119: There are lockers spawned outside of the map, that are given Locker components, but do not have spawnable item chambers. Trying to spawn an item in one of these results in a nullref, so we need to check for that null value, and if it's found, repeat the random selection to find another locker. 

The loop prevents us from looping indefinitely, by limiting the spawning attempts to 50 tries before 'giving up' and not spawning the locker item.

Exiled.CustomItems/CustomItems.cs
joker-119: Didn't remember Exiled.Loader::Random existed, will fix

Exiled.CustomItems/API/Features/CustomItem.cs
joker-119: Because then you would end up with more than 1 item in a small locker, which can cause physics issues that cause one (or sometimes both) item(s) to bounce out of the locker container.

By small locker, I am referring to the big cabinets of small, locked locker chambers that normally contain items like keycards, ammo and grenades.

Exiled.CustomItems/API/Features/CustomItem.cs
BuildBoy12: So, essentially, spawning the item **up to** 50 items, but the majority of the time not hitting 50 due to funky lockers. I assume it hits enough 'funky lockers' for the 50 attempts to not become an issue of over-spawning?

joker-119: Ok so I'm dumb and forgot a break; at the end. It's supposed to loop until it finds a place to spawn it, looping a max of 50 attempts before giving up.

BuildBoy12 (approved):

PR #568, opened 2021-08-06
CustomItems Commands Improvements

SebasCapo:
**Couple of small QoL changes:**

- **"ci give" command's player identifier is now optional:**
Using the give command without providing an identifier will choose the sender as the player who will receive the CustomItem, we're all too lazy to write our own name over and over again. (If an identifier was specified but was invalid, it will still throw a syntax error response)

- **~~You can now use the give command on SCPs:~~**  Reverted
~~While SCPs can't normally open their inventory, if you fake-change a syncvar using Mirror so the client thinks they're a human class, they will be considered SCPs, but be able to use their inventory normally, as well as their items. (That and also currently in the beta SCPs are able to access their items through hotkeys, but that might change in the future so I'm not considering it as a real reason)~~

- **Registered list gets now ordered by IDs:**
It just looks a lot less like a mess, currently if you wanna have it ordered, you must register the items in the order you want it to show up, if you ask me to revert this I will fight you.

I dedicate this video to y'all dev bois
https://www.youtube.com/watch?v=5xooMXyleXM

### DIFF
--- a/Exiled.CustomItems/Commands/Give.cs
+++ b/Exiled.CustomItems/Commands/Give.cs
@@ -16,6 +16,8 @@ namespace Exiled.CustomItems.Commands
     using Exiled.CustomItems.API.Features;
     using Exiled.Permissions.Extensions;
 
+    using RemoteAdmin;
+
     /// <summary>
     /// The command to give a player an item.
     /// </summary>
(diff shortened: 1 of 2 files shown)


Exiled.CustomItems/Commands/Give.cs
```diff
@@ -96,7 +119,7 @@ public bool Execute(ArraySegment<string> arguments, ICommandSender sender, out s
         /// </summary>
         private bool CheckEligible(Player player)
         {
-            return player.IsAlive && !player.IsScp && !player.IsCuffed && player.Inventory.items.Count < 8;
```
joker-119, line 99: This check should remain. You should never send an SCP player a fake syncvar making their client think they are human, as their client will expect to be able to do all human interactions that the server will ignore completely if the server thinks they are an SCP. Instead, you change the player's role to a human one, and then use player.ChangeApperance to make everyone elses client think they are still an SCP.

joker-119 (changes requested):

Exiled.CustomItems/Commands/Give.cs
SebasCapo: Fixed on latest commit

BuildBoy12 (approved):

joker-119 (approved):

PR #572, opened 2021-08-11
Add a property to disable spawn of an CustomItem's pickup when owner escapes

Remuchi:
Currently Exiled spawns a pickup of CustomItem when player escapes with it inside his inventory but sometimes you don't need this  to happen. Now you can set a property **ShouldSpawnPickupOnEscape** to false to disable spawn of pickup on escaping.

### DIFF
--- a/Exiled.CustomItems/API/Features/CustomItem.cs
+++ b/Exiled.CustomItems/API/Features/CustomItem.cs
@@ -66,6 +66,11 @@ namespace Exiled.CustomItems.API.Features
         public virtual float Durability { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether a pickup will be spawned when player escapes or not.
+        /// </summary>
+        public virtual bool ShouldSpawnPickupOnEscape { get; set; } = true;
+
+        /// <summary>
         /// Gets or sets the ItemType to use for this item.
         /// </summary>
         public virtual ItemType Type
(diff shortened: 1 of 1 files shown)


Remuchi:
For some reason, It doesn't work. I've tried doing **ev.Player.RemoveItem(ev.Item)** with **ev.IsAllowed** set to false and true and both variants apparently did nothing - item still appeared after escaping

PintTheDragon:
Does GetRandomSpawnPoint return the player's position, or just somewhere around them?

Remuchi:
Somewhere around them

joker-119:
The only main issue I see with this change is, because items are config objects, every server using Exiled.CustomItems will get all of their item configs regenerated, which iirc does not preserve existing values. If someone can verify for me that the existing config options are saved when it has to regenerate a missing one, i'll approve this.

PintTheDragon:
> 
> 
> The only main issue I see with this change is, because items are config objects, every server using Exiled.CustomItems will get all of their item configs regenerated, which iirc does not preserve existing values. If someone can verify for me that the existing config options are saved when it has to regenerate a missing one, i'll approve this.

Does it do things differently from how plugin configs are handled? I know that plugin configs can do something like this, but I'm not sure how customitems handles it.

joker-119:
Depends on the individual plugins, my CustomItems plugin does the same thing as base-game, but I can't speak for others, most probably just add the configs directly into the normal config class

joker-119:
Closing due to uncertainty about becoming a breaking change, and locking of dev branch in preparation for game update.

Issue #574, opened 2021-08-12
Players escaping with custom items don't get their items spawned in the same position as other items


PintTheDragon:
Title. This might be a bit difficult to fix, but essentially it places them in a random spawnpoint, which'll be near them, but not below their feet. See https://github.com/Exiled-Team/EXILED/pull/572#issuecomment-897413322.

PR #578, opened 2021-08-13
Make CustomItems spawn where the player is after escaping

joker-119:
Addresses #574


PintTheDragon:
What about patching the method that actually clears the player's inventory, and then spawns the items. You could prefix ~~or transpiler~~ it then remove your item first.

joker-119:
I could but it isn't really necessary

sanyae2439 (approved):

PR #579, opened 2021-08-14
Fix XML docs for InternalTranslation

PintTheDragon:
I don't wanna talk about it.


sanyae2439 (approved):
I don't wanna talk about it too.

joker-119 (approved):
kekw

PR #581, opened 2021-08-14
Exiled.Network

Killers0992:
Way of connecting or sending data thru scpsl servers via just 1 plugin. examples idk where i can put but [Link](https://github.com/Killers0992/NetworkedPlugins) aka how that works.


EXILED.sln
```diff
@@ -21,9 +21,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Exiled.Bootstrap", "Exiled.
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Exiled.Patcher", "Exiled.Patcher\Exiled.Patcher.csproj", "{B92E9671-71E6-4DDE-B085-062065110D0A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Exiled.CustomItems", "Exiled.CustomItems\Exiled.CustomItems.csproj", "{B7FBA3C1-6182-4E96-A33B-053EDDCC4F65}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Exiled.CustomItems", "Exiled.CustomItems\Exiled.CustomItems.csproj", "{B7FBA3C1-6182-4E96-A33B-053EDDCC4F65}"
```
PintTheDragon, line 24: Why'd you change this?

Exiled.Network/API/Attributes/NPAddonInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Network.API.Attributes
+{
+    using System;
+
+    /// <summary>
+    /// Info about addon.
```
PintTheDragon, line 13: Info about the addon.

Exiled.Network/API/Attributes/NPAddonInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    public class NPAddonInfo : Attribute
+    {
+        /// <summary>
+        /// Gets or sets AddonID.
```
PintTheDragon, line 18: Gets or sets the AddonID.

Exiled.Network/API/Attributes/NPAddonInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public string AddonID { get; set; }
+
+        /// <summary>
+        /// Gets or sets AddonName.
```
PintTheDragon, line 23: Gets or sets the AddonName.

Exiled.Network/API/Attributes/NPAddonInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public string AddonName { get; set; }
+
+        /// <summary>
+        /// Gets or sets AddonVersion.
```
PintTheDragon, line 28: Gets or sets the AddonVersion.

Exiled.Network/API/Attributes/NPAddonInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public Version AddonVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets AddonAuthor.
```
PintTheDragon, line 33: Gets or sets the AddonAuthor.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+// -----------------------------------------------------------------------
+
+namespace Exiled.Network.API.Interfaces
+{
+    using System;
+    using System.Collections.Generic;
+    using Exiled.Network.API.Models;
+    using LiteNetLib.Utils;
+
+    public interface IAddon<out TConfig> : IComparable<IAddon<IConfig>>
```
PintTheDragon, line 15: Add a summary.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        where TConfig : IConfig
+    {
+        /// <summary>
+        /// Gets or sets network manager.
```
PintTheDragon, line 19: Gets or sets the network manager.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        NPManager Manager { get; set; }
+
+        /// <summary>
+        /// Gets or sets logger.
```
PintTheDragon, line 24: Gets or sets the logger.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        NPLogger Logger { get; set; }
+
+        /// <summary>
+        /// On addon is enabled.
```
PintTheDragon, line 29: Called when the addon is enabled.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        void OnEnable();
+
+        /// <summary>
+        /// On receive message from server.
```
PintTheDragon, line 34: Called when a message is received from a server.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <param name="server">Received from.</param>
+        /// <param name="reader">Reader data.</param>
+        void OnMessageReceived(NPServer server, NetDataReader reader);
+
+        /// <summary>
+        /// On addon is ready.
```
PintTheDragon, line 41: Called when the addon is ready.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <param name="server">Received from.</param>
+        void OnReady(NPServer server);
+
+        /// <summary>
+        /// On command received from dedicated app.
```
PintTheDragon, line 47: Called when a console command is received.

Exiled.Network/API/Interfaces/IAddon.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <param name="cmd"> Command name.</param>
+        /// <param name="arguments"> Command arguments.</param>
+        void OnConsoleCommand(string cmd, List<string> arguments);
+
+        /// <summary>
+        /// On server command response.
```
PintTheDragon, line 54: Called when a server console response is received.

PintTheDragon (changes requested):
Alright I just realized how much is here and I'm gonna have to just take a look later :kek:

EXILED.sln
Killers0992: idk i will revert that

Exiled.DedicatedNetwork/Host.cs
```diff
@@ -0,0 +1,369 @@
...
+        private DedicatedConfig config;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Host"/> class.
+        /// </summary>
+        public Host()
+        {
+            Logger = new ConsoleLogger();
+
+            if (!File.Exists("./config.yml"))
```
joker-119, line 43: I'm fairly sure this will break something on Windows servers. Why not just use File.Exists("config.yml") ? File.Exists always defaults to checking the working directory of the exe/library, which is the same as ./ in Linux.

Exiled.DedicatedNetwork/GlobalSuppressions.cs
```diff
@@ -0,0 +1,12 @@
+// -----------------------------------------------------------------------
```
joker-119, line 1: Why are any of these suppressions necessary? There is a reason these rules exist, and they should be followed in all Exiled.x projects.

joker-119:
Postponing review until after dev branch is unlocked.

RedRangerr:
sex

PintTheDragon:
*this is a message from me, not the EXILED team*

This is a really nice feature, and would be wonderful as a dev tool/dependency (I believe it was at some point but got removed), but it's way too much to include in base EXILED.

EXILED has definitely become less than low-level, now with some larger additions like CustomItems and CustomRoles (which, I will admit, my decision is a bit swayed on), but this adds some pretty heavy fees features compared to them.

First off, it adds networking and tcp sockets, which are pretty heavy. Of course, they're needed to do what this library achieves, but they also do take a bit of a toll on performance. Bigger servers will handle them fine, but smaller ones will not. But, I do recognize that these won't just be turned on automatically, which leads me to my second point.

Many servers won't find this useful. That's just how it's going to be. First off, there's tons of single servers that won't benefit from this because they don't have any other servers to connect to. Then, there will be plenty of servers that just won't use networked plugins. Of course, the same can be said about other parts of EXILED, but many of those are much smaller.

Now, none of this is to say this is a bad addition. Like I said above, this would work wonderfully as a standalone plugin, and has a ton of uses. It just doesn't quite fit in EXILED itself.

PR #589, opened 2021-08-22
Add Room.Color

PintTheDragon:


PintTheDragon:
Something like this will be re-added when the basegame has a proper way to manage lights.

PR #590, opened 2021-08-23
Improving CustomItems API

SebasCapo:
On a previous PR (#464) CI.Spawn was given an out parameter instead of simply returning the Pickup object because we didn't want to break plugins till 3.0, so here ya go

also hello fellow dev boi, did I tell you how pretty you're looking today? please approve my pr soon!!11


SebasCapo:
oop wrong branch

PR #610, opened 2021-08-25
Basegame warning isn't enough, they need to be yelled at more.

joker-119:


sanyae2439 (approved):

BuildBoy12 (approved):

Exiled.Loader/Loader.cs
```diff
@@ -125,12 +128,22 @@ static Loader()
...
+            ? new WindowsPrincipal(WindowsIdentity.GetCurrent()).IsInRole(WindowsBuiltInRole.Administrator)
+            : geteuid() == 0;
+
         /// <summary>
         /// Runs the plugin manager, by loading all dependencies, plugins, configs and then enables all plugins.
         /// </summary>
         /// <param name="dependencies">The dependencies that could have been loaded by Exiled.Bootstrap.</param>
         public static void Run(Assembly[] dependencies = null)
         {
+            if (IsAdministrator)
```
PintTheDragon, line 141: Since this is only used once, the check should be directly embedded inside the if statement.

PintTheDragon (changes requested):

Exiled.Loader/Loader.cs
PintTheDragon: That, or as a local variable above it. It has no need to be a property.

Exiled.Loader/Loader.cs
```diff
@@ -128,17 +128,13 @@ static Loader()
...
-            : geteuid() == 0;
-
         /// <summary>
         /// Runs the plugin manager, by loading all dependencies, plugins, configs and then enables all plugins.
         /// </summary>
         /// <param name="dependencies">The dependencies that could have been loaded by Exiled.Bootstrap.</param>
         public static void Run(Assembly[] dependencies = null)
         {
-            if (IsAdministrator)
+            if ((RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && new WindowsPrincipal(WindowsIdentity.GetCurrent()).IsInRole(WindowsBuiltInRole.Administrator)) || geteuid() == 0)
```
PintTheDragon, line 137: What happens in the case that the platform is windows and they aren't admin? The if should not be ran, but you call an unnecessary `geteuid()`.

Killers0992:
Good pr

Exiled.Loader/Loader.cs
```diff
@@ -134,7 +134,8 @@ static Loader()
         /// <param name="dependencies">The dependencies that could have been loaded by Exiled.Bootstrap.</param>
         public static void Run(Assembly[] dependencies = null)
         {
-            if ((RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && new WindowsPrincipal(WindowsIdentity.GetCurrent()).IsInRole(WindowsBuiltInRole.Administrator)) || geteuid() == 0)
+            if ((RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && new WindowsPrincipal(WindowsIdentity.GetCurrent()).IsInRole(WindowsBuiltInRole.Administrator)) ||
+                ((RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) && geteuid() == 0))
```
PintTheDragon, line 138: The two non-windows OS checks are unnecessary. You can simply check if it's windows. If it's windows, check if it's windows root, otherwise check if it's linux root.

PintTheDragon (approved):
LGTM

BuildBoy12 (approved):
noice

PR #629, opened 2021-08-27
Update Scp096 Max Hume

BuildBoy12:
Mirror 11.0 maximum hume shield change.


joker-119:
tbh I swear I already fixed that.

PR #632, opened 2021-08-27
Added Player::ActiveEffects

NaoUnderscore:


NaoUnderscore:
Fixed using directive.

PR #642, opened 2021-08-28
Changed SyncingData pre-fix patch into a Transpiler

DGvagabond:


BuildBoy12:
Disable SA1118 and reenable 1313.

PR #655, opened 2021-08-29
Added IsNorthwoodStaff and IsGlobalModerator

terminator-97:


Michal78900:
Making these also a setters could potentianally violate VSR or something

zabszk:
> 
> 
> Making these also a setters could potentianally violate VSR or something

Adding the setters DOES NOT violate the VSR, but using these setters on any verified servers does, so it's quite pointless to add the setters if they can't be used.

Michal78900:
> > Making these also a setters could potentianally violate VSR or something
> 
> Adding the setters DOES NOT violate the VSR, but using these setters on any verified servers does, so it's quite pointless to add setters that can't be used.

Yeah, that was what I meant.

zabszk (approved):

PintTheDragon (approved):

PR #656, opened 2021-08-29
RoomColor stuff

babyboucher:
Added RoomColor and ResetRoomColor

### DIFF
--- a/Exiled.API/Features/Room.cs
+++ b/Exiled.API/Features/Room.cs
@@ -69,6 +69,19 @@ namespace Exiled.API.Features
         }
 
         /// <summary>
+        /// Gets or sets the color of the room's lights.
+        /// </summary>
+        public Color RoomColor
+        {
+            get => (Color)FlickerableLightController.WarheadLightColor;
+            set
+            {
+                FlickerableLightController.WarheadLightColor = value;
+                FlickerableLightController.WarheadLightOverride = true;
+            }
+        }
+
+        /// <summary>
         /// Gets a <see cref="IEnumerable{T}"/> of <see cref="Camera079"/> in the <see cref="Room"/>.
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Room.cs
```diff
@@ -68,6 +68,19 @@ public float LightIntensity
             set => FlickerableLightController.Network_lightIntensityMultiplier = value;
         }
 
+        /// <summary>
+        /// Gets or sets the color of the room's lights.
+        /// </summary>
```
PintTheDragon, line 73: The property name and description are a bit deceiving. This does not change the room's color, but rather the room's warhead light color, but that isn't mentioned anywhere. 

My main issue with this is that it interferes with warhead. I made a similar pr but closed it when I realized that it would cause unexpected behavior. The way it works needs to be fully documented.

PintTheDragon (changes requested):

Exiled.API/Features/Room.cs
babyboucher: The thing is that the game itself calls this room color. The roomcolor command does the same thing, well internally its called Warhead color. I think its safe to call it roomcolor because that is what the base game command calls it.

babyboucher:
There

PintTheDragon (approved):

NaoUnderscore:
I'd rename it "Color" and "ResetColor" since they're inside `Room`.
 `Room.RoomColor` and `Room.ResetRoomColor()` are bad imo.
 I'd like more  `Room.Color` and `Room.ResetColor()`

NaoUnderscore (commented):
L103 - Typo: _defualt_ -> _default_

babyboucher:
Done

PR #659, opened 2021-08-29
Change AnnouncingDecontamination patch to transpiler

NaoUnderscore:


NaoUnderscore:
I'll add docs in a few mins

PR #665, opened 2021-08-30
changed Open to IsOpen

Monaldcry7788:


joker-119:
I'm not sure if this change is really necessary?

NaoUnderscore:
I'd prefer using IsOpen, also it's better to change it now to avoid breaking everything in future, like in the Exiled Beta.

Exiled.API/Features/Door.cs
```diff
@@ -53,7 +53,7 @@ public Door(DoorVariant door)
         /// <summary>
         /// Gets or sets a value indicating whether the door is open.
         /// </summary>
-        public bool Open
```
PintTheDragon, line 56: Please add the `Open` property back, but mark it as Obsolete.

PintTheDragon (approved):

Exiled.API/Features/Door.cs
Monaldcry7788: Ok, give me a few minutes

joker-119: That isn't necessary during the alpha stage @PintTheDragon

PintTheDragon (approved):
My first approval was meant to be a change request. I'm actually approving it this time.

PR #666, opened 2021-08-30
Map::ActivatedGenerators code cleanup

green1052:
linq hooray!


joker-119:
When only searching for one condition, linq is slower than a standard foreach() or for() loop

PR #679, opened 2021-08-31
Let's have normal default permissions

PintTheDragon:
Because helper doesn't normally exist and definitely shouldn't have every permission.


PintTheDragon:
FUCK

PR #683, opened 2021-09-01
Custom roles

joker-119:
Early Draft, but it works. Roles themselves need to be tested but example of the API in use can be found here:

https://github.com/joker-119/CustomRoles/tree/exiled-api-update


Exiled.CustomRoles/API/Extensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <returns>A <see cref="ReadOnlyCollection{T}"/> of all current custom roles.</returns>
+        public static ReadOnlyCollection<CustomRole> GetCustomRoles(this Player player)
+        {
+            List<CustomRole> roles = new List<CustomRole>();
+            Component[] components = player.GameObject.GetComponents(typeof(CustomRole));
+            foreach (Component component in components)
+            {
+                if (component is CustomRole role)
+                    roles.Add(role);
+            }
```
PintTheDragon, line 36: Why is this check needed? Doesn't unity ensure they all are custom roles?

PintTheDragon (changes requested):
I think the main issue with this is that:
1. It over uses components.
2. It doesn't feel like a library, but something a bit in between. I don't really feel like a library for managing roles needs to have client console commands.

I would rework it away from using components (perhaps have a per-player RoleManager) component that manages roles and abilities.

PintTheDragon:
I also think that a proper spawning system should be implemented (so the player's role is only set once if they're going to spawn in as a custom role).

joker-119:
> I also think that a proper spawning system should be implemented (so the player's role is only set once if they're going to spawn in as a custom role).

Since it is plugins responsible for assigning the custom roles and not EXILED, it makes sense that they should handle the logic behind when to spawn them

Exiled.CustomRoles/API/Extensions.cs
joker-119: probably but GetComponents(typeof()) returns a Component[] not a CustomRole[] so I need to cast them before adding them to the list anyways.

Also these probably won't remain components since that makes setting up configs for them more annoying than if the CustomRole itself was able to be serialized as a config object like CustomItems are

Exiled.API/Extensions/SpawnExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <param name="location">The <see cref="SpawnLocation"/> to check.</param>
+        /// <returns>The <see cref="Vector3"/> used for that spawn location. Can be <see cref="Vector3.zero"/>.</returns>
+        public static Vector3 GetPosition(this SpawnLocation location)
+        {
+            Transform transform = location.GetDoor();
+
+            if (transform == null)
+                return default;
+
+            return transform.position + (Vector3.up * 1.5f) + (transform.forward * (ReversedLocations.Contains(location) ? -3f : 3f));
```
PintTheDragon, line 67: Magic line maybe needs some info.

Exiled.API/Extensions/SpawnExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+                    return "173_CONNECTOR";
+                case SpawnLocation.InsideServersBottom:
+                    return "SERVERS_BOTTOM";
+                case SpawnLocation.InsideEscapePrimary:
+                    return "ESCAPE_PRIMARY";
+                case SpawnLocation.InsideEscapeSecondary:
+                    return "ESCAPE_SECONDARY";
+                default:
+                    return default;
+            }
```
PintTheDragon, line 137: Any reason to not just use an array?

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -446,7 +451,7 @@ public virtual void SpawnAll()
             if (SpawnProperties == null)
                 return;
 
-            Spawn(SpawnProperties.StaticSpawnPoints, Math.Min(0, SpawnProperties.Limit - Spawn(SpawnProperties.DynamicSpawnPoints, SpawnProperties.Limit)));
+            Spawn(SpawnProperties.StaticSpawnPoints, Math.Min(0, SpawnProperties.Limit - Math.Min(0, Spawn(SpawnProperties.DynamicSpawnPoints, SpawnProperties.Limit) - Spawn(SpawnProperties.RoleSpawnPoints, SpawnProperties.Limit))));
```
PintTheDragon, line 454: Please add a comment explaining what this does.

Exiled.CustomRoles/API/Extensions.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            return roles.AsReadOnly();
+        }
+
+        /// <summary>
+        /// Registers an <see cref="IEnumerable{T}"/> of <see cref="CustomRole"/>s.
+        /// </summary>
+        /// <param name="customRoles"><see cref="CustomRole"/>s to be registered.</param>
+        public static void Register(this IEnumerable<CustomRole> customRoles)
```
PintTheDragon, line 46: Might be nice to have a params option for this. Something like
```
Register(customRole1, customRole2);
```

Exiled.CustomRoles/API/Extensions.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            foreach (CustomRole customItem in customRoles)
+                customItem.TryRegister();
+        }
+
+        /// <summary>
+        /// Unregisters an <see cref="IEnumerable{T}"/> of <see cref="CustomRole"/>s.
+        /// </summary>
+        /// <param name="customRoles"><see cref="CustomRole"/>s to be unregistered.</param>
+        public static void Unregister(this IEnumerable<CustomRole> customRoles)
```
PintTheDragon, line 59: See above.

Exiled.CustomRoles/API/Features/ActiveAbility.cs
```diff
@@ -0,0 +1,141 @@
...
+            ShowMessage(player);
+            AbilityUsed(player);
+            Timing.CallDelayed(Duration, () => EndAbility(player));
+        }
+
+        /// <summary>
+        /// Ends the ability.
+        /// </summary>
+        /// <param name="player">The <see cref="Player"/> the ability is ended for.</param>
+        public void EndAbility(Player player)
```
PintTheDragon, line 71: This may cause issues if the plugin ends the ability before it ends. Maybe add an `ActivePlayers.Contains` check somewhere around here.

Exiled.CustomRoles/API/Features/CustomRole.cs
```diff
@@ -0,0 +1,445 @@
...
+    {
+        /// <summary>
+        /// Gets a list of all registered custom roles.
+        /// </summary>
+        public static HashSet<CustomRole> Registered { get; } = new HashSet<CustomRole>();
+
+        /// <summary>
+        /// Gets or sets the custom RoleID of the role.
+        /// </summary>
+        public abstract uint Id { get; set; }
```
PintTheDragon, line 49: Please don't do this. This reminds me of hellish modded minecraft configs where every id was numeric, and you had to configure each mod to not have conflicting IDs by going through every single config file. Why not simply have string names, or identify by plugin name + numerical id?

Exiled.CustomRoles/API/Features/CustomRole.cs
```diff
@@ -0,0 +1,445 @@
...
+            if (Role != RoleType.None)
+            {
+                player.SetRole(Role, SpawnReason.ForceClass, true);
+            }
+
+            Timing.CallDelayed(1.5f, () =>
+            {
+                Vector3 pos = GetSpawnPosition();
+                if (pos != Vector3.zero)
+                    player.Position = pos + (Vector3.up * 1.5f);
```
PintTheDragon, line 258: Please explain.

Exiled.CustomRoles/API/Features/CustomRole.cs
```diff
@@ -0,0 +1,445 @@
...
+
+            if (SpawnProperties.RoleSpawnPoints.Count > 0)
+            {
+                foreach ((float chance, Vector3 pos) in SpawnProperties.RoleSpawnPoints)
+                {
+                    int r = Loader.Random.Next(100);
+                    if (r <= chance)
+                        return pos;
+                }
+            }
```
PintTheDragon, line 370: Joker please for the love of god just make this a method. Also can't it just not pick a role sometimes?

Exiled.CustomRoles/CustomRoles.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    using System.Collections.Generic;
+
+    using Exiled.API.Features;
+    using Exiled.CustomRoles.Events;
+
+    /// <summary>
+    /// Handles all custom role API functions.
+    /// </summary>
+    public class CustomRoles : Plugin<Config>
```
PintTheDragon, line 18: No `OnDisabled`?

PintTheDragon (changes requested):

Exiled.CustomRoles/API/Features/CustomRole.cs
joker-119: Making sure they don't spawn inside the floor.

Exiled.CustomRoles/API/Features/CustomRole.cs
joker-119: Exiled.CustomItems already does this and there's been no complaints?
Technically the ID numbers could just be removed completely, but they are there to allow, for example, a command that gives a role to a player, that can accept the id # instead of the role's name. This makes life simpler for the creator of said commnad, because they don't need to parse the name etc, they can just use an int.

Exiled.CustomRoles/API/Extensions.cs
joker-119: I'm not sure what you mean by this, please explain? Each CustomRole itself already has a TryRegister() method, this just makes life a tad easier if you have multiple roles, you can create the new classes and add them to a list, then use this extension to register them all at once.

Exiled.API/Extensions/SpawnExtensions.cs
joker-119: Not sure how an array would be useful for converting enums to base-game room names?

Exiled.CustomRoles/API/Features/CustomRole.cs
joker-119: I'm not actually sure that's possible, since each of those 3 values are different types, I'd have to re-build the list of them as generic spawn points for a single method to exist to iterate them and pick one at random.

Exiled.CustomRoles/API/Features/CustomRole.cs
PintTheDragon: My main concern here was that two plugins would register under the same id, and not have an option to change them (as is common with commands). To prevent that, there should be a rule requiring the item/role to be put inside the config, as is used in CustomRoles.

Exiled.API/Extensions/SpawnExtensions.cs
PintTheDragon: I'm marking this resolved as it isn't much of an issue.

Exiled.CustomRoles/API/Features/CustomRole.cs
PintTheDragon: Looks like C# is weird and this cannot be a method, so I guess it's fine.

PintTheDragon (approved):

BuildBoy12 (approved):

PR #685, opened 2021-09-01
Allow exceeding ammo limits

Michal78900:
This thing removes the code that forced player to drop their ammo if they had it too much of it (they were dropping it when they holstered their weapon).

This is my first *working* transpiler. Thanks Nao for help.


joker-119:
This transpiler alters base-game behavior in a way that isn't necessary for EXILED to function, therefore this is the domain of a plugin, not EXILED itself.

PR #686, opened 2021-09-02
Enum is gone I'm cursed and ProcessingHotkeyEventArgs

NaoUnderscore:


DGvagabond:
Has this been tested?

NaoUnderscore:
Yes, you're scamming us as always.

Exiled.Events/EventArgs/ProcessingHotkeyEventArgs.CS
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Gets the player who's pressing the hotkey.
+        /// </summary>
+        public Player Player { get; }
+
+        /// <summary>
+        /// Gets the pressed hotkey.
+        /// </summary>
+        public HotkeyButton Hotkey { get; }
```
joker-119, line 41: Is there a reason this isn't settable?

NaoUnderscore: I tried to make it settable, but I noticed that you deny the event until you change your current item manually.
Short answer: you shouldn't be able to set it.

NaoUnderscore: I mean, you can set the target item, but that causes the same problem.

PR #692, opened 2021-09-02
Room.LockDown and Room.UnlockAll

babyboucher:


joker-119 (changes requested):
Since these methods affect ALL doors, these should be static.

babyboucher:
Done

Exiled.API/Features/Room.cs
```diff
@@ -99,6 +99,35 @@ public Color Color
         /// <param name="duration">Duration in seconds.</param>
         public void TurnOffLights(float duration) => FlickerableLightController?.ServerFlickerLights(duration);
 
+        /// <summary>
+        /// Locks all the doors in the room.
+        /// </summary>
+        /// <param name="duration">Duration in seconds.</param>
+        /// <param name="locktype">DoorLockType of the lockdown.</param>
+        public static void LockDown(float duration, DoorLockType locktype = DoorLockType.Regular079)
```
joker-119, line 107: static methods should be above non-static ones.

Also this should be renamed to LockAll() and the documentation changed to indicate that it affects all rooms.

joker-119 (changes requested):

joker-119:
I'm actually a monke and thought this was in the Door class, not the Room class. I apologize, you were correct with the non-static methods.

babyboucher:
No problem

joker-119 (approved):

PintTheDragon:
Don't doors have a proper locking system? I feel like they should probably be used here.

PintTheDragon:
> 
> 
> > Don't doors have a proper locking system? I feel like they should probably be used here.
> 
> Yes, it's used by changing the DoorLockType property ??

I just think that there should be a way to undo a lockdown for your specific lock type, for when you want to undo the lock, and use the other method for when you want to ensure the door is open. Also, for plugins, I'd say that defaulting to the 079 locktype might not be the best.

joker-119:
> > > Don't doors have a proper locking system? I feel like they should probably be used here.
> > 
> > 
> > Yes, it's used by changing the DoorLockType property ??
> 
> I just think that there should be a way to undo a lockdown for your specific lock type, for when you want to undo the lock, and use the other method for when you want to ensure the door is open. Also, for plugins, I'd say that defaulting to the 079 locktype might not be the best.

I still need to re-do the door lock system to properly use flags in exiled like they are in base-game, once done the code in the PR will need to be changed accordingly, but I will worry about that when the time comes, for now this is acceptable given the API's restrictions on the use of flags.

PR #694, opened 2021-09-04
Added WalkingOnSinkhole event

DGvagabond:


joker-119:
This should be a transpiler

PR #700, opened 2021-09-05
ShootingTarget abstraction/events

steven4547466:
Testing version got in instead of real version. Sad.

I was trying to make it work while in local mode with that offset and I forgot to switch it back 😔


steven4547466:
Do not merge this yet.

steven4547466:
Added `Exiled.API.Enums.ShootingTargetButton`, the event will now call with all of those values instead of the base game ones. Furthermore, the event now actually calls when removing or toggling sync mode.

steven4547466:
This will also work with local mode, but setting IsAllowed to false will not prevent the event from propagating if the button is increasing/decreasing hp or reset time. If this is an issue we can force it to only work on sync mode or call a different event without IsAllowed when it's in local mode but that'll get messy too.

PR #704, opened 2021-09-05
ShootingTarget abstraction and events

steven4547466:


steven4547466:
Original PR deleted itself when I changed the branch name

steven4547466:
If anyone would like to request events/abstracted methods into the wrapper class, now would be the time.

steven4547466:
I'm looking into getting the actual Exiled Item from IDamageDealer in the damage event

Exiled.Events/EventArgs/InteractingShootingTargetEventArgs.cs
```diff
@@ -48,11 +55,48 @@ public InteractingShootingTargetEventArgs(Player player, ShootingTarget shooting
...
         /// Gets or sets a value indicating whether or not the interaction is allowed.
         /// </summary>
-        public bool IsAllowed { get; set; }
+        public bool IsAllowed
+        {
+            get => isAllowed;
+            set
+            {
+                if (!ShootingTarget.IsSynced)
+                    throw new InvalidOperationException("Attempted to set IsAllowed while target is in local mode. Set target's IsSynced to true before setting IsAllowed.");
```
Michal78900, line 97: This can be removed. I'm pretty sure you can `ev.IsAllowed = false` without any issues, even if the target is in local mode.

steven4547466: I forgot to remove it when I added the AutoResetTime and MaxHp properties. However those are still only allowed to be set while in sync mode or you will mess up the target.

Michal78900: Exactly

steven4547466:
Wait before merging this. The logic behind GetNextValue seems to still be flawed.

steven4547466:
Seems to have been working as intended.

PR #722, opened 2021-09-12
Fix GetDoorType()

steven4547466:
Currently GetDoorType() returns UnknownDoor for all of the entrance, heavy, and light containment doors when it should return the correct enum for it. That is now fixed.


steven4547466:
It appears `PrisonDoor`'s also fell under this category.

Michal78900 (approved):

BuildBoy12 (approved):

PintTheDragon (approved):

PR #724, opened 2021-09-14
Fix setting CurrentItem to null or a None item

steven4547466:
Currently you will get this error when setting Player.CurrentItem to null or `new Item(ItemType.None)`:
```
[2021-09-14 16:54:00.049 -04:00] [STDOUT] NullReferenceException: Object reference not set to an instance of an object
[2021-09-14 16:54:00.049 -04:00] [STDOUT]   at Exiled.API.Features.Player.set_CurrentItem (Exiled.API.Features.Items.Item value) [0x0001a] in <9d56c16099cf4088878254ced68c312e>:0
```
This PR fixes that.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -687,12 +687,19 @@ namespace Exiled.API.Features
 
             set
             {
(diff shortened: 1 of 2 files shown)


steven4547466:
Whoops forgot to hard reset

PR #735, opened 2021-09-22
Add ev.Item to ProcessingHotkeyEventArgs

steven4547466:


Exiled.Events/EventArgs/ProcessingHotkeyEventArgs.cs
```diff
@@ -40,6 +43,11 @@ public ProcessingHotkeyEventArgs(Player player, HotkeyButton hotkey, bool isAllo
         /// </summary>
         public HotkeyButton Hotkey { get; }
 
+        /// <summary>
+        /// Gets or sets the item change to.
+        /// </summary>
+        public Item Item { get; set;  }
```
joker-119, line 49: The setter for this should check if the player actually possesses the item, and either throw an invalid operation error, or give them the item if they do not have it.

joker-119 (changes requested):

TheUltiOne:
Did Nao ever make this? Can this be closed?

NaoUnderscore:
I didn't make it yet, because I was dealing with attachments.

TheUltiOne:
yeah but shouldnt this be closed as you're "cloning" this? since it will be a different PR.

PR #739, opened 2021-09-26
Config | Config Types

Jesus-QC:
This allows people to have their configs into separated folders or use the default one.

- This creates a file inside the Config folder to configure the Loader so it doesnt have to read a config to check if it is using separated or default type.


Jesus-QC:
what the fuck

PR #740, opened 2021-09-26
Config | Config Types

Jesus-QC:
This allows people to have their configs into separated folders or use the default one.

 - This creates a file inside the Config folder to configure the Loader so it doesnt have to read a config to check if it is using separated or default type.
 
https://cdn.discordapp.com/attachments/867421794545565699/891640423783743528/configTypesTest.mp4

### DIFF
--- a/Exiled.Events/Commands/Config/EConfig.cs
+++ b/Exiled.Events/Commands/Config/EConfig.cs
@@ -14,7 +14,6 @@ namespace Exiled.Events.Commands.Config
     /// <summary>
     /// The config command.
     /// </summary>
-    [CommandHandler(typeof(RemoteAdminCommandHandler))]
     [CommandHandler(typeof(GameConsoleCommandHandler))]
     public class EConfig : ParentCommand
     {
--- a/Exiled.Events/Commands/Config/EConfig.cs
+++ b/Exiled.Events/Commands/Config/EConfig.cs
@@ -1,0 +1,49 @@
+// -----------------------------------------------------------------------
+// <copyright file="EConfig.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.Commands.Config
+{
+    using System;
+
+    using CommandSystem;
+
+    /// <summary>
+    /// The config command.
+    /// </summary>
+    [CommandHandler(typeof(RemoteAdminCommandHandler))]
+    [CommandHandler(typeof(GameConsoleCommandHandler))]
+    public class EConfig : ParentCommand
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EConfig"/> class.
+        /// </summary>
+        public EConfig() => LoadGeneratedCommands();
+
+        /// <inheritdoc/>
+        public override string Command { get; } = "econfig";
+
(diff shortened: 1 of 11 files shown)


PintTheDragon:
My concern is that old configs would not be easily transferable to this new format. Also your proposing splitting up configs into different files but not translations, which is a bit odd.

Jesus-QC:
@PintTheDragon youre right but people will get used to it, and about translations I completely forgot that, will fix that.

Jesus-QC:
Now translations have their own folder too!

Jesus-QC:
- Forgot to create the translations folder, will fix later.

PintTheDragon:
> @PintTheDragon youre right but people will get used to it

You can't just make a massive change like this and leave everyone using it to deal with it. Also, this completely changes how the config system works and needs to be discussed to see if we even want to implement it, and how to go about making a clean solution that reduces the effort of transitioning for the end-user.

@joker-119 @iopietro @BuildBoy12 @sanyae2439 what do you think about this new config system?

Jesus-QC:
> > @PintTheDragon youre right but people will get used to it
> 
> You can't just make a massive change like this and leave everyone using it to deal with it. Also, this completely changes how the config system works and needs to be discussed to see if we even want to implement it, and how to go about making a clean solution that reduces the effort of transitioning for the end-user.
> 
> joker-119 iopietro BuildBoy12 sanyae2439 what do you think about this new config system?

Yeah, i agree that it should be discused, anyways it doesnt change the entire config system, the actual one can be used and however any new option enhances the framework, people are of course right now in their comfort zone because they already have configured its configs, but some other people would like this option.

BuildBoy12 (changes requested):
Not sure how I feel about the separated config option yet. I'll have to try it a little bit. In the mean time, fix your build errors.

joker-119:
But like, what's the actual point of this, aside from doing the exact same thing but with more steps?

Jesus-QC:
Well, people were complaining about having everything in the same file, this would fix that giving them the option to separate its configs, and it does too give devs the option to keep plugin files organisez putting them into the plugins folder.

Working on the commands.

Jesus-QC:
Ok, commands tested and working:

econfig split
econfig merge

I didnt use config because its a base game command, econfig means exiled-config. Aliases: ecfg

Exiled.Events/Commands/Config/EConfig.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Exiled.Events.Commands.Config
+{
+    using System;
+
+    using CommandSystem;
+
+    /// <summary>
+    /// The config command.
+    /// </summary>
+    [CommandHandler(typeof(RemoteAdminCommandHandler))]
```
PintTheDragon, line 17: There's no reason for this to be done in game. Please remove it from the ra list and add a check to ensure that only the server console is using it.

PintTheDragon (changes requested):

Exiled.Events/Commands/Config/EConfig.cs
Jesus-QC: Ok, that should resolve it

joker-119 (approved):

PintTheDragon:
Please add a check to ensure only the console is running it. People can run console commands from RA still, and this is one that does not ever need to be ran ingame.

Jesus-QC:
Moving it to a new one to avoid this compatible issues and recode it completely

PR #742, opened 2021-10-01
Dummy API

gamehunt:
Converted Michal's prefixes to transpilers as he asked


gamehunt:
Whoops, found tinybrain mistake, gona fix it rn

joker-119:
This only fixes two of the potential issues with dummies. There are many, many more that would need to also get fixed to make them viable to be implemented within EXILED itself. We would also need to add an API to create dummies. Unless EXILED itself is going to create them via an API, I do not see why we should add patches to EXILED itself related to dummies.


Also re-merge dev branch into this PR plz.

gamehunt:
Kk, should I convert this to draft for dummies api then? Also you can use those patches in the AT as a temporary fix anyway

joker-119:
Probably yes

gamehunt:
Closing this in favor of #843

PR #747, opened 2021-10-08
Re-added Player::IsJumping

NaoUnderscore:


NaoUnderscore:
Damn you're right, lemme change the branch.

Michal78900:
You can also add properties for the rest of player's states in this PR (if they work)
```cs
// Token: 0x04000C14 RID: 3092
public bool IsWalking;

// Token: 0x04000C15 RID: 3093
public bool IsSprinting;

// Token: 0x04000C16 RID: 3094
public bool IsSneaking;
```

joker-119:
> You can also add properties for the rest of player's states in this PR (if they work)
> 
> ```cs
> // Token: 0x04000C14 RID: 3092
> public bool IsWalking;
> 
> // Token: 0x04000C15 RID: 3093
> public bool IsSprinting;
> 
> // Token: 0x04000C16 RID: 3094
> public bool IsSneaking;
> ```
> 
> They are in the same class as `isJumping`

If that's the case, why not make a MovementState enum for the player class, instead of a bunch of bools?

PR #749, opened 2021-10-09
Hotfix for DocFX StackOverflowException

ox-ox-ox:
Workaround for [dotnet/docfx#7035](https://github.com/dotnet/docfx/issues/7035) to get functional + updated docs.

I am aware this is a PR to master, if it needs to be moved to dev just let me know via Discord and I can change it.


PintTheDragon:
Do we have a reason to use circular references? Why not just did the circular references instead.

Issue #743, opened 2021-10-03
Typo in CustomItem::Give(Player player, bool displayMessage)


NooTChhYT:
The **Give(Player player, bool displayMessage)** method uses the **type** field with lowercase instead of the uppercase one. In every other method, the **Type** field is used. Overriding instances of **Type** result in giving the player a Janitor Keycard.

![image](https://user-images.githubusercontent.com/68049877/135762524-05325f88-51ed-4edc-930f-7a05efbf55fd.png)

NaoUnderscore:
@NooTChhYT This has been fixed in PR #761

PR #761, opened 2021-10-11
Fix Issue #743

NaoUnderscore:


PintTheDragon:
I'm a bit confused on what this pr actually does.

PR #780, opened 2021-10-14
Added `Player::ApplyRandomEffect`

Marco15453:
Moved to this from #779


PintTheDragon:
I'll have to agree with @SebasCapo. It's a cool feature but seems like it would only be useful for a few plugins.

joker-119 (approved):

BuildBoy12 (approved):

PR #828, opened 2021-10-31
Added SetNoclip

terminator-97:
Added SetNoclip

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1955,5 +1955,11 @@ namespace Exiled.API.Features
 
         /// <inheritdoc/>
         public override string ToString() => $"{Id} {Nickname} {UserId} {Role} {Team}";
+
+        /// <summary>
+        /// Sets noclip status to a specific player.
+        /// </summary>
+        /// <param name="status">The value to be set.</param>
+        public void SetNoclip(bool status) => referenceHub.characterClassManager.SetNoclip(status);      
     }
 }


Exiled.API/Features/Player.cs
```diff
@@ -1955,5 +1955,11 @@ public void ChangeEffectIntensity(EffectType type, byte intensity, float duratio
 
         /// <inheritdoc/>
         public override string ToString() => $"{Id} {Nickname} {UserId} {Role} {Team}";
+
+        /// <summary>
+        /// Sets noclip status to a specific player.
```
Marco15453, line 1960: Sets noclip status to a specific player. -> Sets the player's noclip status.

Exiled.API/Features/Player.cs
```diff
@@ -1955,5 +1955,11 @@ public void ChangeEffectIntensity(EffectType type, byte intensity, float duratio
...
+        /// </summary>
+        /// <param name="status">The value to be set.</param>
```
Marco15453, line 1962: The value to be set -> Whether or not noclip should be activated.

Exiled.API/Features/Player.cs
Marco15453, line 1960: Sets noclip status to a specific player. -> Sets the player's noclip status.

Exiled.API/Features/Player.cs
Marco15453, line 1962: The value to be set -> Whether or not noclip should be activated.

Michal78900:
What's the difference between this and already implemented `NoClipEnabled`?
https://github.com/Exiled-Team/EXILED/blob/master/Exiled.API/Features/Player.cs#L443-L451

Exiled.API/Features/Player.cs
PintTheDragon: Could be just Whether noclip should be activated.

Marco15453:
> What's the difference between this and already implemented `NoClipEnabled`? https://github.com/Exiled-Team/EXILED/blob/master/Exiled.API/Features/Player.cs#L443-L451

This would be absolutely useless because it's already implemented.

Exiled.API/Features/Player.cs
Marco15453: ^^

joker-119:
This is already implemented.

PR #829, opened 2021-10-31
Fix Server.Restart not reconnecting players.

joker-119:
Idk why fastrestart was passed as true here, it makes clients not expect to get disconnected, then they immediately do so they do not reconnect.

### DIFF
--- a/Exiled.API/Features/Server.cs
+++ b/Exiled.API/Features/Server.cs
@@ -137,8 +137,7 @@ namespace Exiled.API.Features
         /// </summary>
         public static void Restart()
         {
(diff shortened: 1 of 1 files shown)


PintTheDragon:
Wait, how does this restart the actual server? From what I can tell it only restarts the round.

PintTheDragon (approved):

joker-119:
> Wait, how does this restart the actual server? From what I can tell it only restarts the round.

By setting the NextRound action from DoNothing to Restart

BuildBoy12 (approved):

PR #831, opened 2021-11-01
Fix double Player.Died calls for 079

joker-119:


joker-119:
Closing in favor of using #774

PR #836, opened 2021-11-03
Add Player::SetRoom

RedRangerr:


README.md
```diff
@@ -7,7 +7,7 @@
   <img src="https://img.shields.io/discord/656673194693885975?logo=discord" alt="Chat on Discord">
 </a>
 
-
+
```
PintTheDragon, line 10: Why?

PintTheDragon (changes requested):

README.md
RedRangerr: i had to trigger the build

RedRangerr:
its fine now?

RedRangerr:
there pint

joker-119 (changes requested):
Why is this it's own method, instead of adding a setter to Player::CurrentRoom ?
You will also need to compensate for the fact that some rooms centers don't put you inside the room itself, or put you in a place you will die or be out of bounds.  For this you should use the base-game ForceSafePos method, or some up with your own way to compensate for specific rooms that fall under the above issues, and offset the position the player is sent to based on the rooms rotation etc.

As it is, I can think of 5 rooms off the top of my head, that this will break for.

NaoUnderscore:
@RedRangerr Close this PR, it's pointless, unless you know what and how you're doing. It's going to break shit with a certain rooms.

PR #837, opened 2021-11-03
Added PickingUpScp330 Event (DO NOT MERGE)

Monaldcry7788:
Added PickingUpScp330EventArgs (added in handlers and in patch)


TheUltiOne:
When is this being added back?

PR #857, opened 2021-11-08
Make DisplayNickname return Nickname if it is null.

TheUltiOne:
I did not test it. It's such a simple addition.
If I _somehow_ made a stupid mistake, please troll :trollface: me.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -242,7 +242,7 @@ namespace Exiled.API.Features
         /// </summary>
         public string DisplayNickname
         {
-            get => ReferenceHub.nicknameSync.Network_displayName;
+            get => ReferenceHub.nicknameSync.Network_displayName == null ? Nickname : ReferenceHub.nicknameSync.Network_displayName;
             set => ReferenceHub.nicknameSync.Network_displayName = value;
         }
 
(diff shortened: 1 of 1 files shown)


Michal78900:
Please sync your PR with the current `dev` branch.

TheUltiOne:
mfw dumb

Michal78900 (approved):

Jesus-QC (approved):

NaoUnderscore:
This PR is a breaking change and it doesn't make sense. Plugins that are using `string::IsNullOrEmpty` on `Player::DisplayNickname` will be broken. If it can be null, there's a reason. You can't spoof or change return types to avoid null. If I was a dev, I'd fight to reject this PR.

PintTheDragon:
What's the point of this?

NaoUnderscore:
There's no point, and even if there were, it'd not justify this breaking change.

steven4547466:
Make this into a new `DisplayName` that returns `DisplayNickname ?? Nickname`

Marco15453:
I have to agree with @NaoUnderscore

joker-119:
This would be considered a breaking changes, as stated, this would break every plugin that checks if the DisplayNickname is null/empty before doing an action on it. This can be important for a plethora of reasons, but EXILED itself does this.
Objects that are able to be null are able to be null for a reason, and that behavior shouldn't be changed without a valid reason.

PR #858, opened 2021-11-09
Potential softlock fix

PintTheDragon:
The softlocks occur during death and escape, and are exasperated by EXILED. This method is ran on death and escape and has a while loop in it that calls a method that potentially wont remove the value from the while loop, creating a while-true.

This pretty much just forces the item to be removed from the inventory if it wasn't, so the while loop can't go on forever.

### DIFF
--- a/Exiled.Events/Patches/Events/Player/ChangingRole.cs
+++ b/Exiled.Events/Patches/Events/Player/ChangingRole.cs
@@ -156,13 +156,37 @@ namespace Exiled.Events.Patches.Events.Player
                     if (inventory.TryGetBodyArmor(out BodyArmor bodyArmor))
                         bodyArmor.DontRemoveExcessOnDrop = true;
                     while (inventory.UserInventory.Items.Count > 0)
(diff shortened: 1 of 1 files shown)


PintTheDragon:
I've tested this and confirmed that it works, although it doesn't have any noticeable effect on the crashes.

Michal78900:
> I've tested this and confirmed that it works, although it doesn't have any noticeable effect on the crashes.

If it doesn't have noticeable effect, why merging something that doesn't change anything?

PintTheDragon:
> > I've tested this and confirmed that it works, although it doesn't have any noticeable effect on the crashes.
> 
> If it doesn't have noticeable effect, why merging something that doesn't change anything?

I'll just leaving it up because the method it calls does have a potential to cause a while true if weird things are happening with items.

Jesus-QC:
I would close it since we have found the issue (#877)

Jesus-QC:
However this could be useful as a temporary fix

PintTheDragon:
> I would close it since we have found the issue (#877)

That pr only patches the EXILED API methods, which this doesn't interact with.

joker-119 (approved):

Issue #826, opened 2021-10-30
[BUG] Player.Zone in Surface gives you Zonetype.Unspecified


zhaguanyang:
**EXILED Version ("latest" is not a version):**
3.3.1

zhaguanyang:
Previously mentioned in https://discord.com/channels/656673194693885975/664263236048388097/902832491625316413

zhaguanyang:
Tested and it happens every time

joker-119:
Can you please test to verify whether this bug has been fixed in 3.4.0-beta?

zhaguanyang:
![image](https://user-images.githubusercontent.com/45165615/140304447-57134de5-df73-4710-8d27-9b01dad97a86.png)
Nope

zhaguanyang:
![image](https://user-images.githubusercontent.com/45165615/140304675-8967a8bf-f14c-4826-bfeb-67eabbb17706.png)

zhaguanyang:
Not sure if this is caused by the Halloween update, I am sure this thing works before Halloween

joker-119:
It's possible nw broke the surface areas zone type in 11.0, I will do since testing in a few hours to find out the issue

zhaguanyang:
Yeah, not cause by halloween. Still having this issue today

NaoUnderscore:
This bug has been fixed. See #864.

PR #859, opened 2021-11-09
Fix FindParentRoom not returning Outside

zhaguanyang:
Since #827 made the default room(first room) in rooms to be Outside. FindParentRoom needs to be updated in order to return Outside.

PS: #826 is the issue I am encountering which uses FindParentRoom


zhaguanyang:
Haven't tested it yet

Issue #868, opened 2021-11-10
[BUG] Player.Handcuff is broken


zhaguanyang:
**Describe the bug**
After Player.Handcuff, the player who got cuffed will not be able to open doors and use inventory, which is fine. However, that player won't get the hint saying he got detained. Moreover, the detained animation is not triggered, so from the other players' perspective, he is not detained. 
If you tried to press E with a gun pointing at the player after Player.Handcuff, he will get detained again(animation and hint are shown), instead of released.

**To Reproduce**
Use Player.Handcuff() 

**Expected behavior**
Other players should view him as a person who got disarmed and that player should see he got detained or not.

**EXILED Version ("latest" is not a version):**
3.6.2

zhaguanyang:
https://youtu.be/QCvZ_ZP3FE0

PR #869, opened 2021-11-10
Fixed Player.Handcuff being broken (Issue #868)

zhaguanyang:
Tested

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -341,7 +341,10 @@ namespace Exiled.API.Features
                 }
 
                 if (value != null)
+                {
                     Inventory.SetDisarmedStatus(value.Inventory);
+                    new DisarmedPlayersListMessage(DisarmedPlayers.Entries).SendToAuthenticated();
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


zhaguanyang:
#868

Exiled.API/Features/Player.cs
```diff
@@ -1276,6 +1276,8 @@ public void Handcuff(Player cuffer)
             if (!IsCuffed && Vector3.Distance(Position, cuffer.Position) <= 130f)
             {
                 Cuffer = cuffer;
+                Inventory.SetDisarmedStatus(cuffer.Inventory);
```
Michal78900, line 1279: The `Cuffer` property setter already calls `Inventory.SetDisarmedStatus()`, so this is redundant.

Exiled.API/Features/Player.cs
```diff
@@ -1276,6 +1276,8 @@ public void Handcuff(Player cuffer)
...
+                new DisarmedPlayersListMessage(DisarmedPlayers.Entries).SendToAuthenticated();
```
Michal78900, line 1280: This should also be moved into `Cuffer`'s setter, so the entire thing will look like this.
```cs
if (value != null)
{
    Inventory.SetDisarmedStatus(value.Inventory);
    new DisarmedPlayersListMessage(DisarmedPlayers.Entries).SendToAuthenticated();
}

```

Michal78900 (changes requested):

Michal78900 (approved):

PintTheDragon (approved):

PR #872, opened 2021-11-11
Improving Exiled.API.Features.Player.Get(string args)

Gnihtonen:
in Get(string args) useless additions to strings and GetDistance in this situation is too slow because we already know that secondString contains firstString => GetDistance = secondString.Length - firstString.Length (961 operations vs 1 operation).
Also we don't have to check firstString.Length < maxNameLength due to the same reason.


joker-119 (changes requested):
Please re-base your patch on the `dev` branch, or switch the target branch and re-pull so we don't generate a ton of commit noise

Michal78900:
Please fix build errors, remove unnecessary variables like `maxNameLength` and `lastnameDifference` (you don't use them at all)

Michal78900:
Revert `Revert "Merge branch 'dev' into patch-1"` as it breaks stuff from other PRs right now

Gnihtonen:
Yeah, sorry I'm using Git for the first time

Michal78900:
`.github/workflows/docs.yml` is also needs to be reverted back to `dev` state

Michal78900 (approved):
Seems to be working
```cs
Log.Info(Player.Get(ev.Player.Nickname) != null);
Log.Info(Player.Get("Michal78900") != null);
Log.Info(Player.Get("Michal") != null);
Log.Info(Player.Get("78900") != null);
Log.Info(Player.Get("Amogus") != null);
Log.Info(Player.Get("asdasd") != null);
Log.Info(Player.Get(ev.Player.Nickname.Reverse().ToString()) != null);
```
image

PintTheDragon (approved):

Jesus-QC (changes requested):
It looks great to me! Fix the issue.

Jesus-QC (approved):

joker-119 (approved):

PR #877, opened 2021-11-14
Server Freeze Problem

ismail0234:
The server freezes when the main game and the exiled item list do not match. Rarely will 2 items with the same serial number appear on the list. The reason is not yet clear

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1343,7 +1343,16 @@ namespace Exiled.API.Features
         public bool RemoveItem(Item item, bool destroy = true)
         {
             if (!ItemsValue.Contains(item))
+            {
                 return false;
+            }
+            
+            if (!Inventory.UserInventory.Items.ContainsKey(item.Serial))
+            {
+                ItemsValue.Remove(item);
+                return false;       
+            }
+
             if (destroy)
             {
                 Inventory.ServerRemoveItem(item.Serial, null);
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1346,11 +1346,11 @@ namespace Exiled.API.Features
             {
                 return false;
             }
-            
+
             if (!Inventory.UserInventory.Items.ContainsKey(item.Serial))
             {
                 ItemsValue.Remove(item);
-                return false;       
+                return false;
             }
 
             if (destroy)


joker-119:
This change is not necessary, as this method calls ServerRemoveItem, which is already patched to remove the item from their player dictionary: https://github.com/Exiled-Team/EXILED/blob/master/Exiled.Events/Patches/Generic/InventoryControlPatch.cs

ismail0234:
> This change is not necessary, as this method calls ServerRemoveItem, which is already patched to remove the item from their player dictionary: https://github.com/Exiled-Team/EXILED/blob/master/Exiled.Events/Patches/Generic/InventoryControlPatch.cs

Look carefully, you can't remove something that is not in the dictionary anyway. Therefore, it is deleted from the exiled list without calling the method.

joker-119:
Apologies, I didn't see the ! on the dictionary check.

PintTheDragon:
Please fix the build errors.

joker-119:
Honestly I feel finding the cause of the desync would be tremendously more productive than add in janky little 'fixes' for it, but i've been unable to reproduce this myself, can you give us a detailed explanation of how to cause this bug to occur in the first place?

ismail0234:
> Honestly I feel finding the cause of the desync would be tremendously more productive than add in janky little 'fixes' for it, but i've been unable to reproduce this myself, can you give us a detailed explanation of how to cause this bug to occur in the first place?

The same serial number freezes the server when the player dies.

Use the code below to recreate:
https://cdn.discordapp.com/attachments/472406493229219860/909384033203527680/unknown.png

Video Example:
https://cdn.discordapp.com/attachments/472406493229219860/909383417555218482/SCP_Secret_Laboratory_2021.11.14_-_13.00.27.06.DVR_Trim_2.mp4

joker-119:
I understand the cause, but this doesn't fix the actual issue of players having items with duplicate serials, we need to find **how** that is happening, and prevent it when it happens, rather than hotfixing a symptom of it occuring. But thusfar none of the dev team can reproduce this in normal gameplay.

Jesus-QC (approved):
Looks great

Jesus-QC:
I would merge it until we find what is causing the duplicated serials.

Jesus-QC:
Or at least as a type of null check

Michal78900 (approved):
I only approving that as a temporarly fix. The issue with duplicated serials needs to be resolved ASAP.

Marco15453:
> I only approving this as a temporarly fix. The issue with duplicated serials needs to be resolved ASAP.

I wonder if it builds xD

PR #878, opened 2021-11-14
Player Ammo methods

SebasCapo:
Initially I wanted to make a separate Ammo class that could be used as a Dictionary, which would also contain some cool methods and all that... Obviously that would've been the fancy way of doing this, so me being very lazy I instead opted for a few cool methods.

Either way, added the following methods to the `Player` class:
- `Player.AddAmmo(AmmoType, ushort)` to easily add ammo to the player's inventory.
- `Player.SetAmmo(AmmoType, ushort)` to set a specific ammo type to the specified amount.
- `Player.DropAmmo(AmmoType, ushort, bool)` to drop a specific amount of ammo. (I honestly don't have a clue as to what `checkMinimals` does)
- `Player.GetAmmo(AmmoType, ushort)` to get how much ammo the player has of a specific ammo type, not entirely necessary given `Player.Ammo` exists, but for the sake of consistency I added it.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1498,6 +1498,29 @@ namespace Exiled.API.Features
         public void ClearBroadcasts() => Server.Broadcast.TargetClearElements(Connection);
 
         /// <summary>
+        /// Adds the amount of a specified <see cref="AmmoType">ammo type</see> to the player's inventory.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


joker-119 (approved):

Michal78900 (approved):
I guess these methods are okay, but I would like to still have something like `player.Ammo[AmmoType.something] = number`, but I understand it may be annoying to make.

PintTheDragon (approved):

PR #888, opened 2021-11-26
Add Map::UnlockAllDoors(ZoneType zoneType)

RedRangerr:
ignore the docs changes i reverted them but this is an overload to unlock doors in a specific zone

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - master
-      - dev
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


Jesus-QC:
I would add a method for unlocking all doors without any arg.
Remove the one with the IEnumerable
And add another param with a new lock state (a bool) so the method can be used to lock too.
And change the summary and change the summary of the methods from unlocks all doors from the facility to unlocks all doors from one zone if the method uses any zonetype as a parameter.

RedRangerr:
https://github.com/RedRangerr/EXILED/blob/dev/Exiled.API/Features/Map.cs#L394
There already is a method that does it without any args

RedRangerr:
And there is Map.LockAllDoors to add a lock

PR #889, opened 2021-11-26
remove foreach in a foreach

Jesus-QC:


Exiled.Loader/Loader.cs
```diff
@@ -228,8 +228,13 @@ public static IPlugin<IConfig> CreatePlugin(Assembly assembly)
         {
             try
             {
-                foreach (Type type in assembly.GetTypes().Where(type => !type.IsAbstract && !type.IsInterface))
+                foreach (Type type in assembly.GetTypes())
                 {
+                    if (type.IsAbstract || type.IsInterface)
+                    {
+                        continue;
```
joker-119, line 235: We should have a debug line here like the one below telling us why we're continuing. Ik it wasn't there before anyways but since we're splitting the check we might as well.

PR #891, opened 2021-11-28
CurrentRoom.Players Null exception fixed

ismail0234:
If a player joins the server later, the currentroom value will return null. This value remains blank until the player respawns. It throws an exception when we use "player.CurrentRoom.Players".


PintTheDragon:
Why not just `player.CurrentRoom?.Transform`?

ismail0234:
> Why not just `player.CurrentRoom?.Transform`?

Beacuse, beacuse, beacuse.... Room value for dead players is the room where they died last. This means miscalculation.

PintTheDragon (approved):

PR #908, opened 2021-12-01
Server Freeze Detected Notification

ismail0234:
A notification message to detect if the server freeze is in progress


joker-119:
1. Why is this necessary?
2. Why is this not a debug message
3. No

PR #918, opened 2021-12-03
Make MaxHealth great again

joker-119:
Having MaxHealth tie directly to the Role.maxHP field makes it affect all players of that role. 
playerStats.StatModules[0].MaxValue is unsettable, thus we have no way to change a specific player's max-health within base-game code.

To allow us to still have and use the MaxHealth value for plugins, we need to detach it from base-game values. Plugins can still get/change the value as normal, and EXILED will change the value back to default for their new role when they change roles.


Jesus-QC:
Yeah, could be the best solution to avoid breaking changes

Jesus-QC (approved):

PintTheDragon (approved):

PR #919, opened 2021-12-03
Fix NW moment

joker-119:
Sometimes the host ReferenceHub gets destroyed when a round is ending, this causes the RoundSummary.RoundInProgress() call to hit a nullref which kills the coroutine responsible for ending the round and triggering a round restart.

This fix checks that the host ReferenceHub and it's associated CCM component are not null before running the above method, if one of them is null, the round will immediately end and a new round will be triggered as normal.


Jesus-QC (approved):

joker-119:
This appears to not solve the issue.

PR #922, opened 2021-12-03
Add new DamageType and rewrite for PocketDimension

louis1706:
- Add DamageType.CustomDamage
- Add DamageType.Scp096
- Rework DamageType.PocketDimension


Exiled.API/Features/DamageHandler.cs
```diff
@@ -91,13 +91,17 @@ public DamageType Type
...
+                            return DamageType.Scp;
                         case ExplosionDamageHandler _:
                             return DamageType.Explosion;
                         case Scp018DamageHandler _:
                             return DamageType.Scp018;
                         case RecontainmentDamageHandler _:
                             return DamageType.Recontainment;
+                        case CustomReasonDamageHandler_:
+                            return DamageType.CustomDamage;
+                        case Scp096DamageHandler _:
```
joker-119, line 103: I don't see why Scp096 needs his own damage type, this should return DamageType.Scp instead.

joker-119 (changes requested):

Exiled.API/Features/DamageHandler.cs
louis1706: is required an 
DamageType.Scp096
Because he use 
Scp096DamageHandler
And not
ScpDamageHandlers

Exiled.API/Enums/DamageType.cs
```diff
@@ -68,6 +68,10 @@ public enum DamageType
         Scp,
 
         /// <summary>
+        /// Damage dealt by SCP-096.
```
joker-119, line 71: This is a breaking change, do not add new values above existing ones in an enum.

joker-119 (changes requested):

Exiled.API/Features/DamageHandler.cs
joker-119: Yes so simple making Scp096DamageHandler return DamageType.Scp aswell.
He doesn't need his own DamageType enum value.

Exiled.API/Enums/DamageType.cs
louis1706: Oh okay i need to add at the ends it's that ?

joker-119:
Closed in favor of #923

PR #923, opened 2021-12-04
Handle damage type conversion more efficiently.

joker-119:
This also includes the changes from #922 in a non-breaking way.

### DIFF
--- a/Exiled.API/Enums/DamageType.cs
+++ b/Exiled.API/Enums/DamageType.cs
@@ -145,6 +145,6 @@ namespace Exiled.API.Enums
         /// <summary>
         /// Damage caused by <see cref="RoleType.Scp106"/>.
         /// </summary>
-        Scp106
+        Scp106,
     }
 }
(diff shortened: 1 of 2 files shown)


PintTheDragon (approved):
I'm going to say we should probably move away from general DamageTypes (ex. SCP, Firearm, etc) and make them more specific, just to make it easier for people to work with.

joker-119:
> I'm going to say we should probably move away from general DamageTypes (ex. SCP, Firearm, etc) and make them more specific, just to make it easier for people to work with.

We could, but I feel that adds more unnecessary work for us updating the values whenever they add a new gun or change the names of existing ones etc.

PintTheDragon:
> > I'm going to say we should probably move away from general DamageTypes (ex. SCP, Firearm, etc) and make them more specific, just to make it easier for people to work with.
> 
> We could, but I feel that adds more unnecessary work for us updating the values whenever they add a new gun or change the names of existing ones etc.

But then that change gets projected onto plugin devs. Isn't the whole point of this to avoid that? We could just add a new entry to the dictionary, or slightly modify one of the keys when this happens. Initial work might be a bit, but I don't really see any way that this can cause trouble with maintenance.

PintTheDragon (approved):

PR #924, opened 2021-12-04
Fix disconnects after rnr while using frr

PintTheDragon:
Basically, rnr is completely broken for servers that use frr, and it kicks everyone when the server restarts. This fixes that by forcing the server to use full restarts when the server is restarting.


joker-119 (approved):

EsserGaming:
Finally!

NaoUnderscore (approved):

PR #929, opened 2021-12-06
MaxHealth Changes

SebasCapo:
Quick draft I made for a possible fix regarding MaxHealth, I don't think it should break base-game code since CHS inherits the original HealthStat class.

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -63,6 +63,7 @@ namespace Exiled.API.Features
 
         private readonly IReadOnlyCollection<Item> readOnlyItems;
         private ReferenceHub referenceHub;
+        private CustomHealthStat healthStat;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Player"/> class.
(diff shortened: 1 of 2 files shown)


joker-119 (approved):

joker-119:
I feel like CustomMaxValue should probably default to something like float.NaN, rather than the standard float default (0.0f), but that's just me.

Exiled.API/Features/Player.cs
```diff
@@ -663,7 +666,11 @@ public float Health
         /// <summary>
         /// Gets or sets the player's maximum health.
         /// </summary>
-        public int MaxHealth { get; set; }
+        public float MaxHealth
```
joker-119, line 669: This should not change to an int, otherwise this becomes a breaking change.

joker-119 (changes requested):

Exiled.API/Features/Player.cs
SebasCapo: Some people wanna see the world burn

BuildBoy12 (approved):

PR #939, opened 2021-12-09
Fixed `StoppingEventArgs`

Marco15453:
Fixing `StoppingEventArgs` not being called while Warhead was locked

**This will make the StoppingEventArgs also being called when the warhead is locked and/or when the timer is under 10 seconds**


joker-119:
Why would we want the event to be called when the warhead timer is under 20s and can no longer be stopped?
Also why would we want this event called when the warhead is locked and cannot be stopped?

Exiled.API/Features/Items/Throwable.cs
```diff
@@ -61,7 +61,7 @@ public float PinPullTime
         /// Throws the item.
         /// </summary>
         /// <param name="fullForce">Whether to use full or half force.</param>
-        public void Throw(bool fullForce = true) => Base.ServerThrow(fullForce);
+        public void Throw(bool fullForce = true) => Base.ServerThrow(fullForce, ThrowableNetworkHandler.GetLimitedVelocity(Base.Owner.playerMovementSync.PlayerVelocity));
```
joker-119, line 64: Just re-merge dev into this PR instead of re-replicating changes.

Issue #952, opened 2021-12-18
[NRE] Player.Zone property throws NRE.


Pro845627:
System.NullReferenceException
                                   at (wrapper managed-to-native) UnityEngine.GameObject.CompareTag(UnityEngine.GameObject,string)
                                   at Exiled.API.Features.Map.FindParentRoom (UnityEngine.GameObject objectInRoom) [0x00008] in <a1737615b9034c54b5c3ad2db2d45a62>:0
                                   at Exiled.API.Features.Player.get_CurrentRoom () [0x00006] in <a1737615b9034c54b5c3ad2db2d45a62>:0
                                   at Exiled.API.Features.Player.get_Zone () [0x00000] in <a1737615b9034c54b5c3ad2db2d45a62>:0

EXILED VERSION: 4.1.6

Marco15453:
What is the complete console log?

Marco15453:
The bug report template exists for a reason, please include all of the information requested in it so we can see the actual exception.

joker-119:
> The bug report template exists for a reason, please include all of the information requested in it so we can see the actual exception.

Nice copy/paste :P

Marco15453:
:P

PR #957, opened 2021-12-20
Updated damage types to 11.1.1

Michal78900:
- Added `DamageType.Hypothermia`
- Added `DamageType.MolecularDisruptor`
- Removed giant if-return case which does not serve any purpose (all damage types are already in `translationConversion` dictionary, which means they would never be used)

### DIFF
--- a/Exiled.API/Features/DamageHandler.cs
+++ b/Exiled.API/Features/DamageHandler.cs
@@ -9,8 +9,6 @@ namespace Exiled.API.Features
 {
     using System.Collections.Generic;
 
-    using Dissonance;
-
     using Exiled.API.Enums;
     using Exiled.API.Features.Items;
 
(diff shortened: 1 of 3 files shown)


joker-119 (changes requested):
Cannot add an enum value anywhere both after the existing ones, otherwise this is a breaking change.

Michal78900:
> Cannot add an enum value anywhere both after the existing ones, otherwise this is a breaking change.

I don't understand, how adding the enum would be a breaking change? We are getting the damage type from dicitinary (hypothermia)/checking the handler type (disruptor) anyways. This would be a problem if these enums had numbers, but they don't and they aren't used that way.

joker-119:
> > Cannot add an enum value anywhere both after the existing ones, otherwise this is a breaking change.
> 
> I don't understand, how adding the enum would be a breaking change? We are getting the damage type from dicitinary (hypothermia)/checking the handler type (disruptor) anyways. This would be a problem if these enums had numbers, but they don't and they aren't used that way.

Then you don't understand how enums work.
Because you are changing the values of the enum. Enums are basically a list of numbers, and that's how the compiler sees them, so

`
EnumValue,
AnotherEnumValue,`

Is the same as
`EnumValue = 1
AnotherEnumValue = 2`

If you do a 
`
if (object.EnumProperty == EnumType.EnumValue)`
the compiler compiles it as
`if (object.EnumProperty == 1)`

When you add a 3rd value inbetween the 2 existing ones, you change the value of "AnotherEnumValue" from 2, to 3, but the plugins compiled before the change will not know the enum values have changed, because they are only compiled as numerical values.

Michal78900:
> > > Cannot add an enum value anywhere both after the existing ones, otherwise this is a breaking change.
> > 
> > 
> > I don't understand, how adding the enum would be a breaking change? We are getting the damage type from dicitinary (hypothermia)/checking the handler type (disruptor) anyways. This would be a problem if these enums had numbers, but they don't and they aren't used that way.
> 
> Then you don't understand how enums work. Because you are changing the values of the enum. Enums are basically a list of numbers, and that's how the compiler sees them, so
> 
> ` EnumValue, AnotherEnumValue,`
> 
> Is the same as `EnumValue = 1 AnotherEnumValue = 2`
> 
> If you do a ` if (object.EnumProperty == EnumType.EnumValue)` the compiler compiles it as `if (object.EnumProperty == 1)`
> 
> When you add a 3rd value inbetween the 2 existing ones, you change the value of "AnotherEnumValue" from 2, to 3, but the plugins compiled before the change will not know the enum values have changed, because they are only compiled as numerical values.
> 
> So if there's a plugin checking if a property is equal to "AnotherEnumValue", which at the time of compiling the plugin is '2', then we add a different value in the middle, '2' now becomes "ANewEnumValue", and that plugins logic check will bork.

I see, so I should put the new enums at the end?

joker-119:
> I see, so I should put the new enums at the end?

That is what I said, yes.

Exiled.API/Features/DamageHandler.cs
```diff
@@ -140,51 +139,31 @@ public DamageType Type
...
-                            if (translation.Id == DeathTranslations.SeveredHands.Id)
-                                return DamageType.SeveredHands;
-
-                            Log.Warn($"{nameof(DamageHandler)}.{nameof(Type)}: No matching {nameof(DamageType)} for {nameof(UniversalDamageHandler)} with ID {translation.Id}, type will be reported as {DamageType.Unknown}. Report this to EXILED Devs.");
-                            break;
-                        }
+                            {
+                                DeathTranslation translation = DeathTranslations.TranslationsById[universal.TranslationId];
+
+                                if (translationConversion.ContainsKey(translation))
```
PintTheDragon, line 161: We should still do ID checks. I'm honestly not very trusting of Northwood to not make a new instance of translations in some places, since they did that with the old DamageTypes.

PintTheDragon (changes requested):

Exiled.API/Features/DamageHandler.cs
Michal78900: Should I make 20 ifs for every single damage type then?

Michal78900: Because again, I don't see a reason since the dictionary exists and if new damage type is added it will fail anyway no matter if there are ifs or not.

joker-119: Whether or not we do the if-return tree comes down to if we trust NW or not, to continue using their dictionary properly.

joker-119:
Closing in favor of #952 957

PR #969, opened 2021-12-30
#957 but better

joker-119:


joker-119:
These changes will just get reverted in 2 days anyways.

PR #970, opened 2021-12-30
Maybe fix for disruptors

joker-119:
Still needs to be verified to actually fix the issues


joker-119:
Disruptors are disappearing in 2 days, not even worth addressing currently.

PR #973, opened 2021-12-31
More args in TriggeringTeslaEvent

louis1706:
- Add The Tesla
- Add The IdlingEffect


joker-119 (approved):
lgtm, the event patch here should be converted to a transpiler at some point though.

PR #976, opened 2022-01-01
Delay API

Thundermaker300:
Minor delay methods - methods that already exist, but feature an overload which includes a delay before calling the method. Thoughts on adding methods like this? It's a slight bit redundant, but it may help with new plugin devs, plus it would be more convenient for plugin devs to simply have the method to call, rather than adding MEC and all. The delay methods are overloads, so it won't clutter the list of methods for plugin devs either
I noticed that the Cassie class already has a delay method (which I've used before for my own plugins), so this may be a helpful addition to the rest of the framework
If you all like, I'll add more methods before requesting to merge pr

### DIFF
--- a/Exiled.API/Features/Door.cs
+++ b/Exiled.API/Features/Door.cs
@@ -277,9 +277,22 @@ namespace Exiled.API.Features
         }
 
         /// <summary>
+        /// Locks the door with the given lock type, after a given delay has passed.
+        /// </summary>
+        /// <param name="delay">Delay in seconds.</param>
+        /// <param name="lockType"><inheritdoc cref="DoorLockType"/></param>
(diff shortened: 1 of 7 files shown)


joker-119:
Personally I'd rather people just learn how to use Timing.CallDelayed() themselves, rather than us needing to bloat out our API with arbitrary method overloads that aren't necessary.

Thundermaker300:
I can understand that. I got the idea from the Cassie delay method lol

joker-119:
Idk why that even exists, ngl

Thundermaker300:
That's fair lol. I'll go ahead and close this and delete my branch, I just figured that the idea may be a useful alternative to referencing the assembly just for MEC and such

PR #979, opened 2022-01-02
Minor item changes

Thundermaker300:
- Added properties to Item class to identify what type of item it is (without casting)
- Added return types to a few methods in Scp330 (these methods previously returned void, so this shouldn't be breaking)
  - AddCandy now returns whether or not the candy was successfully added
  - RemoveCandy now returns the amount of candies that were removed
- Added return type to `Player.AddItem(ItemType itemType, int amount)`. and `Player.AddItem(List<ItemType> items)`
- Added `Ammo.AmmoLimit` - Constant set to `65535`, which is what the game will cap out ammo count


joker-119 (changes requested):
Changing a void method to have a return type is a breaking change because the method signature changes. These need to be made as overloads with out parameters, with the old methods marked obsolete, then when we version bump to 5.x we can remove the out parameters and make them return values instead like we did for custom items between 3.x and 4.x

Thundermaker300:
Understandable, I'll make the changes when I get on my computer later today

joker-119 (changes requested):
Include the true value in the obsolete attributes to enforce build errors.

`[Obsolete("Use xyz", true]`

Thundermaker300:
omg true!!!!!!!!!!!

Exiled.API/Features/Items/Ammo.cs
```diff
@@ -16,6 +16,11 @@ namespace Exiled.API.Features.Items
     /// </summary>
     public class Ammo : Item
     {
+        /// <summary>
+        /// Specifies the maximum amount of ammo that may be held at one time.
+        /// </summary>
+        public const ushort AmmoLimit = 65535;
```
joker-119, line 22: This is a breaking change.

joker-119 (changes requested):

Exiled.API/Features/Items/Ammo.cs
Thundermaker300: How so? It's a new property that didn't exist before.

joker-119: It's not, I'm retarded

joker-119 (approved):

joker-119 (changes requested):
Fix build errors

joker-119:
re-merge dev into this branch :)

joker-119 (approved):

BuildBoy12 (approved):

Exiled.API/Features/Items/Scp330.cs
```diff
@@ -65,22 +65,53 @@ public Scp330()
         /// Adds a specific candy to the bag.
         /// </summary>
         /// <param name="type">The <see cref="CandyKindID"/> to add.</param>
+        [System.Obsolete("Please use RemoveCandy(CandyKindID, out bool) instead.", true)]
```
Jesus-QC, line 68: Please use AddCandy

Jesus-QC (changes requested):

Exiled.API/Features/Items/Ammo.cs
```diff
@@ -16,6 +16,14 @@ namespace Exiled.API.Features.Items
     /// </summary>
     public class Ammo : Item
     {
+        /// <summary>
+        /// Specifies the absolute maximum amount of ammo that may be held at one time, if ammo is forcefully given to the player (regardless of worn armor or server configuration).
+        /// <para>
+        /// For accessing the maximum amount of ammo that may be held based on worn armor and server settings, see <see cref="Player.GetAmmoLimit(ItemType)"/>.
+        /// </para>
+        /// </summary>
+        public const ushort AmmoLimit = 65535;
```
Jesus-QC, line 25: what do you think about 
=> ushort.MaxValue

Jesus-QC (changes requested):

Exiled.API/Features/Items/Ammo.cs
joker-119: I agree this should be a property

Exiled.API/Features/Items/Scp330.cs
Thundermaker300: good catch

PR #984, opened 2022-01-03
Add Player.AddCandy

GabiRP:
* Added Player.AddCandy


joker-119:
This makes no sense, and for some reason doesn't even use our 330 api?
You're adding the candy and then removing it from the bag.
You're iterating over their Items list and using 2 GetEnumerators followed by enumerating over their items again if they do have the bag. The inefficiency is mindboggling.

PR #985, opened 2022-01-03
Optimise tesla event

louis1706:


louis1706:
(don't look all my stupid change) just look the file

joker-119:
Denying because I'm in the process of converting this patch to a transpiler anyways.

PR #989, opened 2022-01-04
Make cassie actually say his announcements if they aren't SCPs.

joker-119:
Human lives matter too, CASSIE!

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1431,7 +1431,13 @@ namespace Exiled.API.Features
         /// Hurts the player.
         /// </summary>
         /// <param name="damageHandlerBase">The <see cref="DamageHandlerBase"/> used to deal damage.</param>
-        public void Hurt(DamageHandlerBase damageHandlerBase) => ReferenceHub.playerStats.DealDamage(damageHandlerBase);
+        public void Hurt(DamageHandlerBase damageHandlerBase)
+        {
+            if (Health - ((StandardDamageHandler)damageHandlerBase).Damage < 1 && Side != Side.Scp && !string.IsNullOrEmpty(damageHandlerBase.CassieDeathAnnouncement))
+                Cassie.Message(damageHandlerBase.CassieDeathAnnouncement);
+
+            ReferenceHub.playerStats.DealDamage(damageHandlerBase);
+        }
 
         /// <summary>
         /// Hurts the player.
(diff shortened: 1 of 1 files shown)


Thundermaker300 (approved):

louis1706 (commented):
Why you not make an bool CassieOnlyForScp for the player wan't the Cassie only for the SCP ?

Michal78900 (approved):

joker-119:
> Why you not make an bool CassieOnlyForScp (default = true) for the player wan't the Cassie only for the SCP ?

Because that would require making another overload for the method just for that one thing. Instead they can just not set a cassie announcement string if they don't want a cassie announcement.

louis1706:
> > Why you not make an bool CassieOnlyForScp (default = true) for the player wan't the Cassie only for the SCP ?
> 
> Because that would require making another overload for the method just for that one thing. Instead they can just not set a cassie announcement string if they don't want a cassie announcement.

Yes but if the dev want to do an damage on everyone the SCP going to doo any annoncement  of there death beceause cassyannoncement is null and it's not going to doo the unknow death of the SCP ?

Exiled.API/Features/Player.cs
```diff
@@ -1431,15 +1431,27 @@ public bool RemoveItem(Item item, bool destroy = true)
         /// Hurts the player.
         /// </summary>
         /// <param name="damageHandlerBase">The <see cref="DamageHandlerBase"/> used to deal damage.</param>
-        public void Hurt(DamageHandlerBase damageHandlerBase) => ReferenceHub.playerStats.DealDamage(damageHandlerBase);
+        public void Hurt(DamageHandlerBase damageHandlerBase)
```
Jesus-QC, line 1434: This method shouldn't contain the cassie message.
It would make anything that uses this method and kills the player to make cassie say that.

Also: If for example you used the other hurt method you changed, and make cassie say something, it then will call this method and make cassie say the same. Therefore you will have repeated the same announcement.

Jesus-QC (changes requested):

Jesus-QC (approved):

Exiled.API/Features/Player.cs
Jesus-QC: ups, my fault

joker-119:
> > > Why you not make an bool CassieOnlyForScp (default = true) for the player wan't the Cassie only for the SCP ?
> > 
> > 
> > Because that would require making another overload for the method just for that one thing. Instead they can just not set a cassie announcement string if they don't want a cassie announcement.
> 
> Yes but if the dev want to do an damage on everyone the SCP going to doo any annoncement of there death beceause cassyannoncement is null and it's not going to doo the unknow death of the SCP ?

I don't believe that's how SCP death announcements work, i'm pretty sure if you leave it empty it will do the unknown death announcement, but I haven't tested it in-game.

MrNeonM:
Yay, my idea got added! 😊

PR #991, opened 2022-01-05
Camera Converting (ID -> Name)

Thundermaker300:
Setting up cameras to use names over camera IDs. Camera IDs can be changed unpredictably, and it's not possible to give certain rooms (halls, x rooms, etc) a CameraType due to Camera IDs. Camera names allow this to be possible.

- Added large dictionary that contains all camera names and their relevant CameraType
- Created method to search through the dictionary while also providing support for different cameras with the same name (via zone checking and "unspecified" enum values)
- Renamed a few cameras to more appropriate names **(BREAKING)**
- Removed now-unused SCP-012 camera types **(BREAKING)**
- Removed camera ID declarations from CameraType enums **(BREAKING)**

Every camera now has its own `CameraType` (believe it or not, this was not the case previously)!!!! Credits to @louis1706 for helping me confirm that I didn't miss any


joker-119 (approved):

Exiled.API/Features/Map.cs
```diff
@@ -457,6 +459,20 @@ public static Pickup GetRandomPickup(ItemType type = ItemType.None)
...
+        /// <summary>
+        /// Given the name of a camera, returns its <see cref="CameraType"/>, or <see langword="null"/> if the camera does not exist.
+        /// </summary>
+        /// <param name="cameraName">The name of the camera.</param>
+        /// <returns>The <see cref="CameraType"/>, or <see langword="null"/> if it does not exist.</returns>
+        public static CameraType? GetCameraType(string cameraName)
+        {
+            Camera079 cam = GetCameraByName(cameraName);
+            if (!cam)
+                return null;
```
louis1706, line 471: Don't make it Null
You can just return string.Empty

Thundermaker300: This doesn't compile, `string.Empty` can't be casted to a `CameraType`. Although I can make it return `CameraType.Unknown` instead of `null`.

louis1706: Yes sorry i am dumb

PR #1001, opened 2022-01-06
API Improvements

SebasCapo:
- Added `Unlock(float, DoorLockType)` & `Lock(float, DoorLockType)` to Door class.
These were made both because there's a base-game extension that supports this, and because it eases work when locking doors for a specific amount.
*A few notes, I'm thinking of renaming `Lock()` to `ChangeLock()` so it stays consistent with the previous method, and I'm still deciding whether the `Lock()` method is necessary at all.*

- Added `Log.DebugObject()` to make debugging easier.
This was added so you can easily debug objects inputted inside parameters.
Example: `Player.List.Where(p => Log.DebugObject(p.IsDead));`

- Added `Player.GetModule<T>()` for easy accessing, instead of having to use `Player.ReferenceHub.playerStats`.

### DIFF
--- a/Exiled.API/Features/Log.cs
+++ b/Exiled.API/Features/Log.cs
@@ -39,15 +39,15 @@ namespace Exiled.API.Features
         /// Server must have exiled_debug config enabled.
         /// </summary>
         /// <typeparam name="T">The inputted object's type.</typeparam>
-        /// <param name="message">The message to be sent.</param>
+        /// <param name="object">The object to be logged and returned.</param>
         /// <param name="canBeSent">Indicates whether the log can be sent or not.</param>
-        /// <returns>Returns the <typeparamref name="T"/> object inputted in <paramref name="message"/>.</returns>
-        public static T DebugObject<T>(T message, bool canBeSent = true)
+        /// <returns>Returns the <typeparamref name="T"/> object inputted in <paramref name="object"/>.</returns>
+        public static T DebugObject<T>(T @object, bool canBeSent = true)
         {
             if (canBeSent)
-                Send($"[{Assembly.GetCallingAssembly().GetName().Name}] {message}", LogLevel.Debug, System.ConsoleColor.Green);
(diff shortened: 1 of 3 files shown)


Exiled.API/Features/Log.cs
```diff
@@ -34,6 +34,22 @@ public static void Debug(object message, bool canBeSent = true)
...
 
+        /// <summary>
+        /// Sends a <see cref="LogLevel.Debug"/> level messages to the game console.
+        /// Server must have exiled_debug config enabled.
+        /// </summary>
+        /// <typeparam name="T">The inputted object's type.</typeparam>
+        /// <param name="object">The object to be logged and returned.</param>
+        /// <param name="canBeSent">Indicates whether the log can be sent or not.</param>
+        /// <returns>Returns the <typeparamref name="T"/> object inputted in <paramref name="object"/>.</returns>
+        public static T DebugObject<T>(T @object, bool canBeSent = true)
```
Thundermaker300, line 45: This method isn't necessary, the base `Log.Debug` message takes an `object` parameter, which already allows anything to be passed.

Thundermaker300 (changes requested):

Exiled.API/Features/Log.cs
SebasCapo: I am aware of that method, as I explained on the PR's description, this is eases testing a lot by allowing you to do things like `Player.List.Where(p => Log.DebugObject(p.IsDead));` without having to create an entire foreach loop that you'll then have to remove.

It is obviously not **necessary**, but a simple tool that could help with testing in some cases.

joker-119: `List.Where()`
would be the same as
`
    if (player.IsDead)`
but with an extra GetEnumerator for seemingly no reason.

SebasCapo: What I meant wasn't that this is supposed to be something to solve an issue currently present, it's mainly just a small niche feature that could make debugging things a little simpler, there's many cases where this could be applied.

PR #1011, opened 2022-01-09
Add DamageType extensions

Remuchi:


Remuchi:
Idk if I should add scp-207 damagetype to IsScp or not

Thundermaker300:
You should definitely add SCP-207 DamageType to IsScp, because it is damaged that is caused by an SCP (it'd be weird if it wasn't considered SCP damage). You could also add it to IsStatusEffect, but I'm not sure about that one

Remuchi:
Maybe there should be IsPlayableScp and IsScp. IsPlayableScp only has playable SCPs like SCP106

louis1706:
> Maybe there should be IsPlayableScp and IsScp. IsPlayableScp only has playable SCPs like SCP106

For IsScp you Can look with string if it the damagetype contain SCP idk if it more optimise

Like 
return damagetype.ToString().contains("Scp");

Remuchi:
I don't think that returning a string value of enum type and when doing string search is more optimal than just bunch of byte(or whatever enum uses) comparisons

louis1706:
> I don't think that returning a string value of enum type and when doing string search is more optimal than just bunch of byte(or whatever enum uses) comparisons 

Idk

joker-119:
If you add SCP items to the IsScp one, I think there should be a bool that defaults true, and when it's false, the extension won't return true for the items, similar to our IsWeapon() extension for ItemTypes (ItemExtensions.IsWeapon) has one for excluding the MicroHid from the check.

PR #1012, opened 2022-01-09
Re-work item API.

joker-119:
Item.Create() should now be used by all plugins, while constructors will be reserved for EXILED internally.

This allows a single line of code to handle creation of items of any subtype with the correct subclass attached (resolves issues where gun ammo breaks etc if you simple `new Item(ItemType.E11SR)` instead of `new Firearm(ItemType.E11SR)`, amongst others)

Obsolete marked ctors will be made internal only in the future, as to alleviate all plugins breaking instantly when 5.0 releases.

### DIFF
--- a/Exiled.API/Features/Items/Ammo.cs
+++ b/Exiled.API/Features/Items/Ammo.cs
@@ -7,6 +7,8 @@
 
 namespace Exiled.API.Features.Items
 {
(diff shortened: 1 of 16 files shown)


Jesus-QC (approved):
pretty cool

PR #1014, opened 2022-01-09
Make Player.Heal() respect negative values

rayzerbrain:
Problem is NW ServerHeal method converts amount to an absolute value.
Player.Health exists of course but doesn't have the option to not override max health in case the value being added is positive


joker-119:
But a negative value here would just hurt them, wouldn't it? Why not just use player.Hurt in that case?

rayzerbrain:
That's the thing, I don't know if the value will be positive or not.

joker-119:
Just use if (value < 0) yourself if you need to deal with negatives. This changes the core behavior of this method, it would no longer make sense to be called Heal() rather than ChangeHealth()

PR #1018, opened 2022-01-12
Allow for proper deserialization of the List<CustomAbility> property.

joker-119:


Exiled.CustomRoles/API/Features/Extensions/ParserExtension.cs
```diff
@@ -0,0 +1,66 @@
...
+    public static class ParserExtension
+    {
+        /// <summary>
+        /// Tries to find a valid mapping entry.
+        /// </summary>
+        /// <param name="parser">The <see cref="ParsingEventBuffer"/> parser.</param>
+        /// <param name="selector">The selector.</param>
+        /// <param name="key">The key found.</param>
+        /// <param name="value">The value found.</param>
+        /// <returns><see langwod="true"/> when a valid mapping entry is found.</returns>
```
NaoUnderscore, line 30: 1. Fix typo in "langwod"
2. `<see langword="true"/> when a valid mapping entry is found; otherwise, <see langword="false">.` is better, since we should use .NET and C# docs style pattern.

Exiled.CustomRoles/API/Features/Interfaces/ITypeDiscriminator.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// Gets the base <see cref="Type"/>.
+        /// </summary>
+        Type BaseType { get; }
+
+        /// <summary>
+        /// Tries to resolve a mapping into a specific <see cref="Type"/>.
+        /// </summary>
+        /// <param name="buffer">The <see cref="ParsingEventBuffer"/> parser buffer.</param>
+        /// <param name="suggestedType">The <see cref="Type"/> to resolve the mapping key.</param>
+        /// <returns><see landwog="true"/> if resolution is successful.</returns>
```
NaoUnderscore, line 29: 1. Fix typo in "landwog"
2. `<see langword="true"/> if resolution is successful; otherwise, <see langword="false">.` is better, since we should use .NET and C# docs style pattern.

NaoUnderscore (changes requested):
I didn't mark the CustomAbility constructor as error, because I'd like you to use lambda =>, since you're setting only a value. But it's still fine.

NaoUnderscore (approved):

PR #1021, opened 2022-01-12
Changes to Player.Hurt to make them more useful.

joker-119:
Re-works existing Player.Hurt methods and adds 2 new overloads, which eliminates the repeated code in them and makes it overall more useful for plugin devs.

### DIFF
--- a/Exiled.API/Features/ExiledDamageHandler.cs
+++ b/Exiled.API/Features/ExiledDamageHandler.cs
@@ -1,0 +1,66 @@
+// -----------------------------------------------------------------------
+// <copyright file="ExiledDamageHandler.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Features
+{
+    using Footprinting;
+
+    using PlayerStatsSystem;
+
+    /// <summary>
+    /// A special <see cref="AttackerDamageHandler"/> that allows you to define the attacker.
+    /// </summary>
+    public class ExiledDamageHandler : AttackerDamageHandler
+    {
+        private Footprint attacker;
+        private float damage;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExiledDamageHandler"/> class.
+        /// </summary>
+        /// <param name="player">The <see cref="Player"/> who is dealing the damage.</param>
+        /// <param name="amount">The amount of damage to deal.</param>
+        /// <param name="cassieAnnouncement">The announcement cassie will make if this damage kills the victim.</param>
+        /// <param name="logText">The text to appear in server logs.</param>
+        /// <param name="allowSelfDamage"><see langword="true"/> if the damage can be dealt to the attacker. Otherwise <see langwod="false"/>.</param>
+        public ExiledDamageHandler(Player player, float amount, string cassieAnnouncement = "default", string logText = "default", bool allowSelfDamage = true)
+        {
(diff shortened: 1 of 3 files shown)


NaoUnderscore:
Seems good, but isn't there any way to re-make the old `Player::Hurt` method with Player parameter?
What if people want to force Player1 to hurt Player2?

joker-119:
The only way to do that would be with an AttackerDamageHandler, which is abstract, the only classes that inherit it are Explosive, Firearm, Scp, MicroHid, Recontainment and Scp018

We'd need to make our own special damage handler and then hope it doesn't break anything, if they don't want to just new FirearmDamageHandler() or something, but it would be possible.

Exiled.API/Features/ExiledDamageHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    using PlayerStatsSystem;
+
+    /// <summary>
+    /// A special <see cref="AttackerDamageHandler"/> that allows you to define the attacker.
+    /// </summary>
+    public class ExiledDamageHandler : AttackerDamageHandler
+    {
+        private Footprint attacker;
+        private float damage;
```
Michal78900, line 20: Why do these exist? Can't both `Attacker` and `Damage` properties made as `{ get; set; }` like they are made in base-game?

NaoUnderscore:
> The only way to do that would be with an AttackerDamageHandler, which is abstract, the only classes that inherit it are Explosive, Firearm, Scp, MicroHid, Recontainment and Scp018
> 
> We'd need to make our own special damage handler and then hope it doesn't break anything, if they don't want to just new FirearmDamageHandler() or something, but it would be possible.

If you're able to do that, I'd like you to do that. Otherwise I'm gonna find a solution in the next days, because I think it's a really important thing to add.

Exiled.API/Features/ExiledDamageHandler.cs
joker-119: Because calling virtual methods (property setters are in fact methods) in a constructor is typically a bad idea. 

In this particular case there's no issue since we know ExiledDamageHandler is the last inherited type, and that the 'methods' won't access things that could be null/etc in the inherited type, so it would be safe, but 1. Better to do it the right way.
2. The compiler will just do this itself anyways
3. Better to be safe than sorry later on if something changes.

joker-119:
> > The only way to do that would be with an AttackerDamageHandler, which is abstract, the only classes that inherit it are Explosive, Firearm, Scp, MicroHid, Recontainment and Scp018
> > We'd need to make our own special damage handler and then hope it doesn't break anything, if they don't want to just new FirearmDamageHandler() or something, but it would be possible.
> 
> If you're able to do that, I'd like you to do that. Otherwise I'm gonna find a solution in the next days, because I think it's a really important thing to add.

I already did, look at the latest commit

Exiled.API/Features/ExiledDamageHandler.cs
Michal78900: I see

NaoUnderscore:
> > > The only way to do that would be with an AttackerDamageHandler, which is abstract, the only classes that inherit it are Explosive, Firearm, Scp, MicroHid, Recontainment and Scp018
> > > We'd need to make our own special damage handler and then hope it doesn't break anything, if they don't want to just new FirearmDamageHandler() or something, but it would be possible.
> > 
> > 
> > If you're able to do that, I'd like you to do that. Otherwise I'm gonna find a solution in the next days, because I think it's a really important thing to add.
> 
> I already did, look at the latest commit

Gotcha, seems fine to me 👍

NaoUnderscore (approved):

PR #1022, opened 2022-01-12
Global variable for Log::Debug state

Klarulor:
Added global variable for logging Log::Debug


joker-119:
I cannot express how horrible of an idea this is, as it would force all debug messages to be enabled for all plugins universally even if you're just needing them for one.

PR #1023, opened 2022-01-13
Revert "Merge pull request

4310V343k:
I was one moment away from creating a PR into master
Notable changes from joker's impl.: 
- Docs are better and more meaningful (i hope)
- Different class for extensions of ParsingEventBuffer is not needed
- Deserializer is changed after loading the plugin assemblies, but before loading configs. This makes possible to store the configs in the main config file (cool, because we have separate configs now. who doesn't use them anyway?)
- Reflection


4310V343k:
Ouch, i thought i changed that

joker-119:
Is there a reason we are using reflection here when it's entirely unnecessary to do so?
Also point of fact -
> Deserializer is changed after loading the plugin assemblies, but before loading configs. This makes possible to store the configs in the main config file (cool, because we have separate configs now. who doesn't use them anyway?)

Mine did this already.

4310V343k:
Originally i wanted to leave Loader.Deserializer a get-only property, but seeing you simply allowing setting it removes the need for reflection at all

PR #1024, opened 2022-01-13
Handcuff dont work normally fixed

ismail0234:
The following solves the problem.
- **Removing Handcuffs Event Args doesn't have player who uncuffed another player**


ismail0234:
Discord Bug Report Url: https://discord.com/channels/656673194693885975/664263236048388097/931251412636143616

NaoUnderscore:
As long as it works, it's fine for me.
I'll convert it to transpiler soon, but that was an annoying bug.

Issue #886, opened 2021-11-21
[BUG] PickingUpItemEventArgs.IsAllowed doesn't work as expected


SebasCapo:
**Describe the bug**
Setting `PickingUpItemEventArgs.IsAllowed` to false makes the item unable to be grabbed after your first attempt at doing so. (This also causes weird behaviors when trying the same thing on a custom item)

**To Reproduce**
1. Subscribe a method to the `PickingUpItemEventArgs`, doesn't need anything else other than `ev.IsAllowed` being set to `false`.
2. Attempt to grab an item.
3. Try grabbing that exact item again.
4. Item becomes pseudo-locked.

**Expected behavior**
Be able to attempt to grab this item, this is important to allow custom behavior to happen when player interacts with it, something that is not possible when setting `Pickup.Locked` to `true`.

**Server logs**
Not necessary

**EXILED Version ("latest" is not a version):**
3.7.2

**Results of `show plugins` command in console:**
Not necessary

**Additional context**
Funnily enough, this works differently on custom items (Mainly due to how the base `CustomItem.cs` pickup logic was made), instead of locking the item, it just lets you grab it. Ideally this would also be fixed in the same PR, but we can't expect god to do all the work. :')

zhaguanyang:
Why are u setting ``Pickup.Locked`` to ``true`` if you want people to be able to pick it up?
Try ``Pickup.InUse = false;`` and ``Pickup.Locked = false;`` (should be one of them tho)

SebasCapo:
> Why are u setting `Pickup.Locked` to `true` if you want people to be able to pick it up?

That is not at all what I said though, I simply mentioned why it's important to fix this, as it's a feature that used to work pre-Parabellum and setting `Pickup.Locked` to true wouldn't be achieve the same effect.

> Try `Pickup.InUse = false;` and `Pickup.Locked = false;` (should be one of them tho)

I will try this and see if it works, though it doesn't fix the issue I'm having. Nice thing about fixing the event is that you can have behavior work differently for specific players when, for example, certain conditions are met.

joker-119:
Can someone re-verify if this bug still exists please?

SebasCapo:
Just tested it on 4.1.7, seems to have been fixed, honestly unsure as to why as I don't see anything on the changelog referring to this, must've been base-game? Maybe? Will check again when we rollback to 11.1.0.

Thank you for looking into it regardless. <3

PR #1029, opened 2022-01-14
Fix Pickup Locking when ev.IsAllowed = false

rayzerbrain:
**PickingUpItemEventArgs.IsAllowed = false** caused the Pickup to permanently lock for some reason. Resetting aspects of the PickupSyncInfo seems to solve the problem. Tested by myself but could be problems I'm unaware of.

~~Not Invalid IL code I promise~~


NaoUnderscore:
"Avoiding the problem" is not a valid solution.
I'm pretty sure you're gonna break CustomItems.
Also, this could lead to many issues, and since you also said "but could be problems I'm unaware of" you already know that this "solution" is gonna lead to many issues instead.
Before approving it I need to make sure it doesn't break anything, although I'm sure it's not the right way to cope this problem.
I'd like to handle this issue, because I'm sure that I can fix it, but as I already said before, this is not the right way to solve this issue.

rayzerbrain:
That makes sense to me, I'll try poking around a bit more for a better solution, unless you want to take care of it instead?

rayzerbrain:
Don't ask me why this works, but it does.

It seems redundant but I did lots of testing and this is the easiest way it seems to work.

Exiled.Events/Patches/Events/Player/PickingUpItem.cs
```diff
@@ -37,6 +37,7 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
             Label cdcLabel = generator.DefineLabel();
 
             LocalBuilder ev = generator.DeclareLocal(typeof(PickingUpItemEventArgs));
+            LocalBuilder syncInfo = generator.DeclareLocal(typeof(PickupSyncInfo));
```
NaoUnderscore, line 40: Remove it

NaoUnderscore (changes requested):

SebasCapo:
When testing these changes, make sure they don't cause #886 to happen again! 😅

rayzerbrain:
That's actually the exact issue I'm trying to fix. Either the NW code changed in the past month to reintroduce the bug or it's only happening in specific situations or something. Can you verify that you currently do or do not have this issue?

joker-119:
The description of your fix and what the code actually does do not match. You're not spawning a new item here anywhere.

rayzerbrain:
> The description of your fix and what the code actually does do not match. You're not spawning a new item here anywhere.

That was my previous method of fixing the problem. My new solution is monkeying around with the PickupBase info, which somehow works

joker-119:
My point was, you should update your first comment to reflect what's actually going on since they no longer match.

joker-119 (approved):

NaoUnderscore (approved):

PR #1037, opened 2022-01-16
New event system

Mooshua:
### ***Untested*** and most certainly buggy as for the love of god I could *not* get this thing to build. Will move from draft to PR when ready.

**What?**
This is a new, backwards compatible event system which uses a cool attribute to discover event subscriptions. Certainly not as fast, but definitely easier to use.

**Why?**
Currently, while the event system is _fast_, it's cumbersome to use and allows people to quickly create code which violates EXILED best practices. This change makes it easier to develop plugins, and also reduces bugs from reload-ability issues during development.

**TL;DR of changes**
1. New event system which uses reflection and a `[Subscribe]` attribute to automatically collect events.
2. Old event system keeps working due to a backwards compatibility layer
3. Uses the type of the first argument to determine what event a `[Subscribe]`'d method belongs to.
4. Registers event on plugin load by searching all classes in that plugin's assembly

**Possible Improvements:**
 - You can make the event manager append to a delegate chain rather than the current iterative system
 - More safety & idiot proofing to lower the IQ needed to use this.
 - Change the subscription collector from `Dictionary<Assembly,...>` to `Dictionary<...>` and use a struct to carry assembly information.

**Sample**
```cs
class EventHandlers
{
    [Subscribe]
    void OnPlayerDied(DiedEventArgs ev)
    {
       Console.WriteLine($"Someone named {ev.Target.Nickname} was killed!")
    }
}
```
A different implementation of this which has been tested can be seen [at this gist.](https://gist.github.com/Mooshua/13f6472449e21434aa08a0f5b4479ab7)


NaoUnderscore:
Let me understand.

This should be "easier to use"? What's the difficulty of using a basic C# feature such as "Events"?
`Events.Event +=/-= EventHandler.OnEvent` is already anti-monkey.
If people can't register events, it's not our problem. They shouldn't even code plugins at this point.
You want an easier solution? Tell everyone to learn coding properly instead of relying on things like that, which is just a way to be different, but still monkey.
We're already using `ExiledSerializableAttribute` for CustomItems, CustomRoles and CustomAbility. It's a good solution because they're not events, and we needed to make huge methods to register/unregister custom things before, which was hell for maintainability index, and that wasn't the right way to handle those things anyway.
But there's no need to make attributes for events. They already work properly.

Also, backwards compatibility is useless. If you want to make a new event system, just break the old one and remove useless code. It doesn't make sense what you're doing.

Anyway, this solution is slow and buggy as hell. I don't recommend you to keep developing this draft if this is the way you want to follow, because I am not going to approve it, no matters what.

Mooshua:
> What’s the difficulty of using a basic C# feature such as events?

There is none. Having a centralized event system poses it’s own unique benefits unseen by traditional C# events, and allows plugins to collaborate in new ways.

High-level events may be emitted on the foundation of lower ones. Increasingly modded servers may benefit from better organization of code. Consider a teleporter mod, which adds a teleporter between the nuke room and the nuke surface cubby. `UsedPortalArgs` would be used to signal that a player should be moved. 

A stat-tracking plug-in would just need to reference the `UsedPortalArgs` type to seamlessly integrate with this plug-in. Or the other way around—`StatTrackArgs` to be invoked with a name and lambda modifier.

It is **possible** for old plugins to run on new versions of EXILED where there are direct compatibility breaks by using different event namespaces and bridging any required changes to the old arg specs. This could give servers time to update plugins during the chaotic days after release, while also breathing life into some legacy plugins which haven’t been updated since <10.0.

A plug-in ecosystem should always be concerned about how plugins interact with each other—even if it is not the primary goal of the system. Think about people using RainbowTags as a patreon benefit!

> Also, backwards compatibility is useless.

Good to know. I thought some would appreciate the gesture, but it is a lot more dead code flying around in an already massive file.

> Anyway, this solution is slow and buggy as hell.

Undoubtedly. This I hope to improve, whether by expressions or IL generation. Specifying generic variance should be added, and everything should become more bulletproof with coming revisions, if any.

Thank you for your feedback! I understand the friction here, and I hope to present a solid contender once the quality of this code has improved drastically :)

Jesus-QC:
@Mooshua where is the backwards compatible part:

You still need to use event arguments into the methods, which if are removed or renamed it will give you a typeload error, the same as the actual exiled.

Also, it is slow as fuck and made for lazy people, not the best for exiled.

Mooshua:
> Also, it is slow as fuck

_Please profile code before complaining about how slow it is._ On my local tree, performance is near identical after some common sense improvements. Meanwhile, EXILED's 'faster' version had a glaring unsunken allocation when retrieving a type's full name, as type names are not cached. The erring code is located [here](https://github.com/Exiled-Team/EXILED/blob/44ce8f1a8e6aa317c1f13b366ed123817d3d9b7a/Exiled.Events/Extensions/Event.cs#L32) and [here](https://github.com/Exiled-Team/EXILED/blob/44ce8f1a8e6aa317c1f13b366ed123817d3d9b7a/Exiled.Events/Extensions/Event.cs#L56)

**Time, in ms, to execute 10,000,000 simple events.** *Lower is better.*

| Run # | 1 | 2 | 3 | 4\* |
| --- | --- | -- | -- | -- |
| new EventManager | 582 | 596 | 606 | 277
| old InvokeSafely | 587 | 629 | 597 | 279
| old InvokeSafely after sunken allocation | 387 | 394 | 381 | 274

*\*Number 4 had 100,000 events and a 50% error chance to demonstrate error handling after sinking.*

Given that this performance is what people have been used to for a while, even though it is not the theoretical best, I'd say the improved experience for plugin developers is 100% worth the slowdown.

> You still need to use event arguments into the methods, which if are removed or renamed it will give you a typeload error, the same as the actual exiled.

That's true. This event system isn't meant to be a whole new plugin framework, but an easier way for plugins to collaborate, an easier plugin development experience, and a new way to implement backwards compatibility with otherwise fine plugins.

joker-119:
I'm not really sure what the benefit is here, as you describe 2 plugins interacting with new event args for something, this is already possible, and would be functionally the same as your example.
Teleport plugin would need to create an event that has some event args, that is invoked when the player uses the teleporter. Other plugins would need to reference that plugin and subscribe to the event to use it for whatever it is they want to use it for.

All i see here is the way events are subscribed being changed from the standard C#
SomeEvent += someMethodToHandleThatEvent
to
[Subscribe]
someMethodToHandleThatEvent

Jesus-QC:
First, I di#nt meant slow when handling events but when sibscribing to them, it needs to find all methods in all classes in an assembly which have that attribute.

Jesus-QC:
Secondly, you were telling us that it has backward compatibility but it doesnt.

joker-119:
agreed, I see nothing here that helps with any backwards compatability, just slower event subscription and changing how events are subscribed

Mooshua:
> All i see here is the way events are subscribed being changed from the standard C#

Yep, **that's pretty much the only change!** 😄 
The bigger benefit, I believe, is being able to disassociate *where* the event comes from. You can effectively hide events while still keeping binary compatibility.

This has the additional benefit of being able to polyfill functionality which was removed or modified in newer versions of EXILED (this is where the 'backwards compatibility' comes in.) If dice replaced coins, `FlippingCoinEventArgs` could tell outdated plugins if the rolled number was even or odd instead of tails/heads.

This is not to say that this means old plugins *will* work on new EXILED--As Northwood makes breaking type changes all the time--But it *does* mean that critical plugins can get online sooner when these changes do occur.

> just slower event subscription

Event subscription experiences a massive slowdown, that's true. But I don't think it's really all that relevant--When we're loading plugins, whether reloading them or launching the server--we already iterate over every type until we find a plugin, on top of some YAML parsing. Given that this only happens once in a plugin's active lifetime, it really shouldn't have too much of an impact on anything.

Jesus-QC:
It wont find the eventarg class therefore it will give a typeload exception, not sure what are you talking about really.

Jesus-QC:
FlippingCoinsEventArgs is removed => typeload exception, where is the backwards compatibility part
FlippingCoinsEventArgs has a new `PlayerSmt` field => typeload exception

It doesn't change anything

Jesus-QC:
Not only that but is making people to dont use OnEnabled and make everything slower

joker-119:
The thing is, this doesn't actually change backwards compatability anyways.
In your example where Dice replace coins, if we remove flippingcoineventargs with rollingdiceeventargs, plugins will his a typeload exception either way.
We would need to keep flipping coin eventargs and have the dice roll event implement both, which we can already do to avoid breaking plugins just fine.

All it does is change (and slowdown) how events are subscribed.

Jesus-QC:
^

PR #1039, opened 2022-01-16
Add TargetOldRole to Died event

GabiRP:


joker-119 (approved):

NaoUnderscore:
We already have DyingEventArgs for that.
Also, DiedEventArgs isn't meant to be used in this way, because as I already said, DyingEventArgs already exists and it offers more things. If you need to know the old player's role, it's because you don't need to use DiedEventArgs.

PR #1043, opened 2022-01-17
VeryImportant Repair bug on Hurt

louis1706:
It's going to be needed on the Master


joker-119:
Not needed on master since this isn't used on master

PR #1046, opened 2022-01-17
Captions

Thundermaker300:
This PR will remain a draft until Northwood fixes `SubtitleType.Custom`, as it does not currently function and the provided methods rely on it. In the meantime, I'd appreciate comments


Thundermaker300:
Given that it is now a VSR violation to use subtitles in forms other than CASSIE, I will be closing this pull request. It was fun while it lasted

PR #1048, opened 2022-01-18
Add Room in Generator

louis1706:
- just adding Room

Co-Authored-By: Nao <60253860+NaoUnderscore@users.noreply.github.com>

### DIFF
--- a/Exiled.API/Features/Generator.cs
+++ b/Exiled.API/Features/Generator.cs
@@ -36,6 +36,11 @@ namespace Exiled.API.Features
         public Scp079Generator Base { get; }
 
         /// <summary>
+        /// Gets the generator's <see cref="Room"/>.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Generator.cs
```diff
@@ -35,6 +35,11 @@ public class Generator
         /// </summary>
         public Scp079Generator Base { get; }
 
+        /// <summary>
+        /// Gets the generator's <see cref="Room"/>.
+        /// </summary>
+        public Room Room => Map.FindParentRoom(Base.gameObject);
```
NaoUnderscore, line 41: `Base.gameObject` is not needed, use `Generator::GameObject` instead.

NaoUnderscore (changes requested):

NaoUnderscore (approved):

PR #1052, opened 2022-01-19
Add door's euler angles.

Ceglaa:


NaoUnderscore:
Euler angles is basically Quaternion to Vector3, which is the only way to represent a quaternion with a non-quaternion data type. You could just get the euler angles from the `Door::Rotation`.
Would it be useful? Yes.
Is it needed? No.

joker-119:
This is literally just Door.Rotation but as a different data type

PR #1053, opened 2022-01-19
Added a few Player::Kill overloads and fixed cassie announcements in ExiledDamageHandler

Remuchi:
Now it works similar to Hurt methods and you can use DamageType enum to kill player


Exiled.API/Features/Player.cs
```diff
@@ -1500,16 +1500,36 @@ public void Heal(float amount, bool overrideMaxHealth = false)
...
+        /// <param name="damageHandlerBase">The <see cref="DamageHandlerBase"/> used to kill.</param>
+        public void Kill(DamageHandlerBase damageHandlerBase) => Kill(new DamageHandler(this, damageHandlerBase));
+
+        /// <summary>
+        /// Kills the player.
+        /// </summary>
+        /// <param name="damageType">The <see cref="DamageType">damage type</see> the player has been killed with.</param>
+        /// <param name="cassieAnnouncement">The cassie announcement to make upon death.</param>
+        /// <param name="attacker">The <see cref="Player"/> who killed player.</param>
+        public void Kill(DamageType damageType = DamageType.Unknown, string cassieAnnouncement = "", Player attacker = null) => Kill(new ExiledDamageHandler(attacker, float.MaxValue, cassieAnnouncement, DamageHandler.TranslationConversion.FirstOrDefault(k => k.Value == damageType).Key.LogLabel));
```
Thundermaker300, line 1524: I feel like the `attacker` parameter should come before the `cassieAnnouncement` parameter. Makes more logical sense to me, the attacker is more likely to be defined than a CASSIE announcement. Although other feedback on this would be appreciated

NaoUnderscore (changes requested):
Move the logic of `Player::Kill(DamageHandler)` inside the new method, and then use `DamageHandler::Base`. In this way we don't have to `newobj` every time.
Add more overloads to easily handle `Player` as parameter.
A quick workaround could be:
`Player::Kill(Player, [optional]DamageType, [optional]string)`

Remuchi:
At this moment this PR contains breaking changes as`Player::Kill(string reason, string cassieAnnouncement)` method was removed in favor of `Player::Kill(string reason, Player killer, string cassieAnnouncement)` method.

Remuchi:
Also as custom subtitles don't work I can't add them to cassie. The closed thing I can make is to use Sanyae's method to display them but then subtitles appear to high on the screen. (CASSIE color can be changed.)
image

NaoUnderscore (approved):

Exiled.API/Features/Player.cs
```diff
@@ -1500,16 +1500,37 @@ public void Heal(float amount, bool overrideMaxHealth = false)
...
+        /// <param name="damageHandlerBase">The <see cref="DamageHandlerBase"/> used to kill.</param>
+        public void Kill(DamageHandlerBase damageHandlerBase)
         {
-            if (Side != Side.Scp && !string.IsNullOrEmpty(cassieAnnouncement))
-                Cassie.Message(cassieAnnouncement);
+            if (Side != Side.Scp && !string.IsNullOrEmpty(damageHandlerBase.CassieDeathAnnouncement.Announcement))
+                Cassie.Message(damageHandlerBase.CassieDeathAnnouncement.Announcement);
 
-            ReferenceHub.playerStats.KillPlayer(new CustomReasonDamageHandler(deathReason, float.MaxValue, cassieAnnouncement));
+            ReferenceHub.playerStats.KillPlayer(damageHandlerBase);
```
Michal78900, line 1509: We shouldn't use that method directly. We should use `DealDamage` with damage value set to `-1f`.
```cs
public void Kill(DamageHandlerBase damageHandlerBase)
{
    if (Side != Side.Scp && !string.IsNullOrEmpty(damageHandlerBase.CassieDeathAnnouncement.Announcement))
        Cassie.Message(damageHandlerBase.CassieDeathAnnouncement.Announcement);

    (damageHandlerBase as StandardDamageHandler).Damage = -1f;
    ReferenceHub.playerStats.DealDamage(damageHandlerBase);
}
```

Exiled.API/Features/Player.cs
```diff
@@ -1500,16 +1500,37 @@ public void Heal(float amount, bool overrideMaxHealth = false)
...
+        /// <param name="handler">The <see cref="DamageHandler"/> used to kill.</param>
+        public void Kill(DamageHandler handler) => Kill(handler.Base);
+
+        /// <summary>
+        /// Kills the player.
+        /// </summary>
+        /// <param name="damageType">The <see cref="DamageType">damage type</see> the player has been killed with.</param>
+        /// <param name="killer">The <see cref="Player"/> who killed player.</param>
+        /// <param name="cassieAnnouncement">The cassie announcement to make upon death.</param>
+        public void Kill(DamageType damageType = DamageType.Unknown, Player killer = null, string cassieAnnouncement = "") => Kill(new ExiledDamageHandler(killer, float.MaxValue, cassieAnnouncement, DamageHandler.TranslationConversion.FirstOrDefault(k => k.Value == damageType).Key.LogLabel));
```
Michal78900, line 1524: Change `float.MaxValue` to `-1f`/

Exiled.API/Features/Player.cs
```diff
@@ -1500,16 +1500,37 @@ public void Heal(float amount, bool overrideMaxHealth = false)
...
+
+        /// <summary>
+        /// Kills the player.
+        /// </summary>
+        /// <param name="deathReason">The reason the player has been killed.</param>
+        /// <param name="killer">The <see cref="Player"/> who killed player.</param>
+        /// <param name="cassieAnnouncement">The cassie announcement to make upon death.</param>
+        public void Kill(string deathReason, Player killer = null, string cassieAnnouncement = "") => Kill(new ExiledDamageHandler(killer, float.MaxValue, cassieAnnouncement, deathReason));
```
Michal78900, line 1532: Same as above + the `deathReason` does not work.

There are 2 solutions:
1) Fixing ExiledDamageHandler
2) Using `CustomReasonDamageHandler`. The problem with that is second constructor, which will always cause NRE (NW issue, they already know(

Michal78900 (changes requested):
View the comments.

Remuchi:
For some reasons you can't create ragdolls with ExiledDamageHandler even if it works the same as all other base-game DamageHandlers (my guess is that ragdolls can only be created with base game DamageHandlers as ragdolls are client-sided). so as a temp fix it should be changed into CustomReasonDamageHandler when killing player

PR #1058, opened 2022-01-20
Class Roles

Thundermaker300:
Classes for roles. An idea I came up with to cleanup some of the clutter in the Player class. This  allows us to give more data to the player regarding a class in a more organized way.  
  
How it works: `Player.Role` has been converted from a `RoleType` to a `Role` class. The `Role` class is an abstract class that is different based on the player's current role. It can be casted into a `HumanRole`, `SpectatorRole`, and `ScpXXXRole`. Each role class is specific to the player, which means that changes to the properties and calls to the methods of the Role will only apply to the player it is accessed by. Each Role contains API that allows for control over the class specifically (eg. 049 reviving, 079 cameras, etc). As an added bonus, the role is cached, and only creates a new object when it is accessed and when the player's role changes.  
  
As an example, `Player.SetCamera(...)` turns into `(Player.Role as Scp079Role).SetCamera(...)`. It seems more confusing at first, however it simplifies the Player class significantly, and gives an organized place to keep SCP-079 related API.  
  
An example of a simplification: `public bool IsHuman => Team == Team.MTF || Team == Team.CDP || Team == Team.CHI || Team == Team.RSC;` is now `public bool IsHuman => Role is HumanRole`.  
  
With proper documentation & tutorials (which I do not mind writing as well), this could be a useful way to organize class-related API (specifically playable SCP classes). This PR is not done yet (not all planned classes have been included), but I'm looking for feedback and thoughts before I continue


NaoUnderscore:
This looks pretty cool ngl, but we'd have to cast everything from now on.
Also, breaking changes are allowed, you don't have to use ObsoleteAttribute.
If it gets approved, please change `Role::RoleType` to `Role::Type`.

Thundermaker300:
One of my biggest concerns was how much casting would be required. On the flip side, I am only moving a few properties/methods from the Player class to the role classes, specifically SCP related ones (in fact, I think I've already moved all of the ones that I plan on moving, although I'd have to double check that).  
  
You gotta admit, it does provide a better way to check if a player is a human, as opposed to the if statement lol

Thundermaker300:
Another idea I had was to rename them to "managers" or "controllers", although that's up for debate too

Ceglaa:
controllers sounds right

NaoUnderscore:
We'd need to cast a lot, therefore you could make some getters for Player class, which implement all the new roles, in a way that they get automatically converted, and if the player has not the authority to use that role because it's currently set as another role, it'll return null, so that people can run their null checks and then directly interact with Player class without casting everything to set certain values.
Another option, which could be added to the first one I've just explained, would be to use TryGet instance methods, which would return an instance of the requested role given a specified data (which could be RoleType in this case), then use `isinst` to check the type, and finally get the instance to interact with.
The second option is pretty almost the same as performing an if check using `isinst`, that's why I didn't include it as must-do thing in the first option.

NaoUnderscore:
> Another idea I had was to rename them to "managers" or "controllers", although that's up for debate too

I'd prefer avoiding to rename them to "controllers" or "managers", we could just wrap everything in Player class as described in the previous comment. Don't fall in the Unity Dev mindset trap 🙃

Thundermaker300:
> We'd need to cast a lot, therefore you could make some getters for Player class, which implement all the new roles, in a way that they get automatically converted, and if the player has not the authority to use that role because it's currently set as another role, it'll return null, so that people can run their null checks and then directly interact with Player class without casting everything to set certain values.

So let me make sure I got this correct. Do you mean something along the lines of this?
```cs
public Scp079Role ExampleProperty
{
    get
    {
        if (player.RoleType == RoleType.Scp079)
            return [class]
        else
            return null;
    }
}
```
For every role?

NaoUnderscore:
> > We'd need to cast a lot, therefore you could make some getters for Player class, which implement all the new roles, in a way that they get automatically converted, and if the player has not the authority to use that role because it's currently set as another role, it'll return null, so that people can run their null checks and then directly interact with Player class without casting everything to set certain values.
> 
> So let me make sure I got this correct. Do you mean something along the lines of this?
> 
> ```cs
> public Scp079Role ExampleProperty
> {
>     get
>     {
>         if (player.RoleType == RoleType.Scp079)
>             return [class]
>         else
>             return null;
>     }
> }
> ```
> 
> For every role?

Yeah, it's not needed actually, but it helps people to avoid filling in all the places inside their code where casting is needed.

IE:
```cs
(Player::Role as Scp079Role)::get_Property
(Player::Role as Scp079Role)::_field
(Player::Role as Scp079Role)::callMethod()
```
would just become:
```cs
Player::Scp079Wrapper::get_Property
Player::Scp079Wrapper::_field
Player::Scp079Wrapper::MethodCall()
```
And I'm pretty sure is more readable and fast to write, isn't it?

Thundermaker300:
Would it be more appropriate to do something along the lines of
```cs
public Scp079Wrapper Scp079Wrapper { get; } = new Scp079Wrapper(this);
```
This way it's always defined, it wouldn't allow you to check the role but it would be an easier way to manage the properties and methods of the player.

NaoUnderscore:
> Would it be more appropriate to do something along the lines of
> 
> ```cs
> public Scp079Wrapper Scp079Wrapper { get; } = new Scp079Wrapper(this);
> ```
> 
> This way it's always defined, it wouldn't allow you to check the role but it would be an easier way to manage the properties and methods of the player.

This would solve the null check problem 👍

louis1706:
I don't get why you have created the Tutorial Role class because it's technically an HumanRole i thinks

Exiled.API/Features/Roles/Scp173Role.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            get => script._tantrumCooldownRemaining;
+            set => script._tantrumCooldownRemaining = value;
+        }
+
+        /// <summary>
+        /// Force places a tantrum.
+        /// </summary>
+        /// <param name="failIfObserved">Whether or not to place the tantrum if SCP-173 is currently being viewed.</param>
+        public void Tantrum(bool failIfObserved = false)
```
NaoUnderscore, line 108: Use same docs and naming convs used by `Player::PlaceTrantrum`.

Exiled.API/Features/Roles/Scp106Role.cs
```diff
@@ -0,0 +1,82 @@
...
+        public float CaptureCooldown
+        {
+            get => script.captureCooldown;
+            set => script.captureCooldown = value;
+        }
+
+        /// <summary>
+        /// Forces SCP-106 to use its portal, if one is placed.
+        /// </summary>
+        public void Teleport() => script.UserCode_CmdUsePortal();
```
NaoUnderscore, line 74: Rename it to UsePortal

Exiled.API/Features/Roles/Scp049Role.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// Returns a boolean describing whether or not SCP-049 is close enough to a ragdoll to revive it.
+        /// <para>
+        /// This method only returns whether or not SCP-049 is close enough to the body to revive it; the body may have expired. Make sure to check <see cref="Ragdoll.AllowRecall"/> to ensure the body can be revived.
+        /// </para>
+        /// </summary>
+        /// <param name="ragdoll">The ragdoll to check.</param>
+        /// <returns><see langword="true"/> if close enough to revive the body; otherwise, <see langword="false"/>.</returns>
+        public bool InRecallingDistance(Ragdoll ragdoll) => Vector3.Distance(Owner.ReferenceHub.transform.position, ragdoll.Position) <= PlayableScps.Scp049.ReviveDistance * 1.3f;
```
NaoUnderscore, line 62: I'd rename it to `CanRecall` or `InRecallRange`.

Exiled.API/Features/Roles/Scp049Role.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                if (!IsRecalling || script._recallHubServer == null)
+                    return null;
+
+                return Player.Get(script._recallHubServer);
+            }
+        }
+
+        /// <summary>
+        /// Returns a boolean describing whether or not SCP-049 is close enough to a ragdoll to revive it.
```
NaoUnderscore, line 55: Docs should begin with _Gets a value indicating whether..._

NaoUnderscore (changes requested):
TutorialRole is pretty useless imo. You could just wrap it into the HumanRole if you really want to make it as a standalone role.

Exiled.API/Features/Roles/HumanRole.cs
```diff
@@ -28,6 +28,6 @@ internal HumanRole(Player player, RoleType type)
         public override Player Owner { get; }
 
         /// <inheritdoc/>
-        public override RoleType Type { get; }
+        internal override RoleType RoleType { get; }
```
NaoUnderscore, line 31: Make it field instead.

NaoUnderscore (changes requested):

Thundermaker300:
I created TutorialRole because IsHuman previously returned `false` for tutorials. It can be included in the human role though

Exiled.API/Features/Roles/HumanRole.cs
Thundermaker300: Fields cannot be abstract

NaoUnderscore: Bruh, how do you think properties work in C#?
Anyway, don't make it abstract, where's the problem?

Exiled.API/Features/Roles/Role.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <returns><see langword="true"/> if the values are not equal.</returns>
+        public static bool operator !=(RoleType type, Role role)
+            => role.Type != type;
+
+        /// <summary>
+        /// Casts the role to the specified role type.
+        /// </summary>
+        /// <typeparam name="T">The type of the class.</typeparam>
+        /// <returns>The casted class, if possible.</returns>
+        public T As<T>()
```
NaoUnderscore, line 103: Why don't you make a safe `As<T>` version? Basically it would be `bool As<T>(out T param)`, which is better than doing `Role::As<T> != null`.

NaoUnderscore:
> I created TutorialRole because IsHuman previously returned `false` for tutorials. It can be included in the human role though

Gotcha, but I think it'll lead to many issues. Could you include it in `HumanRole`?

Thundermaker300:
yeah, any classes that don't have a role default to HumanRole

Exiled.API/Features/Roles/Role.cs
Thundermaker300: sounds hot

joker-119:
Build errors go brr

NaoUnderscore (approved):

sanyae2439 (approved):

PR #1067, opened 2022-01-23
Convert ticket properties from get to get and set

Thundermaker300:
somebody please stop me from making commits


louis1706:
You can use uint and cast it to an int

NaoUnderscore (approved):

PR #1017, opened 2022-01-11
Tutorials Part1

Thundermaker300:
Tutorials for the site

EDIT: Also contains README changes now lol

PR #1068, opened 2022-01-24
Doc enhance

Thundermaker300:
Unlike my other 700 doc prs, this pr focuses specifically on enhancing the currently existing documentation by providing examples, and using lesser-used tags like seealso and paramref. These tags will render on the [exiled site docs](https://exiled-team.github.io/EXILED/api/Exiled.API.Enums.html), which will make the documentation more comprehensive.

This PR goes hand-in-hand with #1017.  
ⁱ ˢʰᵒᵘˡᵈ ᵍᵉᵗ ᵃ ᵈᵒᶜᵘᵐᵉⁿᵗᵃᵗⁱᵒⁿ ᵐᵃⁿ ʳᵒˡᵉ ᵒⁿ ᵗʰᵉ ᵈⁱˢᶜᵒʳᵈ ˡᵒˡ

### DIFF
--- a/Exiled.API/Extensions/ItemExtensions.cs
+++ b/Exiled.API/Extensions/ItemExtensions.cs
@@ -28,6 +28,12 @@ namespace Exiled.API.Extensions
         /// </summary>
         /// <param name="item">The item to be checked.</param>
         /// <returns>Returns whether the <see cref="ItemType"/> is an ammo or not.</returns>
+        /// <example>
+        /// <code language="cs">
+        /// Log.Info(ItemType.Ammo9x19.IsAmmo()); // true
+        /// Log.Info(ItemType.MicroHID.IsAmmo()); // false
+        /// </code>
+        /// </example>
         public static bool IsAmmo(this ItemType item) => item == ItemType.Ammo9x19 || item == ItemType.Ammo12gauge || item == ItemType.Ammo44cal || item == ItemType.Ammo556x45 || item == ItemType.Ammo762x39;
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


NaoUnderscore:
I gotta be honest.
Some fixes are fine, but making an example for a bunch of methods is pretty useless in my opinion.
You should make an example doc for each method/property, but it's still useless. People need to figure out how to what we give them. This PR aims to help people who don't fully understand how C# and Exiled work, but it's not a "real help". As I said before, if people don't know how to use something, they should probably learn C# first.

Thundermaker300:
> I gotta be honest. Some fixes are fine, but making an example for a bunch of methods is pretty useless in my opinion. You should make an example doc for each method/property, but it's still useless. People need to figure out how to what we give them. This PR aims to help people who don't fully understand how C# and Exiled work, but it's not a "real help". As I said before, if people don't know how to use something, they should probably learn C# first.

I definitely understand where you're coming from when you say that. I wouldn't want the Exiled website to be used by people who don't know C# at all; that's not the goal whatsoever. The goal is to further demonstrate examples and usages of the API we provide, so that readers can apply them for their own use cases. The examples I have provided thus far do not go into the C# details, only how certain properties/methods work/what they return/etc. I don't plan on giving every property/method an example, only the ones that could benefit from it.

People who are new to C# and using the website for whatever reason may pick up a few new details (eg. how to use a foreach loop, or define a variable, or whatever it may be), but that isn't the goal with the examples I've provided. Making that claim is like saying that #1017 is useless because it's too informative for people who may be new to C# and people should figure it out themselves.

I don't want new users who don't know C# to be making plugins as much as you don't. However, we shouldn't lack documentation resources for those who *do* know C# but don't understand our framework as well.

This PR also does include other changes like usages of seealso and paramref tags (both of which show on the website). I could include an edit to the index.md (on #1017) that states that the site should not be used until the reader is familiar with C#.

NaoUnderscore:
I understand you want you're trying to make me understand, and about the `seealso` and other improvements I'm ok with them.
But I'd not add examples to "some methods" or "some properties", because it's not a common usage.
Also, we've the official exiled docs generated by docfx, but it can still be edited, and I'd like you to improve that instead.

Let's make a quick example:
**HarmonyLib (Harmony)** - It's pretty difficult for beginners, right? Even if docs are made pretty well, people don't understand sometimes, and that's where Harmony docs comes in.
In fact, Harmony uses simple documentation like ours, but it uses another tool to achieve what you're trying to do.

Said that, I'm not telling you that I'm against to examples or tutorials, I think that everyone should be able to make plugins, even people with lacks on C# knowledge. **But, what you're trying to do, is not meant to be done with the tool you're using.** Summary docs help people to understand how something works, but it's up to them understanding how to use it correctly. Harmony, as I already said before, used the right tool for the right task, which is not "writing examples in summary". I could even help you to write docs, as well as other people I guess, but we need to use the right tool, and I can assure you that Summary is not the right one.

Jesus-QC:
Fetch dev to fix build errors

PR #1079, opened 2022-02-06
Handler and Patch parts where UsedItem was ItemUsed

RyogenX:
Title say everything

### DIFF
--- a/Exiled.Example/Example.cs
+++ b/Exiled.Example/Example.cs
@@ -92,6 +92,7 @@ namespace Exiled.Example
             Exiled.Events.Handlers.Player.Shooting += playerHandler.OnShooting;
             Exiled.Events.Handlers.Player.ReloadingWeapon += playerHandler.OnReloading;
             Exiled.Events.Handlers.Player.ReceivingEffect += playerHandler.OnReceivingEffect;
+            Exiled.Events.Handlers.Player.TriggeringTesla += playerHandler.OnTriggeringTesla;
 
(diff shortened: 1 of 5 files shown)


Jesus-QC:
why

RyogenX:
> why

Just wanted to fix that little error :(

Exiled.Example/Example.cs
```diff
@@ -92,6 +92,7 @@ private void RegisterEvents()
             Exiled.Events.Handlers.Player.Shooting += playerHandler.OnShooting;
             Exiled.Events.Handlers.Player.ReloadingWeapon += playerHandler.OnReloading;
             Exiled.Events.Handlers.Player.ReceivingEffect += playerHandler.OnReceivingEffect;
+            Exiled.Events.Handlers.Player.TriggeringTesla += playerHandler.OnTriggeringTesla;
```
joker-119, line 95: You never unregister this event.

Also I don't see the point of any of the changes in Exiled.Example, I feel there's already more than enough code examples for people to get an idea of what to do.

joker-119 (changes requested):

Exiled.Example/Example.cs
RyogenX: ups sorry, never mean for this to got to the pr, just the correction of the event, experimenting with pr rn

RyogenX: I don't actually know how to do PR efficienly, If I need to delete the fork and make another one with just the UsedItem correction I can do it.

PR #1081, opened 2022-02-07
Smol Changes <:

SplatTab:
<:


NaoUnderscore (changes requested):
Firstly, you're adding some handlers randomly, without even un/registering them.
Secondly, this clearly means that you didn't test your pull request, and I hope for you that's the last time you PR non-tested changes.
We're not here to give away free purple roles, that's not why Exiled is open source, and that's not why we give those roles. Contributing means keeps doing it, and using brain to do it as well.
 It's like giving a free driving license without knowning how to drive. It's useless, because you will never use it, and if you'll ever use it, you're gonna have a bad time.
About the first two changes, it's fine.
If you want to add all the existing events into Exiled.Example, well, devtools is a thing.

PR #1082, opened 2022-02-07
Introducing Log::Assert

Jesus-QC:
**Log.Assert** is a useful method for **debugging** which throws an exception when the value isnt the expected, stopping the execution of the method.

```cs
// only works if the build of the plugin is a DEBUG build. Otherwise the method will be ignored.

Player ply = Player.Get(3);
Log.Assert(ply != null, "The player 3 is null. THAT SHOULDNT HAPPEN!!"); 

SomeMethod(ply);

```

If the value isn't the expected (Not Null) it will log an error and stop the execution, therefore `SomeMethod` wont be called.

### DIFF
--- a/Exiled.API/Features/Log.cs
+++ b/Exiled.API/Features/Log.cs
@@ -148,7 +148,6 @@ namespace Exiled.API.Features
         /// <param name="condition">The conditional expression to evaluate. If the condition is true it will continue.</param>
         /// <param name="message">The information message. The error and exception will show this message.</param>
         /// <exception cref="System.Exception">If the condition is false. It throws an exception stopping the execution.</exception>
-        [System.Diagnostics.Conditional("DEBUG")]
         public static void Assert(bool condition, object message)
         {
             if (condition)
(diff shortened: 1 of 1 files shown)


louis1706:
You can just add a condition on the error like the Log.Debug(string,bool = true) no ?

Jesus-QC:
Yeah but that doesn't stop the execution.

Jesus-QC:
Instad of using Debug.Assert you can also use 5 lines to have the same but this is to help with debugging

Jesus-QC:
Also this is method is not compiled if it isn't a debug build.

NaoUnderscore (approved):

joker-119:
Won't *exiled* have to be compiled as debug for this method to be compiled?
Therefore in our releases any plugin trying to use the method will hit missing method exceptions?
:hmm:

NaoUnderscore:
> Won't _exiled_ have to be compiled as debug for this method to be compiled? Therefore in our releases any plugin trying to use the method will hit missing method exceptions? :hmm:

Are you drunk?
immagine

NaoUnderscore:
Oh wait, I am drunk.

NaoUnderscore (changes requested):
Remove conditional thing.

Jesus-QC:
However I dont care, if people wants to use this on RELEASE builds let them.

PR #1100, opened 2022-02-17
Fix map spawning item event not calling

rayzerbrain:
This calls it before the null check so the ev.Pickup might be null, but I never saw it happen while testing


rayzerbrain:
Calling the event before anything else makes it easier to maintain if the original method ever changes, but if null checks for subscribing methods ever become necessary it'd introduce some redundancy because the nw method already has a null check. But as I already said, it seems they're never null anyway

joker-119:
No. There's a much, much simpler way to fix this that doesn't require redundant pickup == null checks

PR #1101, opened 2022-02-17
TeslaGate API and Event changed

louis1706:
- Make easyer for patch it
- Repair the actual idle bug
- Add some method in TeslaGate API


louis1706:
He have a actual bug please don't merged it

Exiled.Events/EventArgs/TriggeringTeslaEventArgs.cs
```diff
@@ -21,14 +21,12 @@ public class TriggeringTeslaEventArgs : EventArgs
         /// </summary>
         /// <param name="player"><inheritdoc cref="Player"/></param>
         /// <param name="teslaGate"><inheritdoc cref="Tesla"/></param>
-        /// <param name="isInHurtingRange"><inheritdoc cref="IsInHurtingRange"/></param>
-        /// <param name="isTriggerable"><inheritdoc cref="IsTriggerable"/></param>
-        public TriggeringTeslaEventArgs(Player player, TeslaGate teslaGate, bool isInHurtingRange, bool isTriggerable)
```
joker-119, line 26: Why would we remove these parameters, just to re-run the same methods for them? It's inefficienct, especially considering each of these methods preforms a Math.sqrt operation.

louis1706: it's not for re-run i have thinking of only run it in the event and don't run this method in the normal method

joker-119: Oh I misread the patch part, my bad

louis1706: yes it's the reason why i have made the prefix just for explain how it's could be ^^'

louis1706:
Nao wante change:
- re-added the Method ``CanBeTriggered``
- Trigger method not going to check if it's already in progress
Method name
- Changing for original name of the method

louis1706:
In Favor of Nao PR

PR #1114, opened 2022-02-21
Repair ``TryAddCandy``

louis1706:


joker-119:
Idk why you re-wrote the entire thing when you can just add ``.ToList()`` to the scp330.Candies foreach

PR #1116, opened 2022-02-22
Exiled.CreditTags config

zhaguanyang:


zhaguanyang:
> What is the point of this change?

The description said by default it should be false. So I am making it stay consistent with the docs

zhaguanyang:
image
It just doesn't make sense when the docs said it would be ``false``, while it's actually ``true``

sanyae2439 (approved):

PR #1119, opened 2022-02-23
Added new EscapingEventArgs ctr

Klarulor:
I think its should be for custom escaping in some plugins


louis1706:
Plugin can just create their constructor if he need it idk it's really good to implement it

joker-119:
1. PRs aren't to be made against master
2. Please actually test that your code compiles in the future before making the PR.

PR #1121, opened 2022-02-24
Register of CustomAbility and Reflection 

IRacle1:
Fixed of CustomAbility Register and BaseType check


BuildBoy12:
Please correct me if I am wrong, but isn't this effectively the same code?

joker-119:
This is exactly the same code.

IRacle1:
This isn't same lol.
For example:
```csharp
public class MyCustomAbility : ActiveAbility
{

}
```

BaseType for this class will be `ActiveAbility` and `typeof(MyCustomAbility).BaseType == typeof(CustomAbility)` returns `false`
But `typeof(MyCustomAbility).IsSubclassOf(typeof(CustomAbility)` returns `true`

PR #1122, opened 2022-02-24
Repair ``TeslaGate`` in ``Room``

louis1706:
-Remove the useless CallDelayed
-Remove useless Error
-Repair the TeslaGate in Room

### DIFF
--- a/Exiled.API/Features/Room.cs
+++ b/Exiled.API/Features/Room.cs
@@ -8,6 +8,7 @@
 namespace Exiled.API.Features
 {
     using System;
+#pragma warning disable 1584
     using System.Collections.Generic;
     using System.Linq;
 
--- a/Exiled.API/Features/Room.cs
+++ b/Exiled.API/Features/Room.cs
@@ -8,7 +8,6 @@
 namespace Exiled.API.Features
 {
     using System;
-#pragma warning disable 1584
     using System.Collections.Generic;
     using System.Linq;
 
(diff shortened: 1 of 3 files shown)


joker-119:
1. Don't remove my debug logs.
2. The delay is necessary because lockers don't generate immediately, which is why the delay was added in the first place.

Exiled.API/Features/Room.cs
```diff
@@ -8,7 +8,6 @@
 namespace Exiled.API.Features
 {
     using System;
-#pragma warning disable 1584
```
joker-119, line 11: Don't remove my pragma disables :anger:

louis1706: Oops

BuildBoy12 (approved):

Exiled.Events/Handlers/Internal/MapGenerated.cs
```diff
@@ -50,23 +50,20 @@ internal static class MapGenerated
...
-                GenerateCache();
-                Door.RegisterDoorTypesOnLevelLoad();
-            });
+            GenerateCache();
+            Door.RegisterDoorTypesOnLevelLoad();
         }
 
         private static void GenerateCache()
         {
+            GenerateTeslaGates();
```
joker-119, line 59: Wont this being before GenerateRooms cause TeslaGate.Room to be null?

louis1706: Yes it's that

PR #1124, opened 2022-02-24
Repair ScpRole Script

louis1706:
- the Script are actualise when it's null


louis1706:
It's have tested it on my server and i have not see any error (from 096 and 079)

PR #1126, opened 2022-02-24
Fix scripts being null

joker-119:
In oppoisition to #1124

### DIFF
--- a/Exiled.API/Features/Roles/Scp079Role.cs
+++ b/Exiled.API/Features/Roles/Scp079Role.cs
@@ -14,6 +14,8 @@ namespace Exiled.API.Features.Roles
     /// </summary>
     public class Scp079Role : Role
     {
+        private Scp079PlayerScript script;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Scp079Role"/> class.
         /// </summary>
@@ -22,6 +24,11 @@ namespace Exiled.API.Features.Roles
 
         /// <inheritdoc/>
         public override Player Owner { get; }
+
+        /// <summary>
+        /// Gets the <see cref="Scp079PlayerScript"/> script for the role.
+        /// </summary>
+        public Scp079PlayerScript Script => script ? script : script = Owner.ReferenceHub.scp079PlayerScript;
 
         /// <summary>
         /// Gets or sets the camera SCP-079 is currently controlling.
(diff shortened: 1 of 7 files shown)


Exiled.API/Features/Roles/Scp079Role.cs
```diff
@@ -23,6 +25,11 @@ public class Scp079Role : Role
         /// <inheritdoc/>
         public override Player Owner { get; }
 
+        /// <summary>
+        /// Gets the <see cref="Scp079PlayerScript"/> script for the role.
+        /// </summary>
+        public Scp079PlayerScript Script => script ? script : script = Owner.ReferenceHub.scp079PlayerScript;
```
NaoUnderscore, line 31: bruh?

Exiled.API/Features/Roles/Scp106Role.cs
```diff
@@ -20,29 +20,30 @@ public class Scp106Role : Role
...
-        }
+        internal Scp106Role(Player player) => Owner = player;
 
         /// <inheritdoc/>
         public override Player Owner { get; }
 
+        /// <summary>
+        /// Gets the <see cref="Scp106PlayerScript"/> script for the role.
+        /// </summary>
+        public Scp106PlayerScript Script => script ? script : script = Owner.ReferenceHub.scp106PlayerScript;
```
NaoUnderscore, line 31: bruh?

NaoUnderscore (changes requested):

Exiled.API/Features/Roles/Scp079Role.cs
joker-119: rider said to do this for these 2 scripts instead of the script ?? (script = ) thing

Exiled.API/Features/Roles/Scp106Role.cs
joker-119: see above

Exiled.API/Features/Roles/Scp079Role.cs
NaoUnderscore: Yeah ok it doesn't make sense

louis1706:
Can you make this same for 939 and 0492 like i have do in
#1124

(i just add this for remind)

PR #1127, opened 2022-02-25
Intercom class

zhaguanyang:
Added ``Intercom`` class.
Added ``Intercom.SpeechRemainingTime``.

Moved all Intercom stuff from the ``Map`` class to the new ``Intercom`` class. **_(Breaking changes)_**

### DIFF
--- a/Exiled.API/Features/Intercom.cs
+++ b/Exiled.API/Features/Intercom.cs
@@ -36,7 +36,7 @@ namespace Exiled.API.Features
         public static bool InUse => State == global::Intercom.State.Transmitting || State == global::Intercom.State.TransmittingBypass || State == global::Intercom.State.AdminSpeaking;
 
         /// <summary>
-        /// Gets the <see cref="Player"/> that is using the intercom. Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.
+        /// Gets the <see cref="Player"/> that is using the intercom.<br>Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.</br>
         /// </summary>
         public static Player Speaker => !InUse ? null : Player.Get(global::Intercom.host.speaker);
 
--- a/Exiled.API/Features/Intercom.cs
+++ b/Exiled.API/Features/Intercom.cs
@@ -33,7 +33,7 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets a value indicating whether or not the intercom is currently being used.
         /// </summary>
-        public static bool InUse => State == global::Intercom.State.Transmitting || State == global::Intercom.State.TransmittingBypass || State == global::Intercom.State.AdminSpeaking;
+        public static bool InUse => global::Intercom.host._inUse;
 
         /// <summary>
         /// Gets the <see cref="Player"/> that is using the intercom. Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.
--- a/Exiled.API/Features/Intercom.cs
+++ b/Exiled.API/Features/Intercom.cs
@@ -38,7 +38,7 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets the <see cref="Player"/> that is using the intercom. Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.
         /// </summary>
-        public static Player Speaker => Player.Get(global::Intercom.host.speaker);
+        public static Player Speaker => !InUse ? null : Player.Get(global::Intercom.host.speaker);
 
         /// <summary>
         /// Gets or sets the remaining speech time of the intercom.
--- a/Exiled.API/Features/Intercom.cs
+++ b/Exiled.API/Features/Intercom.cs
@@ -22,14 +22,18 @@ namespace Exiled.API.Features
         }
 
         /// <summary>
-        /// Gets the current state of the intercom.
+        /// Gets or sets the current state of the intercom.
         /// </summary>
-        public static global::Intercom.State State => global::Intercom.host.IntercomState;
+        public static global::Intercom.State State
+        {
+            get => global::Intercom.host.IntercomState;
+            set => global::Intercom.host.IntercomState = value;
+        }
 
         /// <summary>
         /// Gets a value indicating whether or not the intercom is currently being used.
         /// </summary>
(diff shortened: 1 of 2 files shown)


rayzerbrain:
Could you rename ```Intercom.Content``` to something like ```Intercom.DisplayText```? I think it'd be easier to understand that way

zhaguanyang:
> Could you rename `Intercom.Content` to something like `Intercom.DisplayText`? I think it'd be easier to understand that way

Sure, sounds good

Michal78900:
Fix build errors please

zhaguanyang:
F, I renamed IntercomState but forgot stuff using it

Exiled.API/Features/Intercom.cs
```diff
@@ -0,0 +1,59 @@
...
+        public static global::Intercom.State State
+        {
+            get => global::Intercom.host.IntercomState;
+            set => global::Intercom.host.IntercomState = value;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether or not the intercom is currently being used.
+        /// </summary>
+        public static bool InUse => State == global::Intercom.State.Transmitting || State == global::Intercom.State.TransmittingBypass || State == global::Intercom.State.AdminSpeaking;
```
Jesus-QC, line 36: Couldnt you just check if its in cooldown or waiting

Exiled.API/Features/Intercom.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// Gets the <see cref="Player"/> that is using the intercom. Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.
+        /// </summary>
+        public static Player Speaker => !InUse ? null : Player.Get(global::Intercom.host.speaker);
+
+        /// <summary>
+        /// Gets or sets the remaining speech time of the intercom.
+        /// </summary>
+        public static float SpeechRemainingTime
+        {
+            get => !InUse ? 0f : global::Intercom.host.speechRemainingTime;
```
Jesus-QC, line 48: Return the real value

Exiled.API/Features/Intercom.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether or not the intercom is currently being used.
+        /// </summary>
+        public static bool InUse => State == global::Intercom.State.Transmitting || State == global::Intercom.State.TransmittingBypass || State == global::Intercom.State.AdminSpeaking;
+
+        /// <summary>
+        /// Gets the <see cref="Player"/> that is using the intercom. Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.
+        /// </summary>
+        public static Player Speaker => !InUse ? null : Player.Get(global::Intercom.host.speaker);
```
Jesus-QC, line 41: You can use Player.Get as it will return null if its null

Jesus-QC (changes requested):

Exiled.API/Features/Intercom.cs
zhaguanyang: > You can use Player.Get as it will return null if its null

Same as above

Jesus-QC: No? Then it has 2 null checks

Exiled.API/Features/Intercom.cs
Jesus-QC: @joker-119 any reasons for that

Exiled.API/Features/Intercom.cs
zhaguanyang: > Couldnt you just check if its in cooldown or waiting

What's the purpose of that? I could check ``Intercom.host.remainingCooldown``, however, that gets changed to ``3f`` when the player just pressed the button.

Jesus-QC: The purpose of it is checking less states

Jesus-QC: Instead of 3, 2

Exiled.API/Features/Intercom.cs
joker-119: Because if it's not in use the remaining time should be 0, since it's not being used. Means we don't need to do checks like
if (Intercom.InUse && Intercom.SpeechRemainingTime > 10f)

We can simply if (Intercom.SpeechRemainingTime > 10f)

Exiled.API/Features/Intercom.cs
joker-119: Using just Player.Get(host.speaker) will have a null check on the speaker hub
Then if it's not null, it still still use a Dictionary.TryGet() which is another null check

If we check InUse first and it's false, we can just return null without the null check and enumeration involved with Player.Get()

I'm also not trusting that NW actually clears host.speaker when someone stops speaking.

Exiled.API/Features/Intercom.cs
```diff
@@ -0,0 +1,59 @@
...
+            get => !InUse ? 0f : global::Intercom.host.speechRemainingTime;
+            set => global::Intercom.host.speechRemainingTime = value;
+        }
+
+        /// <summary>
+        /// Plays the intercom's sound.
+        /// </summary>
+        /// <param name="start">Sets a value indicating whether or not the sound is the intercom's start speaking sound.</param>
+        /// <param name="transmitterId">Sets the transmitterId.</param>
+        public static void PlayIntercomSound(bool start, int transmitterId = 0) => global::Intercom.host.RpcPlaySound(start, transmitterId);
```
NaoUnderscore, line 57: Rename it to `Intercom::PlaySound` since `Intercom::PlayerIntercomSound` is redundant.

Exiled.API/Features/Intercom.cs
```diff
@@ -0,0 +1,59 @@
...
+            set => global::Intercom.host.IntercomState = value;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether or not the intercom is currently being used.
+        /// </summary>
+        public static bool InUse => State == global::Intercom.State.Transmitting || State == global::Intercom.State.TransmittingBypass || State == global::Intercom.State.AdminSpeaking;
+
+        /// <summary>
+        /// Gets the <see cref="Player"/> that is using the intercom. Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.
```
NaoUnderscore, line 39: `<br>Will be <see langword="null"/> if <see cref="InUse"/> is <see langword="false"/>.</br>`

NaoUnderscore (changes requested):

BuildBoy12 (approved):

PR #1129, opened 2022-02-26
Sexiled chance

Jesus-QC:
The old sexiled chance was 14%, this makes it 6.9% or lower


Jesus-QC:
We can even make that 0.69%

NaoUnderscore:
6.9% is fine

NaoUnderscore:
I'm watching Attack on Titan, I can't rn.

PR #1132, opened 2022-03-08
Fixes CustomAbility.Check

KrisPrs:
I met a strange bug in my own customrole's customability. Method Check was always returning false in any registered events because `CustomAbility.Players.Count == 0`:
image
image
image
This is the fix of this problem.
First my idea was to make property `CustomAbility.Players` static, but it would make Players tracking common for all CustomRoles at all. So it will be needed to make separated static field for each CustomAbility type, but i have no idea, is it possible.
The only solution i found out it to enumerate all the players and check their CustomRoles and CustomAbilities.
`TrackedPlayers.Add(player);` is moved, because i think, that player already must be in `CustomRole.TrackedPlayers` at the moment, when CustomAbility.AbilityAdded is called for some logic needs.

### DIFF
--- a/Exiled.CustomRoles/API/Features/CustomAbility.cs
+++ b/Exiled.CustomRoles/API/Features/CustomAbility.cs
@@ -36,6 +36,7 @@ namespace Exiled.CustomRoles.API.Features
         /// <summary>
         /// Gets all players who have this ability.
         /// </summary>
+        [YamlIgnore]
         public IEnumerable<Player> Players
         {
             get
--- a/Exiled.CustomRoles/API/Features/CustomAbility.cs
+++ b/Exiled.CustomRoles/API/Features/CustomAbility.cs
@@ -34,6 +34,18 @@ namespace Exiled.CustomRoles.API.Features
         public static HashSet<CustomAbility> Registered { get; } = new HashSet<CustomAbility>();
 
         /// <summary>
+        /// Gets all players who have this ability.
+        /// </summary>
+        public IEnumerable<Player> Players
+        {
+            get
+            {
+                return Player.List.Where(x =>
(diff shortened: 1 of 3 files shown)


KrisPrs:
(Ignore Exiled.Events excluding from compilation, it was accidentally)

KrisPrs:
Also, please, consider rechecking https://github.com/Exiled-Team/EXILED/pull/1121
I'm an active CustomRoles API user, abilities aren't really registered with current EXILED code.

Exiled.CustomRoles/API/Features/CustomAbility.cs
```diff
@@ -33,6 +33,19 @@ public abstract class CustomAbility
...
 
+        /// <summary>
+        /// Gets all players who have this ability.
+        /// </summary>
+        [YamlIgnore]
+        public IEnumerable<Player> Players
+        {
+            get
+            {
+                return Player.List.Where(x =>
```
NaoUnderscore, line 44: 1. Lamdba exists
2. `Player::Get` exists

NaoUnderscore (changes requested):

joker-119:
The amount of needless enumerations in that linq expression makes me want to vomit.

It would be much more prudent to see why CustomAbility isn't adding them to the hashset properly, rather then resorting to an extremely janky, inefficient linq expression.

KrisPrs:
It happeneds because of custom ability registraton and usage specific.
When you register a custom ability, you need to create an instance and call SubscribeEvents. But in future calls it will execute `Check` method only for that instance, that was used in registration - not for the real ones.

Now CustomAbility.Players are separeted for each instances of the same ability, this is the reason

It could be solved, if we were able to make different Players property static for every CustomAbility child, but not for CustomAbility class

PR #1141, opened 2022-03-22
api

Thundermaker300:
some random api that could probably be useful idk
also fixed some docs

### DIFF
--- a/Exiled.API/Features/Ragdoll.cs
+++ b/Exiled.API/Features/Ragdoll.cs
@@ -7,6 +7,7 @@
 
 namespace Exiled.API.Features
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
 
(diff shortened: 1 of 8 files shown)


Exiled.API/Features/Ragdoll.cs
```diff
@@ -174,6 +173,11 @@ public bool AllowAnimations
         /// </summary>
         public Player Owner => Player.Get(ragdoll.Info.OwnerHub);
 
+        /// <summary>
+        /// Gets the time that the ragdoll was created.
+        /// </summary>
+        public System.DateTime CreationTime => new System.DateTime((long)NetworkInfo.CreationTime);
```
NaoUnderscore, line 179: `using System;`

NaoUnderscore (changes requested):

Exiled.API/Features/Ragdoll.cs
Thundermaker300: so true

PR #1144, opened 2022-03-23
Mute Method

louis1706:


Exiled.API/Features/Player.cs
```diff
@@ -604,15 +604,7 @@ public bool IsBypassModeEnabled
         public bool IsMuted
         {
             get => ReferenceHub.dissonanceUserSetup.AdministrativelyMuted;
-            set
```
joker-119, line 607: Since we are making methods for this, and we still need to call IssuePersistentMute or RevokePersistentMute when this setter is called, we should do
`
set => value ? Mute() : UnMute();
`

joker-119 (approved):

BuildBoy12 (approved):

PintTheDragon (approved):

PR #1146, opened 2022-03-24
Room.Players null check

Thundermaker300:


Exiled.API/Features/Room.cs
```diff
@@ -76,7 +76,7 @@ public class Room : MonoBehaviour
         /// <summary>
         /// Gets a <see cref="IEnumerable{T}"/> of <see cref="Player"/> in the <see cref="Room"/>.
         /// </summary>
-        public IEnumerable<Player> Players => Player.List.Where(player => player.IsAlive && player.CurrentRoom.Transform == Transform);
+        public IEnumerable<Player> Players => Player.List.Where(player => player.IsAlive && player.CurrentRoom != null && player.CurrentRoom.Transform == Transform);
```
joker-119, line 79: ``&& (!(player.CurrentRoom is null) && player.CurrentRoom.Transform == Transform)``

Thundermaker300: hmmmm yes

PintTheDragon (changes requested):
Do we have access to `is not null`? If so, please change it to that, otherwise feel free to mark this as resolved.

Exiled.API/Features/Room.cs
```diff
@@ -76,7 +76,7 @@ public class Room : MonoBehaviour
         /// <summary>
         /// Gets a <see cref="IEnumerable{T}"/> of <see cref="Player"/> in the <see cref="Room"/>.
         /// </summary>
-        public IEnumerable<Player> Players => Player.List.Where(player => player.IsAlive && player.CurrentRoom.Transform == Transform);
+        public IEnumerable<Player> Players => Player.List.Where(player => player.IsAlive && !(player.CurrentRoom is null) && player.CurrentRoom.Transform == Transform);
```
NaoUnderscore, line 79: Pattern matching sucks in C# 7.2 or whatever is it, just `!= null`

NaoUnderscore (changes requested):

Exiled.API/Features/Room.cs
joker-119: or !(is null)

NaoUnderscore: `!= null` is just better in this case, cause it doesn't destroys the pattern matching

joker-119: Well, I'm bumping the project to C# 9 and re-working all of the null checks to use pattern matching `is not null` or `is null` anyways so, it won't really matter, I'd just be changing this either way.

PR #1147, opened 2022-03-25
Player.Get Documentation

zhaguanyang:


PintTheDragon (changes requested):
We should maybe change `id` to `ID`, as it's capitalized everywhere else in the sentence.

zhaguanyang:
I was thinking about having it lowercase might be better for distinguishing it from userid initially.

PR #1149, opened 2022-03-29
C# 9.0 go brr

joker-119:


Michal78900 (commented):
StyleCop tells me that it should be `new ()` not `new()`

NaoUnderscore:
> StyleCop tells me that it should be `new ()` not `new()`

Because you don't know how to set Stylecop.

joker-119:
> StyleCop tells me that it should be `new ()` not `new()`

EXILED.ruleset was updated in this PR already to resolve that.

PR #1160, opened 2022-04-09
Add default value to duration parameter in Map::ShowHint

Ceglaa:


louis1706:
Why adding it only on Map and why adding it ?

PR #1163, opened 2022-04-10
Added Generator::Position

Monaldcry7788:


NaoUnderscore (changes requested):
Add Transform and Rotation too.

Ceglaa:
Maybe add NetworkServer.UnSpawn() and NetworkServer.Spawn() to the position/rotation set method? Same as in other object APIs

NaoUnderscore:
> Maybe add NetworkServer.UnSpawn() and NetworkServer.Spawn() to the position/rotation set method? Same as in other object APIs

What

Michal78900 (changes requested):
Setters won't work at all. I also think that UnSpawning and Spawning won't either, but I'm not sure.

PR #1164, opened 2022-04-10
Custom Items Registering Fix

Ceglaa:
Credits to @graczbeznicku, because he spotted this issue in Exiled CustomRoles API, and i have checked CustomItems API and it hade the same issue..


joker-119:
Fixed in 687613ae0c9e969ee45b3b00d5fd87337551ef7f

PR #1168, opened 2022-04-11
SCP:SL 11.2

louis1706:
- Adding DamageType ParticleDisruptor
- Remove the fix CustomReasonDamageHandler because now it's good
- Item::IsScp get true for SCP1853
- ItemConversion adding ParticleDisruptor
- Fix for not Spam Enable effect (Poisoned)


Exiled.Events/Patches/Fixes/Scp1853Fix.cs
```diff
@@ -0,0 +1,43 @@
...
+    using CustomPlayerEffects;
+
+    using HarmonyLib;
+    #pragma warning disable SA1313
+
+    /// <summary>
+    /// Patch the <see cref="Scp1853.OnUpdate"/>.
+    /// Fix Spamming EnableEffect.
+    /// </summary>
+    [HarmonyPatch(typeof(Scp1853), nameof(Scp1853.OnUpdate))]
```
louis1706, line 21: If someone made it in Transpiler i deleted this Prefix and let the guys with the transpiler post

PR #1182, opened 2022-04-13
Warhead.Controller Fix

ismail0234:
Every round fixes the changed value.


ismail0234:
It's not finished yet, don't merge

ismail0234:
I haven't had a chance to test it, but it shouldn't be a problem.

PR #1185, opened 2022-04-13
Update to mirror base-game round end conditions

BuildBoy12:


Exiled.Events/Patches/Events/Server/RoundEnd.cs
```diff
@@ -91,10 +91,11 @@ private static IEnumerator<float> Process(RoundSummary roundSummary)
                 int num3 = newList.scps_except_zombies + newList.zombies;
                 int num4 = newList.class_ds + RoundSummary.EscapedClassD;
                 int num5 = newList.scientists + RoundSummary.EscapedScientists;
+                RoundSummary.SurvivingSCPs = newList.scps_except_zombies;
```
PintTheDragon, line 94: Maybe move this down/up a bit so that it isn't just sitting between the variable declarations.

PR #1189, opened 2022-04-16
`Features.Prefabs`

Jesus-QC:
This is a large contribution that will add some classes that will contain all prefabs registered into `NetworkClient::Prefabs` therefore spawnable.

- [x] Structures
- [ ] Pickups
- [ ] Grenades
- [ ] Doors
- [ ] Primitives (move the helper class)
- [ ] Other (Player, PlayBackLobby, Tantrum ..)

### DIFF
--- a/Exiled.API/Enums/StructureType.cs
+++ b/Exiled.API/Enums/StructureType.cs
@@ -1,0 +1,88 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="StructureType.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Enums
+{
+    /// <summary>
+    /// All of the spawnable structures.
+    /// </summary>
+    public enum StructureType
+    {
+        /// <summary>
+        /// The Work Station structure.
+        /// </summary>
+        WorkStation,
+
+        /// <summary>
+        /// The Large Gun Locker structure.
+        /// </summary>
+        LargeGunLocker,
+
+        /// <summary>
+        /// The Rifle Rack structure.
+        /// </summary>
+        RifleRack,
+
+        /// <summary>
+        /// The Misc Locker structure.
+        /// </summary>
+        MiscLocker,
+
+        /// <summary>
+        /// The 079 Generator structure.
+        /// <remarks>
+        /// When you spawn this structure, it is automatically added to the generators list.
+        /// </remarks>
+        /// </summary>
+        Generator,
+
+        /// <summary>
+        /// The Regular Medkit structure.
+        /// </summary>
+        RegularMedkit,
+
+        /// <summary>
+        /// The Adrenaline Medkit structure.
+        /// </summary>
+        AdrenalineMedkit,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-018.
+        /// </summary>
+        Scp018Pedestal,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-207.
+        /// </summary>
+        Scp207Pedestal,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-244.
+        /// </summary>
+        Scp244Pedestal,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-268.
+        /// </summary>
+        Scp268Pedestal,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-500.
+        /// </summary>
+        Scp500Pedestal,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-1853.
+        /// </summary>
+        Scp1853Pedestal,
+
+        /// <summary>
+        /// The SCP Pedestal of the SCP-2176.
+        /// </summary>
+        Scp2176Pedestal,
+    }
+}
--- a/Exiled.API/Features/Prefabs/Structures.cs
+++ b/Exiled.API/Features/Prefabs/Structures.cs
@@ -35,72 +35,72 @@ namespace Exiled.API.Features.Prefabs
         /// <summary>
         /// Gets the WorkStation structure.
         /// </summary>
-        public static GameObject WorkStation => workstation ? workstation : workstation = NetworkClient.prefabs[Guid.Parse("ad8a455f-062d-dea4-5b47-ac9217d4c58b")];
+        public static GameObject WorkStation => workstation ??= NetworkClient.prefabs[Guid.Parse("ad8a455f-062d-dea4-5b47-ac9217d4c58b")];
 
         /// <summary>
         /// Gets the LargeGunLocker structure.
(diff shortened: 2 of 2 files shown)


rayzerbrain:
I put a quick reference to all prefabs and guids in the resources channel, did you want to implement the rest of them in a similar fashion? I feel like using a dictionary for some would be easier

Jesus-QC:
@rayzerbrain 

I already had a list in the resources channel wdym?

Tell me how is your dictionaries idea.

rayzerbrain:
I was thinking something like this for a rough draft: (since most of them are enums we could even just use arrays)
```cs
private Guid[] structureGuids;

static Prefabs()
{
    structureGuids = new Guid[Enum.GetNames(typeof(StructureType)).Length];
    structureGuids[(int)StructureType.WorkStation] = Guid.Parse("ad8a455f-062d-dea4-5b47-ac9217d4c58b");
    //same for rest
}

public GameObject GetPrefab(StructureType st) => NetworkClient.prefabs[structureGuids[(int)st]];
    
```
It might not be as memory effective but the static init wouldn't get called until the class is used anyway. We could also fit all of them in one class this way too.

Jesus-QC:
> I was thinking something like this for a rough draft: (since most of them are enums we could even just use arrays)
> 
> ```cs
> private Guid[] structureGuids;
> 
> static Prefabs()
> {
>     structureGuids = new Guid[Enum.GetNames(typeof(StructureType)).Length];
>     structureGuids[(int)StructureType.WorkStation] = Guid.Parse("ad8a455f-062d-dea4-5b47-ac9217d4c58b");
>     //same for rest
> }
> 
> public GameObject GetPrefab(StructureType st) => NetworkClient.prefabs[structureGuids[(int)st]];
>     
> ```
> 
> It might not be as memory effective but the static init wouldn't get called until the class is used anyway. We could also fit all of them in one class this way too.

I don't really like the idea.

Jesus-QC:
I will keep this system for now. If you have any other suggestion, feel free to tell me.

rayzerbrain:
> I don't really like the idea.

What don't you like about it? It doesn't have to be the exact impl. but it seems like a lot of work to have a field and getter for all 83 prefabs.

Jesus-QC:
> > I don't really like the idea.
> 
> What don't you like about it? It doesn't have to be the exact impl. but it seems like a lot of work to have a field and getter for all 83 prefabs.

Well, first of all we arent going to create an enum for each prefab, like the PlayBackLobby etc.

Secondly, breaking changes, if some prefab is removed instead of breaking a plugin because the enum isnt found, we could just make the getter obsolete.

rayzerbrain:
> > > I don't really like the idea.
> > 
> > 
> > What don't you like about it? It doesn't have to be the exact impl. but it seems like a lot of work to have a field and getter for all 83 prefabs.
> 
> Well, first of all we arent going to create an enum for each prefab, like the PlayBackLobby etc.
> 
> Secondly, breaking changes, if some prefab is removed instead of breaking a plugin because the enum isnt found, we could just make the getter obsolete.

I agree it'd be unnecessary to make enums for every prefab, but most already have usable ones (ItemType for pickups, maybe ZoneType for doors, etc.), so some would still have to be getters. We could always use dictionaries if arrays are weird to work with.

Also it's probably inevitable an exception'll be thrown if a prefab is removed entirely: even using an obsolete getter will fail if it tries to use a guid the prefabs dict doesn't have anymore.

xRoier:
I would make the Structure class like the Door class, for example, a mix between rayzer's idea and the actual scheme that EXILED has, I also would make a class wrapper for every structure implementing the components the object has, something like ReferenceHub.

Jesus-QC:
> I would make the Structure class like the Door class, for example, a mix between rayzer's idea and the actual scheme that EXILED has, I also would make a class wrapper for every structure implementing the components the object has, something like ReferenceHub.

Uhmm, yeah but we arent talking about structures but prefabs, that would be another external pull request

Jesus-QC:
@rayzerbrain 
> Also it's probably inevitable an exception'll be thrown if a prefab is removed entirely: even using an obsolete getter will fail if it tries to use a guid the prefabs dict doesn't have anymore.

No because if we have the getter system instead, we can just make the getter obsolete and return null

rayzerbrain:
> @rayzerbrain
> 
> > Also it's probably inevitable an exception'll be thrown if a prefab is removed entirely: even using an obsolete getter will fail if it tries to use a guid the prefabs dict doesn't have anymore.
> 
> No because if we have the getter system instead, we can just make the getter obsolete and return null

That'll still throw a null reference when people try to spawn it, unless people already have null checks.

Even so, the ```GetPrefab``` method could do the same thing and return null for missing enums or prefabs, or we could also have a  ```TryGetPrefab(out GameObject)``` so that null checks aren't redundant.

rayzerbrain:
Yes, the point is to do the same thing in a better way for everyone involved. I don't think your method is as optimal for implementation, and since mine isn't as memory efficient, why don't we do something like this
```cs
private Dictionary<ItemType, GameObject> pickupPrefabs;
private Dictionary<StructureType, GameObject> structurePrefabs;

public GameObject GetPickupPrefab(ItemType type)
{
    if ((pickupPrefabs ??= new()).TryGetValue(type, out GameObject obj))
        return obj;
    
    Guid guid = type switch
    {
        ItemType.Painkillers => Guid.Parse("guid here"),
        // etc.
        _ => throw new ArgumentOutOfRangeException()
    };

   GameObject newObj = NetworkClient.prefabs[guid];
   pickupPrefabs.Add(type, newObj);
   return newObj;
}
public GameObject GetStructurePrefab(StructureType type)
{/* Something similar */}
```

I think the save on lines is worth the small amount of extra memory used, don't you?

Jesus-QC:
> 

This doesnt solve anything, still the enum implementation and not memory effective??

rayzerbrain:
> @rayzerbrain This doesnt solve anything, still the enum implementation.

What do you mean?

Jesus-QC:
> > @rayzerbrain This doesnt solve anything, still the enum implementation.
> 
> What do you mean?

Again, we arent making enums for everything since it is not needed

Jesus-QC:
Also, the extra memory used isnt worth for the amunt of lines, Im implementing this, dont worry about it being very heavy to implement at the beggining

Jesus-QC:
Once I finish it, you will just have to add 3 lines for esch prefab you want to add or edit

rayzerbrain:
> > > @rayzerbrain This doesnt solve anything, still the enum implementation.
> > 
> > 
> > What do you mean?
> 
> Again, we arent making enums for everything since it is not needed

Like I said, we wouldn't necessarily have to, we would only do this for the ones that already have enums we can use for them (which are most of them).

It just seems to me like a lot of clutter for no reason, with also needing a big switch-case for each enum you use and multiple classes.

But in the end it's your choice, I guess

Jesus-QC:
@rayzerbrain 

The switch case is more or less like the dict,
Im going to recode it so its easier to maintain anywais. Will let you know by then.

Jesus-QC:
Closing it for now

PR #1190, opened 2022-04-17
Fix NRE on PlayerHasHint patch

Xname7:
Sometimes NRE is thrown on this patch, so it's the sollution to that.
It happens to me when the player leaves the server while i try to show a hint to him.

### DIFF
--- a/Exiled.Events/Patches/Generic/PlayerHasHint.cs
+++ b/Exiled.Events/Patches/Generic/PlayerHasHint.cs
@@ -30,7 +30,7 @@ namespace Exiled.Events.Patches.Generic
         private static void Postfix(HintDisplay __instance, Hint hint)
         {
             // Try to get the player, if it doesn't exist, just return
-            if (__instance == null || !(Player.Get(__instance.gameObject) is Player player))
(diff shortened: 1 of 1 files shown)


Exiled.Events/Patches/Generic/PlayerHasHint.cs
```diff
@@ -30,7 +30,7 @@ internal static class PlayerHasHint
         private static void Postfix(HintDisplay __instance, Hint hint)
         {
             // Try to get the player, if it doesn't exist, just return
-            if (__instance.gameObject is null || !(Player.Get(__instance.gameObject) is Player player))
+            if (__instance == null || !(Player.Get(__instance.gameObject) is Player player))
```
louis1706, line 33: is null

louis1706 (changes requested):

Exiled.Events/Patches/Generic/PlayerHasHint.cs
joker-119: __instance?.gameObject is null
would be better.

Xname7: okay, will change that quick

PR #1195, opened 2022-04-21
Update CurrentSpectatingPlayers

zhaguanyang:
CurrentSpectatingPlayers shouldn't include the player itself


zhaguanyang:
might also need to check to see the `IsDead` check is necessary or not

PR #1196, opened 2022-04-21
Spectator properties fix

Michal78900:
# Changelog:
- Fixed `Player::CurrentSpectatingPlayers` returning also the player himself
- Removed setter for `SpectatorRole::SpectatedPlayer` because it doesn't syncs with the client


NaoUnderscore:
This is a breaking change and it's not going pass since we can't mark it as obsolete.
The fix for the other property is fine. I'll approve it whenever you're gonna remove the breaking change part.

Jesus-QC:
#1211

Jesus-QC (approved):

PR #1197, opened 2022-04-21
Added `Spawned` event for the player

Michal78900:
This PR adds `Spawned` event, it is called **after** player has fully spawned.
The difference between this event `ChangingRole`, and `Spawning` event is, that it is called last, after the player has been safely spawned (after the player's position has been set correctly).

The reason for introducing this event is my attempt to fix the very known bug. The bug is being unable to reliable set player's position after his spawn. The easy fix was adding a delay, but it didn't always worked.

The reason behind the bug is most likely they way a player is spawned - in short, the game tries to sets player's position, waits 0.2 seconds and then checks if it's valid (for example, the player has floor under them). If not, it tries again and again until the 50th time when the game gives up and kills the player with `"Client has failed to spawn."` error.

My theory is that players with higher ping are more likely to have more "attempts" of spawn. This result previously mentioned delay too short, because the base-game coroutine still hasn't finished.  On the other hand, players with better connection sometimes need to wait additional time to be moved by the plugin, which is pointless.

The solution would be calling an event after all base-game stuff have been finished. This is basically `Spawned` event.

`Spawned` event is meant to be used either if:
- The player's spawn properties, their spawn inventory, etc. don't need to be changed.
- The dev wants to directly do `player.Position = something` without adding an unreliable delays and they can't/don't want to use `Position` property inside `Spawning` event.
- Calling event only if the player has been safely spawn. Both `ChangingRole` and `Spawning` are called before spawn checks so it is possible that the player will die, because he wasn't able to spawn for some reason. `Spawned` event will be called **only** if a player has spawned successfully. It won't be called at all if he dies while trying to spawn.

The event uses base-game action event. This means I don't patch any method - no transpiler, no prefix. The event contains only a player getter, because it is called after everything else is done, but again if dev wants to change something they should use an other event.

### DIFF
--- a/Exiled.Events/EventArgs/SpawnedEventArgs.cs
+++ b/Exiled.Events/EventArgs/SpawnedEventArgs.cs
@@ -1,0 +1,30 @@
+// -----------------------------------------------------------------------
+// <copyright file="SpawnedEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.EventArgs
+{
+    using System;
+
+    using Exiled.API.Features;
+
+    /// <summary>
+    /// Contains all information after spawning a player.
+    /// </summary>
+    public class SpawnedEventArgs : EventArgs
+    {
+        /// <summary>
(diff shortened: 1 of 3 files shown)


BuildBoy12 (approved):

Jesus-QC:
image
Just use OnSpawning, which works perfectly

Jesus-QC (approved):

PR #1199, opened 2022-04-24
GetMaxAmmo(Pickup) 

louis1706:


NaoUnderscore (changes requested):
Firstly, documentation is bad. Ask @Thundermaker300 to add some valid docs.
Secondly, the code is bad and inefficient. Use a for loop and do what the base-game is doing to get all the attachments from `Firearm`/`FirearmPickup`s. Since we do not have an API for base-game pickups aside from the `ItemPickupBase`, we can't use the existing attachments API to interact/manipulate the attachments of `FirearmPickup`s as we're doing for `Item`s.

Thundermaker300:
aaaaaaaaaaaaaaaaa

louis1706:
Needs to be more efficient

louis1706:
Nobody want to make it better xDDD

PR #1200, opened 2022-04-25
Add setter for ItemsToDrop in ItemsToDrop

antt0n:


Exiled.Events/EventArgs/DyingEventArgs.cs
```diff
@@ -31,7 +31,7 @@ public class DyingEventArgs : EventArgs
         public DyingEventArgs(Player target, DamageHandlerBase damageHandler)
         {
             Handler = new CustomDamageHandler(target, damageHandler);
-            ItemsToDrop = new List<Item>(target.Items.ToList());
+            ItemsToDrop = target.ResetInventory(target.Items.ToList());
```
zhaguanyang, line 34: Have you even tried to compile it? ``ResetInventory`` is a void method, it doesn't return anything.

zhaguanyang (changes requested):

NaoUnderscore:
What does this PR mean?

BuildBoy12:
I'm taking this as an actual joke considering it does not compile, let alone correspond with what the title says. Closing.

zhaguanyang:
> What does this PR mean?

Cuz ItemsToDrop has a setter in the event args, while setting it does nothing.

PR #1219, opened 2022-05-03
Cringe foreaches fix

IRacle1:


Exiled.API/Features/Player.cs
```diff
@@ -1816,11 +1816,11 @@ public void AddItem(Item item, int amount, IEnumerable<AttachmentIdentifier> ide
         /// <param name="items">The list of items to be added.</param>
         public void AddItem(IEnumerable<Item> items)
         {
-            IEnumerable<Item> enumerable = items.ToList();
-            if (enumerable.Any())
+            List<Item> enumerable = items.ToList();
```
sanyae2439, line 1819: why still use ToList()? it not better

sanyae2439 (changes requested):

Exiled.API/Features/Player.cs
joker-119: The only real efficient change that could have been made to the original is saving enumerable.Count() as a local variable, and using that instead of the .Any() and .Count() so that it doesn't enumerate twice.

IRacle1: Because i dont wanna use LINQ `.Count()`, and `List` enumerate is a bit efficient.
And i dont understand this `if` why do this check need?

IRacle1: If i remove this `if`, and `list.Count` usage, so `.ToList()` become unnecessary, yes

IRacle1: Sooooooo should I delete it?

PR #1220, opened 2022-05-04
Added RoomExtensions and ElevatorExtensions

Ficus-x:


NaoUnderscore (changes requested):
ElevatorExtensions are useless.
RoomExtensions are fine, use lambda though.
`IsScpRoom` is not efficient, allocating memory to make a list every time it gets called is the worst thing we could even implement.
Documentation is bad, refer to the up to date documentation introduced by me, I'll update documentation for all the project soon anyway.
Rename `IsScpRoom` to something else, because Room is a repetition.

lifyxus:
I updated the extensions

Exiled.API/Extensions/RoomExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> is a Hcz checkpoint or not.</returns>
+        public static bool IsHczCheckpoint(this RoomType room) => room is RoomType.HczChkpA or RoomType.HczChkpB;
+
+        /// <summary>
+        ///  Checks if a <see cref="RoomType">room type</see> contains any SCP.
+        /// </summary>
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> contains any SCP or not.</returns>
+        public static bool ContainsScp(this RoomType room) => ScpRooms.Contains(room);
```
IRacle1, line 61: I think the best solution will be delete this list, and using `or` (imho)

IRacle1: I don't like allocating memory for a this list

zhaguanyang:
build errors go brrrrrr

lifyxus:
I solved [it](https://github.com/Exiled-Team/EXILED/runs/6327108432?check_suite_focus=true)

Exiled.API/Extensions/RoomExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
```
joker-119, line 68: `public static bool IsScp(this RoomType room) => room is RoomType.Lcz173 or Lcz330 or ..`
There is no reason to allocate a list here.

joker-119 (changes requested):

lifyxus:
I did it 😶

IRacle1 (approved):

Exiled.API/Extensions/RoomExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> is a Hcz checkpoint or not.</returns>
+        public static bool IsHczCheckpoint(this RoomType room) => room is RoomType.HczChkpA or RoomType.HczChkpB;
+
+        /// <summary>
+        ///  Checks if a <see cref="RoomType">room type</see> contains any SCP.
+        /// </summary>
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> contains any SCP or not.</returns>
+        public static bool IsScp(this RoomType room)
```
louis1706, line 53: Adding a bool for playable Scp

Exiled.API/Extensions/RoomExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> is a gate or not.</returns>
+        public static bool IsGate(this RoomType room) => room is RoomType.EzGateA or RoomType.EzGateB;
+
+        /// <summary>
+        /// Checks if a <see cref="RoomType">room type</see> is a checkpoint.
+        /// </summary>
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> is a checkpoint or not.</returns>
+        public static bool IsCheckpoint(this RoomType room) => room is RoomType.LczChkpA
```
louis1706, line 30: Adding a bool if player want to take RoomType.HczChkp because it's don't have the checkpoints door and people can be confuse with it

Exiled.API/Extensions/RoomExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// A set of extensions for <see cref="RoomType"/>.
+    /// </summary>
+    public static class RoomExtensions
+    {
+        /// <summary>
+        /// Checks if a <see cref="RoomType">room type</see> is a gate.
+        /// </summary>
+        /// <param name="room">The room to be checked.</param>
+        /// <returns>Returns whether the <see cref="RoomType"/> is a gate or not.</returns>
+        public static bool IsGate(this RoomType room) => room is RoomType.EzGateA or RoomType.EzGateB;
```
louis1706, line 23: I would put an other method like this HasGate for all the room who have a Gate

NaoUnderscore: I don't really agree, in fact, many rooms can and should `true` because they just have a `ContainmentDoor` inside them. For instance, let's take the Gate A and the room after the main Gate A's room, following your logic/idea, we should check the next room whether it has a gate, and we'd expect it to return `true`.
Hence, I prefer `IsGate` over `HasGate` because it's not about rooms, rather it's about `RoomType`, the `RoomType` can be "Gate", but there's no way to define whether the `RoomType` HAS the specified gate, without manually doing some checks, which wouldn't lead to extension(s) method(s) anymore; that's an API-related feature.
Let's just keep it as it as at the moment.

Exiled.API/Extensions/RoomExtensions.cs
NaoUnderscore: I can't understand your sentence, hence I'm out 🙃

Exiled.API/Extensions/RoomExtensions.cs
NaoUnderscore: Would make sense, but make it `[optional]`.

louis1706: > Would make sense, but make it `[optional]`.

My comment It's already optional no ?

Exiled.API/Extensions/RoomExtensions.cs
louis1706: It's because it's checkpoint should give true only on room with the checkpoint door and the HczChkpt don't have this checkpoint door

PR #1221, opened 2022-05-04
`AnnouncingNtfEntrance` to transpiler

IRacle1:


IRacle1:
I tested this on my local server, so i think this will work

Exiled.Events/EventArgs/AnnouncingNtfEntranceEventArgs.cs
```diff
@@ -30,9 +30,9 @@ public AnnouncingNtfEntranceEventArgs(int scpsLeft, string unitName, int unitNum
         }
 
         /// <summary>
-        /// Gets the number of SCPs left.
+        /// Gets or sets the number of SCPs left.
         /// </summary>
-        public int ScpsLeft { get; }
+        public int ScpsLeft { get; set; }
```
zhaguanyang, line 35: Hmmm, why are you adding a setter to ``ScpsLeft``?

NaoUnderscore (changes requested):
This is clearly made using Sharplab or something like that. I consider it a contribution of course, but it doesn't really say that you know IL enough to re-make the same transpiler from scratch without any external tools. Aside from that,
IL can be more efficient; hence do not use external tools like sharplab, because it won't give the most efficient code, unless you don't know how to C# -> IL a specific code block. Just take a look to the instructions and figure out what can be added/removed or changed to another instruction.

IRacle1:
I used only rider il viewer, i'll check this later

NaoUnderscore:
> I used only rider IL viewer, i'll check this later

Yeah whatever, still an external tool.

Exiled.Events/Patches/Events/Map/AnnouncingNtfEntrance.cs
```diff
@@ -25,18 +29,92 @@ namespace Exiled.Events.Patches.Events.Map
...
+                new(OpCodes.Ldloc_S, unitInformation.LocalIndex),
+                new(OpCodes.Ldc_I4_1),
+                new(OpCodes.Ldelem_Ref),
+                new(OpCodes.Call, Method(typeof(int), nameof(int.Parse), new[] { typeof(string) })),
+                new(OpCodes.Ldc_I4_1),
+                new(OpCodes.Newobj, GetDeclaredConstructors(typeof(AnnouncingNtfEntranceEventArgs))[0]),
+                new(OpCodes.Dup),
+                new(OpCodes.Dup),
+                new(OpCodes.Stloc_S, ev.LocalIndex),
+                new(OpCodes.Call, Method(typeof(Map), nameof(Map.OnAnnouncingNtfEntrance))),
```
IRacle1, line 83: I think this is my mistake in logic, I don't like accessing the `unitInformation` here, but I don't have any ideas how to improve it

IRacle1 (commented):
ok

Exiled.Events/Patches/Events/Map/AnnouncingNtfEntrance.cs
```diff
@@ -84,28 +84,28 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
                 new(OpCodes.Call, Method(typeof(string), nameof(string.Format), new[] { typeof(string), typeof(object), typeof(object) })),
+                new(OpCodes.Dup),
                 new(OpCodes.Starg_S, 1),
 
                 // cassieUnitName = this.GetCassieUnitName(unitName);
-                new(OpCodes.Ldarg_0),
-                new(OpCodes.Ldarg_1),
-                new(OpCodes.Call, Method(typeof(NineTailedFoxNamingRule), nameof(NineTailedFoxNamingRule.GetCassieUnitName))),
+                new(OpCodes.Callvirt, Method(typeof(NineTailedFoxNamingRule), nameof(NineTailedFoxNamingRule.GetCassieUnitName))),
                 new(OpCodes.Stloc_S, 0),
```
NaoUnderscore, line 104: Meh.

NaoUnderscore (changes requested):

Exiled.Events/Patches/Events/Map/AnnouncingNtfEntrance.cs
IRacle1: Line 91, `Ldarg_0`
And a little upper i use `Dup` Instead of `Ldarg_1`

IRacle1: Because `Stloc.S, 0` is a `cassieUnitName`, and this variable should change

NaoUnderscore: [`OpCodes.Stloc.0`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.stloc_0?view=net-6.0) exists.

IRacle1: Oh, i'm cringe, fix that later

IRacle1: Well, later means tomorrow :(
Also thank you Nao for pointing out my mistakes, i'll try to write my next transpiler without any IL viewers!

NaoUnderscore: 👍🏼

joker-119 (approved):

PR #1222, opened 2022-05-05
`door.Unlock` better than `door.ChangeLock(DoorLockType.None)`

IRacle1:


NaoUnderscore (commented):
It's literally the same thing. Unlock() points to the same method.

IRacle1:
Yes, i think this a more readable

Jesus-QC:
> It's literally the same thing. Unlock() points to the same method.

Well, the good thing herr is that nw change how doors are managed it wont break

Jesus-QC (approved):

PR #1231, opened 2022-05-12
Feature friendly fire config and more. 

Undid-Iridium:
Contains changes to how csproj file builds files (allows ENV variable to copy paste over to exiled directory to make building/testing faster)

Also includes friendly fire change to allow users to specify what rules they want a player to have based on role, and what role is attacking them. This depends on a custom role string variable to allow users to specify whatever they want to have unique rules to. This is not optimized yet, and not cleaned up yet. I have to resolve the other IndividualFriendlyFire scenarios (Grenades, Explosion effects, etc). However, processing damage does cover majority of all damaging events from one player to another. I will do testing later. 

Reminder, this is not complete. Just putting it here first for people to see and comment during my process. 

Update:
Ok, I added a fix for ExplodingFlashGrenade not working on self-grenade since NW code funny, and Exiled didn't handle it. 

Update 2:
Ok, I have mostly resolved majority of the possible combination, verified suicide still work, scp on scp is still safe, fake 035 to scp works, d class to rifleman works, etc. Let me know if you have any questions


xRoier (commented):
LGTM for now, keep it up!

joker-119:
Your custom build events need to have conditions to handle both Windows and Linux, as their commands are different and as it is now, it makes EXILED only buildable on windows.

Exiled.Loader/Config.cs
```diff
@@ -33,6 +33,12 @@ public sealed class Config : IConfig
         [Description("The working environment type (Development, Testing, Production, Ptb)")]
         public EnvironmentType Environment { get; set; } = EnvironmentType.Production;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether gets or sets the whether to show debug during production.
+        /// </summary>
+        [Description("Whether to allow debugs during production")]
+        public bool DebugProduction { get; set; } = false;
```
joker-119, line 40: What in the fuck is this even for? We literally already have config values for showing debugs, the only debugs that aren't shown in the Release build are the #if DEBUG #endif compiler flags, which values like this cannot change since the code just isn't included if that flag isn't met during compilation.

Exiled.API/Features/Player.cs
```diff
@@ -128,6 +128,21 @@ public Player(GameObject gameObject)
         /// </summary>
         public static Dictionary<int, Player> IdsCache { get; } = new(20);
 
+        /// <summary>
+        /// Gets or sets a <see cref="Dictionary{TKey, TValue}"/> containing cached <see cref="RoleType"/> and their FF multiplier. This is for non-unique roles.
+        /// </summary>
+        public Dictionary<RoleType, int> FriendlyFireRules { get; set; } = new();
```
joker-119, line 134: Rename to `FriendlyFireMultiplier` or `FriendlyFireMultiplierValues`

Exiled.API/Features/Player.cs
```diff
@@ -128,6 +128,21 @@ public Player(GameObject gameObject)
...
+
+        /// <summary>
+        /// Gets or sets a <see cref="Dictionary{TKey, TValue}"/> containing cached <see cref="string"/> and their  <see cref="Dictionary{TKey, TValue}"/> which is cached Role with FF multiplier. This is for unique custom roles.
+        /// </summary>
+        public Dictionary<string, Dictionary<RoleType, int>> UniqueFriendlyFireRules { get; set; } = new();
```
joker-119, line 139: CustomRoleFriendlyFireMultipliers or something along those lines.

Exiled.API/Features/Player.cs
```diff
@@ -128,6 +128,21 @@ public Player(GameObject gameObject)
...
+
+        /// <summary>
+        /// Gets or sets a unique custom role that does not adbide to base game for this player. Used in conjunction with <see cref="UniqueFriendlyFireRules"/>.
+        /// </summary>
+        public string UniqueRole { get; set; } = string.Empty;
```
joker-119, line 144: Why would this feature not just be integrated directly into CustomRoles where it belongs?

joker-119:
Also, Will this friendly fire stuff work if the Server's FriendlyFire config is set to false? Because if not, it ignores the entire purpose of having this player property in the first place.
What will happen if server FF is disabled? It looks like that config is now just completely disabled and all servers will be FF enabled, unless I am missing something?

To be honest, this entire new system around friendlyfire seems a little over top to be included in EXILED. This looks like a plugin feature trying to masquerade as an API feature.

xRoier:
Give this man Dev Boi rn

Undid-Iridium:
> Your custom build events need to have conditions to handle both Windows and Linux, as their commands are different and as it is now, it makes EXILED only buildable on windows.

Bet, I will add conditions and test with linux

Exiled.API/Features/Player.cs
Undid-Iridium: Sure

Exiled.API/Features/Player.cs
Undid-Iridium: Sure

IRacle1:
imho, "UniqueClasses" things i think are not needed or must be in CustomRoles.
Also `DebugProduction` is useless, EXILED already have `EnvironmentType::Development` and debugging with preprocessor directives

Exiled.API/Features/Player.cs
Undid-Iridium: Because a lot of people rather make custom roles themselves, and while I agree that it should all be consolidated to reduce issues - we live in reality where that is not necessarily going to happen. Therefore, it is better to provide some level of accessibility rather than not doing anything at all and having 200x versions, at least if I give some level of control, it reduces how far the variants go, hopefully.

Undid-Iridium:
> imho, "UniqueClasses" things i think are not needed or must be in CustomRoles. Also `DebugProduction` is useless, EXILED already have `EnvironmentType::Development` and debugging with preprocessor directives

See previous comment above about differences

Undid-Iridium:
@joker-119 All in all, ultimately this is a feature quest - If you, and the other dev-bois/contributors believe this is overkill, not needed, or should be consolidated into CustomRoles, then I understand and will follow your lead. This is not really my choice, but rather my suggestion. 



> Also, Will this friendly fire stuff work if the Server's FriendlyFire config is set to false? Because if not, it ignores the entire purpose of having this player property in the first place. What will happen if server FF is disabled? It looks like that config is now just completely disabled and all servers will be FF enabled, unless I am missing something?
> 
> To be honest, this entire new system around friendlyfire seems a little over top to be included in EXILED. This looks like a plugin feature trying to masquerade as an API feature.

image

I left that in there from your original design; however, I have been testing this entire time with FF off; therefore, your concern is taken care of. However, more testing can and should be done.

To your last point, I personally feel that it is a benefit to the system rather than a hindrance. I have seen many implementations of custom roles and if statement hell associated with custom role 035, 966, blah blah blah. People who can't or don't want to use transpilers doing very janky solution to bypass FF, which fails sometimes, inconsistent/edge case hell, etc. I wrote this to solve those problems; however, I am not a dev-boi, I do not get to choose what goes into exiled. This is, as I said before, a suggested feature. Thanks all for taking your time to look, I appreciate it

Exiled.Loader/Config.cs
Undid-Iridium: Also, to your point about preprocessor directives, yeah that's fair - that is more efficient than a new boolean flag check on log.debug; however, there were examples of that so I went down that route. Agreed

Undid-Iridium:
Additionally, if you want I can showcase the differences of behavior both for the transpilers I replaced/fixed, and the new FF feature. Whether that be by recording, or streaming it in discord/twitch, if that's something you would like to verify overall changes. 

Furthermore, I can't seem to find any information that states VS20X post build events behave different on linux - it claims it uses a DOS based implementation 

image

I would presume that means it would operate fine. However, more research needs to be done

Exiled.API/Features/Player.cs
joker-119: This is 1. Stupid and 2. Counterintuitive
Firstly, you don't even incorporate this "UniqueRole" stuff in the existing CustomRoles API process. At the very least, player.UniqueRole should be set when they gain/lose a CustomRole.
Secondly, I don't really care what tinybrains decide to do for implementing their own custom roles, the EXILED method of doing it involves using the CustomRole API, and any code that is related to "custom player roles" should be contained soley within that API unless absolutely necessary for EXILED to function.

Undid-Iridium: 1. Would it not be your job, the creator of CustomRoles to incorporate it? This system is independent of that system, and would be accessibly by those components. If it is my job since I adding a feature, I suppose. However, they are not at the moment connected.
2. I disagree that users are tinybrain for doing their own CustomRole API's - I do agree it is something that can create issues; however, the CustomRole code is not perfect, or may limit their intents and purposes. I cannot claim that to be fully true though, as I have not played with the CustomRole API itself.
3. While I understand your position, I do not believe this is Counterintuitive - it simply allows another facet for users to control their own player, and their FF rules. Even with CustomRoles, you need to be able to identify a unique role, which this supplies. Maybe I need to actually delve into the CustomRoles code, but it seems quiet efficient to provide this functionality whether or not it is directly folded into CustomRoles. 
4. With all I have previously said, I will not force this down anyone's throat. You tell me you do not want this, and it's gone. I will abide to the rules set whether they conflict with my own views or not. I simply am offering an alternative

Exiled.Loader/Config.cs
Undid-Iridium: Though, I do wonder if it is simpler to just add a new Config EnvironmentType where

EnvironmentType.ProductionDebug where it would reduce this extra bool to just a different loadup config. This would then allow ShouldDebugBeShownProduction to disappear, and just add a flag to ShouldDebugBeShown to check if it is ProductionDebug

Exiled.API/Features/Player.cs
Undid-Iridium: I have read through your CustomRole code, I understand your concerns a little bit more although there are a few arguments to justify why someone would not want to use this utility (A plugin, in essence). However, I see many benefits in this system itself.

Undid-Iridium:
I added a fix for can_tutorial_block_scp173 because it was in the way of my FF testing

Exiled.API/Features/Player.cs
joker-119: 1. No because it is *you* adding the feature, so it is *you* who should incorporate it into the CustomRoles API that's included in EXILED.
2. Irrelevant. Since EXILED offers a CustomRole API, all of our 'CustomRole' related things should be contained within it.
3. They could manage this themselves via their plugin that they would need to make to add these rules anyways. Again, it seems like this is a feature that belongs in a plugin like Common Utilities, but is shoved into the API instead.
4. I'm not completely opposed to having the role-specific FF multipliers as part of the Player API, but I think the "custom role" stuff should be moved to and managed by the CustomRole API, as that would make much more sense in the context of a plugin properly using the EXILED APIs. If someone wants to make their own custom Role system, they can implement FF multipliers in that system themselves. It shoudn't be our responsibility to account for that, when we already provide a system for custom roles.  It would be akin to accounting for people making custom items without using our custom item API.

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -11,55 +11,144 @@ namespace Exiled.Events.Patches.Generic
     using System.Collections.Generic;
```
joker-119, line 11: All of the changes in this file need to be removed from this PR and put in their own, as the changes are not dependant on this PR and do not have anything to do with the other features implemented in this PR.

Exiled.Loader/Config.cs
joker-119: Adding a new EnvironmentType is also acceptable.

joker-119:
> image
> 
> I left that in there from your original design; however, I have been testing this entire time with FF off; therefore, your concern is taken care of. In essence, the transpiler adds unique FF scenario, and if the new CheckFriendlyFirePlayer logic returns false, then we default to base game FF logic. However, more testing can and should be done.

I missed that, I just want to make sure that in an environment where a plugin has not added any special FF rules, Server.FriendlyFire is still obeyed, both when it is true and when it is false.

> To your last point, I personally feel that it is a benefit to the system rather than a hindrance. I have seen many implementations of custom roles and if statement hell associated with custom role 035, 966, blah blah blah. People who can't or don't want to use transpilers doing very janky solution to bypass FF, which fails sometimes, inconsistent/edge case hell, etc. I wrote this to solve those problems; however, I am not a dev-boi, I do not get to choose what goes into exiled. This is, as I said before, a suggested feature. Thanks all for taking your time to look, I appreciate it

I do understand there are janky solutions for custom roles and they have janky FF implementations that barely work. However this is due to either a lack of knowing the CustomRole API exists, or a lack of ability to learn how it works based on the code in the public plugins adding custom roles properly.
I honestly don't feel that it should be our job to not only provide an API to developers, but also hold their hands when they are not smart enough to use said API.

Of the Custom role plugins I've seen, I'd say about half do not use the API we provide, and of those, I'd say only 2-3 in total were implemented well and didn't have a plethora of issues, FF issues typically being the least major of said issues. I think our time would be better spent on making the Custom Role API easier to use, fix any bugs still existing in it, or make it better documented, so that devs can/will use it more, rather than providing tools to the devs who can't be bothered to use it.

joker-119:
> Additionally, if you want I can showcase the differences of behavior both for the transpilers I replaced/fixed, and the new FF feature. Whether that be by recording, or streaming it in discord/twitch, if that's something you would like to do to verify overall changes.
> 
> Furthermore, I can't seem to find any information that states VS20X post build events behave different on linux - it claims it uses a DOS based implementation
> 
> image
> 
> I would presume that means it would operate fine. However, more research needs to be done
> 
> Update 1: Nevermind, apparently you can't run VS2022 on Linux, https://docs.microsoft.com/en-us/visualstudio/releases/2022/system-requirements
> 
> Update 2:
> 
> https://stackoverflow.com/questions/38683634/how-to-make-postbuildevent-work-in-project-rider
> 
> You can use AfterBuild target from MSBuild. PostBuildEvent it is Visual Studio specific way. https://blogs.msdn.microsoft.com/msbuild/2005/10/12/how-to-beyond-prebuild-and-postbuild-steps/
> 
> Therefore, it should not break linux based rider dev env's (As VS20X is not on Linux, apparently).

Exec Command executes the command in the same way as it would execute in a terminal, on Linux. Since `copy` isn't a command on linux, it will fail, and while the dll's will be built, the build will fail when it tries to execute the AfterBuild event.
I'm honestly not sure how Linux (or Rider in general) will handle there being Visual Studio specific build events in the csproj files, but if I remember correctly, it breaks the build entirely.

joker-119:
Also you should remove the post build events from the Bootstrap, and make sure the Loader and API dlls go into ../ and /dependencies as appropriate

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: Done https://github.com/Exiled-Team/EXILED/pull/1239

Exiled.API/Features/Player.cs
Undid-Iridium: 1. Alright.. More work :( 
2. Sure
3. Alright Alright, I get it
4. Fair, I will fold it into your CustomRole code (basically just going to move the dict over, and then during transpiler check "Is this person a custom role, ok check their custom role dict". I will also copy over the helper functions into CustomRoles

Undid-Iridium:
> > image
> > I left that in there from your original design; however, I have been testing this entire time with FF off; therefore, your concern is taken care of. In essence, the transpiler adds unique FF scenario, and if the new CheckFriendlyFirePlayer logic returns false, then we default to base game FF logic. However, more testing can and should be done.
> 
> I missed that, I just want to make sure that in an environment where a plugin has not added any special FF rules, Server.FriendlyFire is still obeyed, both when it is true and when it is false.
> 
> > To your last point, I personally feel that it is a benefit to the system rather than a hindrance. I have seen many implementations of custom roles and if statement hell associated with custom role 035, 966, blah blah blah. People who can't or don't want to use transpilers doing very janky solution to bypass FF, which fails sometimes, inconsistent/edge case hell, etc. I wrote this to solve those problems; however, I am not a dev-boi, I do not get to choose what goes into exiled. This is, as I said before, a suggested feature. Thanks all for taking your time to look, I appreciate it
> 
> I do understand there are janky solutions for custom roles and they have janky FF implementations that barely work. However this is due to either a lack of knowing the CustomRole API exists, or a lack of ability to learn how it works based on the code in the public plugins adding custom roles properly. I honestly don't feel that it should be our job to not only provide an API to developers, but also hold their hands when they are not smart enough to use said API.
> 
> Of the Custom role plugins I've seen, I'd say about half do not use the API we provide, and of those, I'd say only 2-3 in total were implemented well and didn't have a plethora of issues, FF issues typically being the least major of said issues. I think our time would be better spent on making the Custom Role API easier to use, fix any bugs still existing in it, or make it better documented, so that devs can/will use it more, rather than providing tools to the devs who can't be bothered to use it.

Understood, see https://github.com/Exiled-Team/EXILED/pull/1231#discussion_r873169784 but in essence I will make the changes to fold into CustomRoles, if anything someone can look through the commit history/ask how to do it the non-custom role way which the answer will be here. Thanks for the feedback and response/help!

Exiled.API/Features/Player.cs
Undid-Iridium: @joker-119 

My brain might be a little fried from fixing the last few things; however, let me know if this is dumb.

The minimum you stated was:
"Firstly, you don't even incorporate this "UniqueRole" stuff in the existing CustomRoles API process. At the very least, player.UniqueRole should be set when they gain/lose a CustomRole." 

and 

"I'm not completely opposed to having the role-specific FF multipliers as part of the Player API, but I think the "custom role" stuff should be moved to and managed by the CustomRole API, as that would make much more sense in the context of a plugin properly using the EXILED APIs. If someone wants to make their own custom Role system, they can implement FF multipliers in that system themselves. It shoudn't be our responsibility to account for that, when we already provide a system for custom roles. It would be akin to accounting for people making custom items without using our custom item API."

That means you will be at least alright with:

AddRole(){
....
// Something unique for the role since I assume there's no exact name for the custom role themselves
player.UniqueRole =  player.ReferenceHub.GetInstanceID().ToString();
}

Or something like that - If that is not accurate/dumb let me know. If that is the minimum you want, then everything else can remain the same, correct? If I were to actually move everything to CustomRoles then it would mean the transpiler would not exactly have access the information it needs. What I exactly mean is

Current code checks Player.(Custom..) and then does logic based off that. Because CustomRoles is a plugin, I cannot just do

Dict<CustomRoles, RoleRules..> in the Player class, meaning I would need to pass over the CustomRole object, at least to the player one, but I can't do that either because it is an external tool (and it should not be forced in).

Therefore, It makes sense that the role-specific FF stays a part of it, and CustomRoles just does initialization in a simpler way. Meaning, I would update custom roles to have the config file specify "What are the FF rules you will allow" similar to

  friendly_fire_rules:
    ClassD: 0
    Scp173: 1
    Scp93953: 2
    Scp93989: 3

So that for every custom role you define, you can specify in the config what rules should be associated. That's the only way I can think of right now on how to approach it. Let me know what you think, thanks

Undid-Iridium:
I have done more testing, both with CustomRole, and directly modifying the player code. It should work fine, but I am still not 100% sure if I am adhering to the type of architecture you may want. 

image

As for testing, CustomRole seems to work with the changes I've provided. However, this is a bug in game where Repressor is being viewed as Marauder. I believe Cyanox reported it for the visual affect but apparently the role itself is also reversed, as far I can tell. I can do more testing but it is an oddity

Exiled.API/Exiled.API.csproj
```diff
@@ -33,4 +33,11 @@
     <Reference Include="UnityEngine.PhysicsModule" HintPath="$(EXILED_REFERENCES)\UnityEngine.PhysicsModule.dll" Private="false" />
   </ItemGroup>
 
+  <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
+    <PostBuildEvent>if not "$(EXILED_DEV_REFERENCES)"=="" copy /y "$(OutputPath)$(AssemblyName).dll" "$(EXILED_DEV_REFERENCES)\Plugins\dependencies\"</PostBuildEvent>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(OS)' == 'Unix' ">
+    <PostBuildEvent>if [[ ! -z "$EXILED_DEV_REFERENCES" ]]; then cp "$(OutputPath)$(AssemblyName).dll" "$EXILED_DEV_REFERENCES\Plugins\dependencies\" fi</PostBuildEvent>
```
joker-119, line 40: The linux ones are incorrect. This may have been my fault when I was telling you what to use for them, but the correct code here should be:

`if [[ ! -z "$EXILED_DEV_REFERENCES" ]]; then cp "$(OutputPath)$(AssemblyName).dll" "$EXILED_DEV_REFERENCES/Plugins/dependencies/"; fi`

Note the different slashes (`/` instead of `\`) and the `;` before `fi`

Undid-Iridium: See newest commit, thanks!

joker-119 (approved):

PR #1234, opened 2022-05-14
Server.SessionVariables

steven4547466:
Just makes it easier to put session variables on the server if people forget `Server.Host` exists (definitely not me).


steven4547466:
Alternatively it could just be made into its own dictionary, I just thought using the Host's session variables would be the easiest.

steven4547466:
Under more consideration I feel it would be better if it were its own dictionary so that it can persist across rounds.

joker-119 (approved):

PR #1238, opened 2022-05-15
feat(loader): allow to load plugins derived from an abstract `Plugin`

Nekonyx:
This is a very useful feature if we have our own abstract `Plugin` class that inherits from `Exiled.API.Features.Plugin`, with the predefined fields, properties, and methods needed for a group of plugins of a certain project:
```cs
public abstract class OurPlugin<TConfig, TTranslation> : Plugin<TConfig, TTranslation> ...
{
  ...


  public abstract void OnLoaded();

  ...
}
```

This patch doesn't break anything and everything works as intended.


IRacle1:
I think in this case you can use [Type::IsSubclassOf](https://docs.microsoft.com/en-us/dotnet/api/system.type.issubclassof?view=netframework-4.7.2)

Nekonyx:
> I think in this case you can use [Type::IsSubclassOf](https://docs.microsoft.com/en-us/dotnet/api/system.type.issubclassof?view=netframework-4.7.2)

Perhaps. I just took the existing way of checking into a separate method and threw it into the loop.

PR #1239, opened 2022-05-15
Bug fix to Scp 173 being looked at for can_tutorial_block

Undid-Iridium:
Fix for problem is to inject check before if(flag) to prevent the observingPlayers to be updated. Attempting to add then remove then add is not sufficient because this code is not particularly safe. It is easier to prevent the player from being added than trying to remove them with the system NW designed.


NaoUnderscore:
Already fixed. The issue was the patch being commented. It works as it as.

PR #1242, opened 2022-05-16
DisplayName is better like this

louis1706:


louis1706:
It's just better because it's should never return an null value instead it's return the nickname
DisplayName

Jesus-QC (approved):

PR #1243, opened 2022-05-16
Prims can't be shot through anymore

rayzerbrain:


louis1706:
pretty strange because it's should be actualise in the Game 
AutoUpdate
So it's possibly MER Who cancel this update for optimisation

rayzerbrain:
It happens to me without any other plugins interfering with things. It seems to work fine spawning them through remote admin because this line is part of basegame code as well
image

Jesus-QC:
Yeah, Syncvars are kinda borked on spawn so this may happen.

PR #1244, opened 2022-05-17
Small features

IRacle1:


IRacle1:
Also, i think that constant lists/dictionaries types need set to `ReadOnly`, instead of normal (example `Exiled::API::ExtensionsDamageTypeExtensions::ItemConversion`, you can change it via plugin)

joker-119:
I agree, consistent list/dicts that never change should be made readonly, I'm not sure if that would be considered a breaking change in the case of public ones though.

If you could make a quick little plugin that uses some of them as they are now, then rebuild EXILED with them as readonly to make sure that plugin doesn't hit a missing method exception because the signature changes, then you can change them over if you want and it doesn't break the plugin.

joker-119:
> You don't understand me, I want to change `Dictionary` into `ReadOnlyDictionary` Not `readonly` modifier

Oooohhhh, right.
Yea, I mean I still think that's a good idea, and should be done at some point, but if they are public or protected currently that will need to wait until a major version change to implement.

joker-119 (approved):

IRacle1:
Okay, I'll write PR to make it ~~(if lazy will not seize me)~~

> > You don't understand me, I want to change `Dictionary` into `ReadOnlyDictionary` Not `readonly` modifier
> 
> Oooohhhh, right. Yea, I mean I still think that's a good idea, and should be done at some point, but if they are public or protected currently that will need to wait until a major version change to implement.

Okay, i'll write another PR to make this

BuildBoy12 (approved):

PR #1245, opened 2022-05-17
ZoneType for Exiled 6

louis1706:


louis1706:
Whe have two option this or Removed flags

Exiled.API/Features/Room.cs
```diff
@@ -320,20 +320,21 @@ private static RoomType FindType(string rawName)
...
-            Transform transform = gameObject.transform;
+            rawName = rawName.RemoveBracketsOnEndOfName();
 
-            if (transform.parent is null)
+            if (rawName.StartsWith("EZ_"))
+                return ZoneType.Entrance;
+            if (rawName.StartsWith("HCZ_"))
+                return rawName.StartsWith("HCZ_EZ_") ? ZoneType.EntranceHeavy : ZoneType.HeavyContainment;
+            if (rawName.StartsWith("LCZ_"))
+                return rawName.StartsWith("LCZ_Chkp") ? ZoneType.HeavyLightContainment : ZoneType.LightContainment;
```
louis1706, line 332: Not sure if people who want this

steven4547466:
Maybe for consistency across the other exiled flags, you should start it at 0 rather than 1.

louis1706:
> Maybe for consistency across the other exiled flags, you should start it at 0 rather than 1. Also, is pocket really inside the facility?

I considered yes because nuke kill all people in facility with pocket if i remind good

louis1706:
> Maybe for consistency across the other exiled flags, you should start it at 0 rather than 1. Also, is pocket really inside the facility?

And no flags should start at 1 because 
2^0 = 1
2^1 = 2 
Ect

EXILED.props
```diff
@@ -15,6 +15,7 @@
 
   <PropertyGroup>
     <!-- This is the global version and is used for all projects that don't have a version -->
+
```
zhaguanyang, line 18: unnecessary change

louis1706: > unnecessary change

Idk how it's appears but it's not really a problem ^^

louis1706: I go look after but i don't think i have commit it probably a merge of exiled dev have make this ^^'

PR #1248, opened 2022-05-18
Fix ``TurnOffAllLights(float, ZoneType)``

louis1706:


Jesus-QC:
Breaking Changes

steven4547466:
Does this not conflict with the fixes to the flag in #1245? As the only reason this doesn't work right now is because the flag is messed up, right? So if this is merged, eventually when #1245 is merged, this will become no longer necessary and prevent turning lights off in multiple zones at the same time.

joker-119 (approved):

louis1706:
> Not a breaking change.
> 
> This is ok as a temporary 'make it work' fix, but #1245 will need to make sure that these changes are reversed, or whatever is needed to make that PR's changes work properly.

It's because i have removed the flags property in ZoneType before

Jesus-QC:
It was a breaking change, now it isnt

Jesus-QC:
Should be reverted with #1245

PR #1254, opened 2022-05-21
`Player::DroppingItem` rework

IRacle1:


Exiled.Events/Patches/Events/Player/ItemDrop.cs
```diff
@@ -33,27 +33,29 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
+                // isThrow = ev.IsThrown;
                 new(OpCodes.Ldarg_0),
                 new(OpCodes.Ldfld, Field(typeof(InventorySystem.Inventory), nameof(InventorySystem.Inventory._hub))),
                 new(OpCodes.Call, Method(typeof(API.Features.Player), nameof(API.Features.Player.Get), new[] { typeof(ReferenceHub) })),
                 new(OpCodes.Ldarg_1),
-                new(OpCodes.Call, Method(typeof(ItemDrop), nameof(ItemDrop.GetItem))),
-                new(OpCodes.Stloc_S, item.LocalIndex),
-                new(OpCodes.Ldloc_S, item.LocalIndex),
+                new(OpCodes.Ldloca_S, item.LocalIndex),
+                new(OpCodes.Callvirt, FirstMethod(typeof(API.Features.Player), info => info.Name == "TryGetItem" && info.GetCustomAttribute<ObsoleteAttribute>() == null)),
```
IRacle1, line 58: It's a bad check here cause `TryGetItem` have `ref` parameter

Exiled.Events/Patches/Events/Player/ItemDrop.cs
```diff
@@ -33,27 +33,29 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
+            LocalBuilder ev = generator.DeclareLocal(typeof(DroppingItemEventArgs));
             Label returnLabel = generator.DefineLabel();
             Label notNullLabel = generator.DefineLabel();
 
             newInstructions.InsertRange(0, new[]
             {
-                // Item item = GetItem(player, itemSerial)
-                // if (item is null)
+                // if (!player.TryGetItem(itemSerial, out Item item)) {
                 //    Handlers.Player.OnDroppingNull(new DroppingNullEventArgs(Player));
```
IRacle1, line 44: I think that `OnDroppingNull` looks strange, so I would like to rename that to `OnDroppingEmpty` or `OnDroppingNothing`

Exiled.Events/Patches/Events/Player/ItemDrop.cs
IRacle1: It might be possible to do better

Exiled.Events/Patches/Events/Player/ItemDrop.cs
zhaguanyang: > I think that `OnDroppingNull` looks strange, so I would like to rename that to `OnDroppingEmpty` or `OnDroppingNothing`

That would be a breaking change, only allowed in EXILED 6.0

IRacle1: Yes, I just want to know dev bois opinion

Exiled.Events/Patches/Events/Player/ItemDrop.cs
joker-119: `new (OpCodes.Callvirt, Method(typeof(API.Features.Player), nameof(API.Features.Player.TryGetItem), new[] { typeof(ushort), typeof(API.Features.Items.Item) },` ?

IRacle1: > `new (OpCodes.Callvirt, Method(typeof(API.Features.Player), nameof(API.Features.Player.TryGetItem), new[] { typeof(ushort), typeof(API.Features.Items.Item) },` ?

Tested, not working

Exiled.Events/Patches/Events/Player/ItemDrop.cs
joker-119: Personally, I don't see why OnDroppingNull even exists. I don't remember who added it or why, but I see no reason it can't still call OnDroppingItem but with a null item property.

Exiled.Events/Patches/Events/Player/ItemDrop.cs
joker-119: What's the error, specifically? Invalid opcode during patching or something during runtime? Ideally we should get rid of that extra funky reflection, I know it's possible.
If nothing else we can re-make the helper method that calls TryGetItem() in plain C#, an extra call and stloc is still going to be faster than that reflection used in FirstMethod()

Exiled.Events/Patches/Events/Player/ItemDrop.cs
IRacle1: > Personally, I don't see why OnDroppingNull even exists. I don't remember who added it or why, but I see no reason it can't still call OnDroppingItem but with a null item property.

Okay, few reasons:

1. `ev.IsAllowed` will be useless
2. People will be catch one more NRE
3. This event will invoke when you rightclick on empty slot, a little different logic then `OnPickingUpItem`
So i dont see any reasons to delete this, but i dont like naming

joker-119: Fair points.

I think OnDroppingNothing would be the best choice for a rename.

You can implement it with the new name while keeping the original and marking it as obsolete, just make sure both events get called.

joker-119: That or comment the current event with a // TODO: so I will see it when I am prepping the 6.0 release.

Exiled.Events/Patches/Events/Player/ItemDrop.cs
IRacle1: No, its cause this method have a `ref` parameter, and I haven't found a way to get this method using `AccessTools.Method`
If there is no normal solution, I will make a helper method

joker-119: Tbh that may be why I ended up making a helper method originally, I probably couldn't figure it out either.

IRacle1: Close this review, i found solution

PR #1256, opened 2022-05-24
UsingBreakneckSpeeds event

steven4547466:
Adds an event to 173 for using breakneck speeds. 

Allows for bypassing the cooldown (i.e. it will fire with `ev.IsAllowed == false` if on cooldown, setting it to true will allow the ability to run regardless of this cooldown).


joker-119:
Forgive my lack of knowing the 173 rework since I don't play, what's the difference between this and blinking ?

steven4547466:
BreakneckSpeeds is used when a person presses shift and they move really fast

BuildBoy12 (approved):

joker-119 (approved):

PR #1257, opened 2022-05-24
`Exiled::CustomItem` PInv command Support

IRacle1:
Now, `pinv` command will be log only `ItemType`, and i make it to log CustomItem's names


IRacle1:
Also, CustomItems already had a `Harmony` instance, idk what for.

PR #1260, opened 2022-05-26
FF cleanup/fix, I was not done testing.

Undid-Iridium:
I was not done testing nor cleaning up. Includes fix for Server.FriendlyFire (return true was not correct, while it did give FF with a default value of 1, it should of defaulted back to NW logic, that was an incorrect interpretation on my part. That way it would default to whatever the user specified, if the Server.FriendlyFire is false, then we do custom FF logic if there is any, if not it falls into the last return false. I was busy with work so I was not testing the last few days but since it was merged I sat down and did my final testing (Unless something comes up).  @NaoUnderscore I would recommend you merge this before you do a release. The already merged branch won't "break" but it won't have the correct logic for Server.FriendlyFire. If this break anything after people use it, I will fix it, but it should be fine from my testing. I should of stated I was going to do a final testing of everything, my apologies


NaoUnderscore:
I will consider this a bug report, hence will directly push it.

PR #1267, opened 2022-05-30
C# 9

louis1706:


Exiled.API/Features/Player.cs
```diff
@@ -272,7 +272,6 @@ public AuthenticationType AuthenticationType
 
         /// <summary>
         /// Gets or sets the player's display nickname.
-        /// May be <see langword="null"/>.
```
Michal78900, line 275: I dont't now why would you remove that.

louis1706: Because it's can't be null now

BuildBoy12: #1268

PR #1269, opened 2022-05-31
[Breaking Change] DisplayName

louis1706:


PintTheDragon (changes requested):
Could you please clarify the difference between the three nicknames.

louis1706:
**For CustomName**
Gets or sets the player's nickname if not null ortherwise it's send the original nickname.
**For Nickname**
Gets the original player's nickname. 
**For DisplayNickname**
Gets or sets the player's display nickname.

**Pseudo "Yamato#6246"
Rename with server "YamatoSenpai"**

Nickname // Yamato#6246
CustomName // YamatoSenpai
DisplayNickname // YamatoSenpai*

**Pseudo "Yamato#6246"
Not rename so it's will be null**

Nickname // Yamato#6246
CustomName // 
DisplayNickname // Yamato#6246

PR #1272, opened 2022-06-01
[Breaking change] Scp914::Start(Player)

louis1706:


rayzerbrain:
I don't think the basegame method even uses the ref hub parameter..

louis1706:
> I don't think the basegame method even uses the ref hub parameter..

If it's don't use it i should get an error :) and i don't have any error and you can see 
Server.Host.ReferenceHub was used before so i don't know why you say that

rayzerbrain:
> > I don't think the basegame method even uses the ref hub parameter..
> 
> If it's don't use it i should get an error :) and i don't have any error and you can see Server.Host.ReferenceHub was used before so i don't know why you say that

Not that it doesn't have it as a parameter, but that ```ServerInteract()``` literally does not use it anywhere. I think the reason it passed host automatically was because changing the parameter doesn't affect anything.

louis1706:
> > > I don't think the basegame method even uses the ref hub parameter..
> > 
> > 
> > If it's don't use it i should get an error :) and i don't have any error and you can see Server.Host.ReferenceHub was used before so i don't know why you say that
> 
> Not that it doesn't have it as a parameter, but that `ServerInteract()` literally does not use it anywhere. I think the reason it passed host automatically was because changing the parameter doesn't affect anything.

It does for the Event (and possible log of server)

PR #1273, opened 2022-06-02
Bugfixes `and short form`

IRacle1:


Exiled.API/Extensions/DamageTypeExtensions.cs
```diff
@@ -23,13 +22,38 @@ public static class DamageTypeExtensions
         /// <summary>
         /// Gets conversion information between <see cref="DeathTranslation.Id"/>s and <see cref="DamageType"/>s.
         /// </summary>
-        public static Dictionary<byte, DamageType> TranslationIdConversion { get; } = // TODO: To readonly
-            TranslationConversion.ToDictionary(key => key.Key.Id, value => value.Value);
+        public static Dictionary<byte, DamageType> TranslationIdConversion { get; } = new() // TODO: ReadOnlyDictionary
```
IRacle1, line 25: Remove LINQ, because sometimes it thrown `TranslationConversion` NRE

joker-119:
Was the removal of the separate patch for EnteringFemurBreaker event necessary to fix some sort of bug? 
Because if not, this combining of the two patches shouldn't be done.

IRacle1:
Because `CheckForLurePatch` wasn't working lol, and it's will fix all `Scp106Container` things

NaoUnderscore:
The short form can jump a less amount of bytes; using the short form doesn't mean making everything more maintainable nor optimized, it's just about a logical thing. That said, did you test everything to make sure it works properly without some random weird errors during patch process and/or in-game processes?

IRacle1:
I changed logic only `FailingEscapePocket` and `LureContainer` patch, and yes, i tested

joker-119:
I mean, it's the microest of micro optimizations, but if it's tested and working, I see no issue.

Exiled.Events/Patches/Generic/CheckForLurePatch.cs
```diff
@@ -47,7 +53,12 @@ private static void CheckPlayers()
             {
                 if (Scp106Container.CanBeKilled(player))
                 {
-                    player.Hurt(new UniversalDamageHandler(-1f, DeathTranslations.UsedAs106Bait, null));
+                    var ev = new EnteringFemurBreakerEventArgs(player);
```
joker-119, line 56: Please do not use 'var'.
You can use `EnteringFemurBreakerEventArgs ev = new(player);` instead.

joker-119 (approved):

PR #1274, opened 2022-06-02
Client desync Obsolete

louis1706:


joker-119:
We should find a way to implement these properly without causing desync, rather than just removing their use entirely. Has anyone investigated where the desync occurs and what causes it? Is it a client issue (IE: The client doesn't sync to the new values properly?)?

louis1706:
> We should find a way to implement these properly without causing desync, rather than just removing their use entirely. Has anyone investigated where the desync occurs and what causes it? Is it a client issue (IE: The client doesn't sync to the new values properly?)?

For implement it you should change the Player Speed Client side without changing it ServerSide

joker-119 (approved):

louis1706:
Oups i am going to fix it later

PR #1276, opened 2022-06-03
Added reconnect and connect to another port methods 

Klarulor:


NaoUnderscore:
PR should always target dev branch.
Also, `Player::Reconnect` is already a thing, why should we implement more overloads that are clearly useless?

Klarulor:
im stupid, thanks!

PR #1282, opened 2022-06-12
Little `IndividualFriendlyFire` fix

IRacle1:


Exiled.Events/Patches/Generic/IndividualFriendlyFire.cs
```diff
@@ -291,10 +279,7 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
                 // targetReferenceHub
                 new(OpCodes.Ldloc_3),
 
-                new(OpCodes.Ldarg_0),
```
Undid-Iridium, line 294: This patch is disabled - why not just delete at this point? If it's needed again it's in the commit history

Exiled.Events/Patches/Generic/IndividualFriendlyFire.cs
```diff
@@ -123,7 +111,7 @@ public static bool CheckFriendlyFirePlayerRules(ReferenceHub attackerHub, Refere
                         }
                     }
                 }
-                else if(!attacker.UniqueRole.Equals(string.Empty))
+                else if(!string.IsNullOrEmpty(attacker.UniqueRole))
```
Undid-Iridium, line 114: This should never be null unless a developer screwed up but sure - this works as well

Exiled.Events/Patches/Generic/IndividualFriendlyFire.cs
IRacle1: Yea, you right

Exiled.Events/Patches/Generic/IndividualFriendlyFire.cs
IRacle1: You're right, but `IsNullOrEmpty` makes more sense here then `Equals`, and an extra check for `null` wouldn't hurt

Exiled.Events/Patches/Generic/IndividualFriendlyFire.cs
IRacle1: Although no, it is marked as TODO, so let it exist(maybe ask dev bois later)

Undid-Iridium: > Although no, it is marked as TODO, so let it exist(maybe ask dev bois about removing it later)

That was Joker, and it was left there by me incase I screwed up the ff stuff for grenades, thus far I have not heard a report so, I am going to assume it's fine to remove. Though you can double check

IRacle1: Sure, you right

BuildBoy12 (approved):

IRacle1:
That wasn't bug, but okay

NaoUnderscore:
That was a fix, that's all.

PR #1283, opened 2022-06-12
Bug fix scp173 looked at

Undid-Iridium:
Based on @zhaguanyang server testing, prefix did not cause unstoppable peanut. Therefore, this is a transpiler version (and prefix is commented out in case of need to revert to it).

The only thing I did not add is the hashset NW pool because I never really used it in the prefix and I don't really understand it quite well (other than some locking mechanism maybe?). I don't think it served a purpose but it was there anyway so I remove it in transpiler, so that's really the only differences.

### DIFF
--- a/Exiled.API/Features/Scp173.cs
+++ b/Exiled.API/Features/Scp173.cs
@@ -17,6 +17,6 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets a list of player ids who will be turned away from SCP-173.
         /// </summary>
-        public static HashSet<Player> TurnedPlayers { get; } = new(20);
+        public static HashSet<Player> TurnedPlayers { get; } = new(32);
     }
 }
--- a/Exiled.API/Features/Scp173.cs
+++ b/Exiled.API/Features/Scp173.cs
@@ -17,6 +17,6 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets a list of player ids who will be turned away from SCP-173.
         /// </summary>
-        public static HashSet<Player> TurnedPlayers { get; } = new(32);
+        public static HashSet<Player> TurnedPlayers { get; } = new(20);
     }
 }
--- a/Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
+++ b/Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
@@ -144,27 +144,19 @@ namespace Exiled.Events.Patches.Generic
         public static bool SkipPlayer(ref PlayableScps.Scp173 instance, GameObject curPlayerGameObj, ReferenceHub curPlayerHub)
         {
             Player player = Player.Get(curPlayerGameObj);
-            if (player != null)
+            if (player is not null)
             {
                 if (player?.Role?.Type == RoleType.Tutorial)
                 {
                     if (!Exiled.Events.Events.Instance?.Config?.CanTutorialBlockScp173 ?? false)
                     {
-                        if (instance._observingPlayers.Contains(curPlayerHub))
-                        {
-                            instance._observingPlayers.Remove(curPlayerHub);
-                        }
-
+                        instance._observingPlayers.Remove(curPlayerHub);
                         return true;
                     }
                 }
                 else if (API.Features.Scp173.TurnedPlayers.Contains(player))
                 {
-                    if (instance._observingPlayers.Contains(curPlayerHub))
-                    {
-                        instance._observingPlayers.Remove(curPlayerHub);
-                    }
-
+                    instance._observingPlayers.Remove(curPlayerHub);
                     return true;
                 }
             }
@@ -180,7 +172,7 @@ namespace Exiled.Events.Patches.Generic
             Label cnt = generator.DefineLabel();
             Label isTutorial = generator.DefineLabel();
 
-            LocalBuilder player = generator.DeclareLocal(typeof(Player));
+            LocalBuilder currentPlayer = generator.DeclareLocal(typeof(Player));
             LocalBuilder scp173Player = generator.DeclareLocal(typeof(Player));
             LocalBuilder turnedPlayers = generator.DeclareLocal(typeof(HashSet<Player>));
 
--- a/Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
+++ b/Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
@@ -9,6 +9,7 @@ namespace Exiled.Events.Patches.Generic
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Reflection.Emit;
     using System.Text;
 
@@ -34,18 +35,9 @@ namespace Exiled.Events.Patches.Generic
     [HarmonyPatch(typeof(PlayableScps.Scp173), nameof(PlayableScps.Scp173.UpdateObservers))]
     internal static class Scp173BeingLooked
     {
-        /// <summary>
-        /// Gets a.
-        /// </summary>
-#pragma warning disable SA1401 // Fields should be private
         public static LayerMask Scp173CurMask;
-#pragma warning restore SA1401 // Fields should be private
 
-        /// <summary>
-        /// Gets a.
-        /// </summary>
-        public static LayerMask Scp173Mask
-        {
+        public static LayerMask Scp173Mask {
             get
             {
                 if (Scp173CurMask == 0)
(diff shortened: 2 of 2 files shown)


Undid-Iridium:
**Do not merge** until transpiler is tested by others as well

zhaguanyang (commented):
Tested both prefix and transpiler versions on my public server. Looks good to me. Didn't encounter the issue which 173 can move freely.

zhaguanyang (commented):
Why did you added an empty GlobalSuppressions file to Exiled.Events?

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -7,13 +7,20 @@
...
     using System.Collections.Generic;
+    using System.Linq;
     using System.Reflection.Emit;
+    using System.Text;
 
     using Exiled.API.Features;
 
     using HarmonyLib;
 
+    using InventorySystem.Items.Firearms.Modules;
```
IRacle1, line 20: Useless usings

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -22,12 +29,141 @@ namespace Exiled.Events.Patches.Generic
 
     using static HarmonyLib.AccessTools;
 
+    ///// <summary>
```
IRacle1, line 32: Why not delete these comments?

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -22,12 +29,141 @@ namespace Exiled.Events.Patches.Generic
...
+        /// <param name="instance"> Scp173 instance <see cref="PlayableScps.Scp173"/>. </param>
+        /// <param name="curPlayerGameObj"> Current player game object. </param>
+        /// <param name="curPlayerHub"> Current player referencehub. </param>
+        /// <returns> True if to be skipped, false if not. </returns>
+        public static bool SkipPlayer(ref PlayableScps.Scp173 instance, GameObject curPlayerGameObj, ReferenceHub curPlayerHub)
+        {
+            Player player = Player.Get(curPlayerGameObj);
+            if (player is not null)
+            {
+                if (player?.Role?.Type == RoleType.Tutorial)
```
IRacle1, line 149: You check `player` for `null` in line above, so `player?.` is not a valid construction(also, as I remember, `player.Role` can't be `null` too)

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -22,12 +29,141 @@ namespace Exiled.Events.Patches.Generic
...
+                {
+                    if (!Exiled.Events.Events.Instance?.Config?.CanTutorialBlockScp173 ?? false)
```
IRacle1, line 151: Interesting, I don't think `Config` can be null if the plugin is not `null`.

Exiled.API/Features/Scp173.cs
```diff
@@ -17,6 +17,6 @@ public static class Scp173
         /// <summary>
         /// Gets a list of player ids who will be turned away from SCP-173.
         /// </summary>
-        public static HashSet<Player> TurnedPlayers { get; } = new(20);
+        public static HashSet<Player> TurnedPlayers { get; } = new(32);
```
IRacle1, line 20: Why? :) 

I am not against this change, but it almost doesn't make sense

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -22,12 +29,141 @@ namespace Exiled.Events.Patches.Generic
...
+                    {
+                        instance._observingPlayers.Remove(curPlayerHub);
+                        return true;
+                    }
+                }
+                else if (API.Features.Scp173.TurnedPlayers.Contains(player))
+                {
+                    instance._observingPlayers.Remove(curPlayerHub);
+                    return true;
```
IRacle1, line 160: Hmmm, These `if` do the same thing, why not combine the 3 `if` into one?

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -36,63 +172,42 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
-                new(OpCodes.Ldc_I4_S, (int)RoleType.Tutorial),
-
-                // Skip to checking if tutorial should be blocked
-                new(OpCodes.Beq_S, isTutorial),
-
-                // Scp173.TurnedPlayers.Remove(Current 173)
-                new(OpCodes.Call, PropertyGetter(typeof(API.Features.Scp173), nameof(API.Features.Scp173.TurnedPlayers))),
-                new(OpCodes.Stloc, turnedPlayers.LocalIndex),
-                new(OpCodes.Ldloc, turnedPlayers.LocalIndex),
+            newInstructions.InsertRange(removeTurnedPeanut, new CodeInstruction[]
```
IRacle1, line 182: You can write this without `scp173Player` variable, something like:
```cs
new(OpCodes.Call, PropertyGetter(typeof(API.Features.Scp173), nameof(API.Features.Scp173.TurnedPlayers))),
new(OpCodes.Ldarg_0),
new(OpCodes.Ldfld, Field(typeof(PlayableScp), nameof(PlayableScp.Hub))),
new(OpCodes.Callvirt, Method(typeof(Player), nameof(Player.Get), new[] { typeof(ReferenceHub) })),
new(OpCodes.Callvirt, Method(typeof(HashSet<Player>), nameof(HashSet<Player>.Remove))),
new(OpCodes.Pop),
```

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -36,63 +172,42 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
             Label cnt = generator.DefineLabel();
             Label isTutorial = generator.DefineLabel();
 
-            int addCheckOffset = 4;
-            int addCheck = newInstructions.FindLastIndex(instruction => instruction.Calls(Method(typeof(Physics), nameof(Physics.Linecast), new[] { typeof(Vector3), typeof(Vector3), typeof(int) }))) + addCheckOffset;
-
-            LocalBuilder player = generator.DeclareLocal(typeof(Player));
-
+            LocalBuilder currentPlayer = generator.DeclareLocal(typeof(Player));
+            LocalBuilder scp173Player = generator.DeclareLocal(typeof(Player));
```
IRacle1, line 176: In transpiler, you use only `cnt` label and `scp173Player` local var(but you can do without it, review below), others local vars and labels are useless

IRacle1 (commented):
LGTM, but

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
```diff
@@ -22,12 +29,141 @@ namespace Exiled.Events.Patches.Generic
...
     internal static class Scp173BeingLooked
     {
+        /// <summary>
+        /// Checks if the current player is to be skipped.
+        /// </summary>
+        /// <param name="instance"> Scp173 instance <see cref="PlayableScps.Scp173"/>. </param>
+        /// <param name="curPlayerGameObj"> Current player game object. </param>
+        /// <param name="curPlayerHub"> Current player referencehub. </param>
+        /// <returns> True if to be skipped, false if not. </returns>
+        public static bool SkipPlayer(ref PlayableScps.Scp173 instance, GameObject curPlayerGameObj, ReferenceHub curPlayerHub)
```
IRacle1, line 144: Why do you need the `curPlayerGameObj` parameter when you can do `Player.Get(curPlayerHub)`?

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: kk

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: Fair, I just wanted someone to have the prefix to compare against

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: Fair enough, optimization is true, it was a copy paste where I was trying to protect against everything but yes, I will change, good catch

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: I would imagine so as well but I (Per my previous comment) was protecting against everything

Exiled.API/Features/Scp173.cs
Undid-Iridium: image
image

I don't know if there is a fundamental difference but, I don't see why we even have it set to a value. I guess to just optimize but is there a fundamental difference under the covers? If not, I'll put it back to 20 but I only upped it out of pure curiosity. I don't mind bringing it down, it _should not_ cause a difference I would hope

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: What if's do the same thing? They're completely different

Exiled.API/Features/Scp173.cs
IRacle1: Since you've already changed it, leave it as it is.

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
Undid-Iridium: Minor optimization but sure, that works

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
IRacle1: Ah, I see what you mean, you're right, then it would be too long `if`

Exiled.Events/Patches/Generic/Scp173BeingLooked.cs
IRacle1: I don't think you need to use gameObject, when you can do without it
(`Player.Get()` with gameObject is faster, yea, but it's not that critical.)

Undid-Iridium: My only other concern is, is there every a possibility that the reference hub, and or game object not match exiled version? (I am curious). With this method, it's impossible to be out of sync because I am providing what NW expects, so if Exiled doesn't have it, they're broken. Otherwise, it should cause no behavior difference. I am willing to remove it but only if we're confident that Exiled will not be something that will have a different result for a referencehub -> game object

IRacle1: Also, to get player gameObject, need 2 Instructions:
```cs
new(OpCodes.Ldloca_S, 2),
new(OpCodes.Callvirt, PropertyGetter(typeof(KeyValuePair<GameObject, ReferenceHub>), nameof(KeyValuePair<GameObject, ReferenceHub>.Key))),
```
So removing the gameObject parameter will make the transpiler smaller, but this is subjective, the existence of this parameter will not make it worse

Undid-Iridium: If we're confident Exiled will always have a correct matching gameobject to referencehub, I do not mind removing

IRacle1: But everything happens :)

Undid-Iridium: Alright, I'll update

IRacle1: > My only other concern is, is there ever a possibility that the reference hub, and or game object not match exiled version? (I am curious). With this method, it's impossible to be out of sync because I am providing what NW expects, so if Exiled doesn't have it, they're broken. Otherwise, it should cause no behavior difference. I am willing to remove it but only if we're confident that Exiled will not be something that will have a different result for a referencehub -> game object

Honestly, I've never heard of this logic of gameObjects behavior, if it can somehow cause problems in this patch, have Misaka check it out

Undid-Iridium: I've made the change, I only hope that it won't happen

PR #1285, opened 2022-06-17
Got feedback from owner's that they rather have this be toggable.

Undid-Iridium:
Long story short, they would prefer the game to have this odd behavior. Therefore, I made configurable.

### DIFF
--- a/Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
+++ b/Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
@@ -128,7 +128,7 @@ namespace Exiled.Events.Patches.Events.Map
         {
             foreach (Player player in players)
             {
-                if(Player.Get(grenade.PreviousOwner.Hub) == player)
+                if(Exiled.Events.Events.Instance.Config.CanFlashBangsAffectThrower && Player.Get(grenade.PreviousOwner.Hub) == player)
                 {
                     grenade.ProcessPlayer(player.ReferenceHub);
                 }
--- a/Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
+++ b/Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
@@ -38,13 +38,10 @@ namespace Exiled.Events.Patches.Events.Map
         {
             List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Shared.Rent(instructions);
 
-            if (Exiled.Events.Events.Instance.Config.CanFlashBangsAffectThrower)
-            {
-                // Remove check if player is thrower. Grenade on self should affect themself.
-                int removeSelfCheckOffset = -4;
-                int removeSelfCheck = newInstructions.FindIndex(instruction => instruction.LoadsField(Field(typeof(Footprint), nameof(Footprint.Hub)))) + removeSelfCheckOffset;
-                newInstructions.RemoveRange(removeSelfCheck, 7);
-            }
+            // Remove check if player is thrower. Grenade on self should affect themself.
+            int removeSelfCheckOffset = -4;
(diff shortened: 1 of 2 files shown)


Undid-Iridium:
Do not merge until I say so. Thanks

zhaguanyang:
> Do not merge until I say so. Thanks

Just put it in draft then

Exiled.Events/Patches/Events/Map/ExplodingFlashGrenade.cs
```diff
@@ -38,10 +38,13 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
         {
             List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Shared.Rent(instructions);
 
-            // Remove check if player is thrower. Grenade on self should affect themself.
-            int removeSelfCheckOffset = -4;
-            int removeSelfCheck = newInstructions.FindIndex(instruction => instruction.LoadsField(Field(typeof(Footprint), nameof(Footprint.Hub)))) + removeSelfCheckOffset;
-            newInstructions.RemoveRange(removeSelfCheck, 7);
+            if (Exiled.Events.Events.Instance.Config.CanFlashBangsAffectThrower)
```
IRacle1, line 41: If you already changed `ProcessPlayers`, why need to change this patch? if `Config.CanFlashBangsAffectThrower` will be false, same player never been flashed,  `reload cfgs` with `Config.CanFlashBangsAffectThrower` equals to `true`, will not affect this

I think check in `ProcessPlayers` will be enough, and `if (Exiled.Events.Events.Instance.Config.CanFlashBangsAffectThrower)` in transpiler cannot support dynamic update of configs

Undid-Iridium: I would recommend you look at the location being patched

image

If CanFlashBangsAffectThrower is true, patch will be applied (removal of NW self-check), and you can reach ProcessPlayer (for self-flash)
If CanFlashBangsAffectThrower is false, patch will not be applied, and you won't reach ProcessPlayer (for self-flash)

I am aware these configs are not hot-swappable, that's on the user to set it up as they desire. They can always add their own plugin to modify the function themselves. 

Furthermore, the check within ProcessPlayers is overkill, but I considered that in case someone called that function directly for some reason, the check should be there to ensure consistent behavior.

IRacle1: > If CanFlashBangsAffectThrower is true, patch will be applied (removal of NW self-check), and you can reach ProcessPlayer (for self-flash)
> If CanFlashBangsAffectThrower is false, patch will not be applied, and you won't reach ProcessPlayer (for self-flash)

Yes, i said about this, I think we need exactly dynamic configurations

> I am aware these configs are not hot-swappable, that's on the user to set it up as they desire. They can always add their own plugin to modify the function themselves.

I don't see why we can't do it dynamically if we can? Besides, we can just save your changes in `ProcessPlayers` and it will work perfectly

Undid-Iridium: > hey can always add their own plugin to modify the function themselves.

I believe I understand your point, keep the removal of lines that the transpiler handle the logic. My only thought process would be for the ExplodingGrenadeEventArgs but I think you're right - it would be fine to keep the line removal and keep process player logic

Undid-Iridium: Yep, I'm convinced, it should work. Thanks for the feedback!

IRacle1: > > hey can always add their own plugin to modify the function themselves.
> 
> I believe I understand your point, keep the removal of lines that the transpiler handle the logic. My only thought process would be for the ExplodingGrenadeEventArgs but I think you're right - it would be fine to keep the line removal and keep process player logic

Yea, that's what I mean

> Yep, I'm convinced, it should work. Thanks for the feedback!

My pleasure

Michal78900 (approved):

BuildBoy12 (approved):

PR #1287, opened 2022-06-19
This is a dumb way to fix Player.Hurt, but is simple.

Undid-Iridium:
Dumb fix, but temporary, also cassie doesn't seem to work well ~and on first incantation says 106 died~? Unsure what NW code on offline side is doing


Undid-Iridium:
So, I can't get Cassie to work. Seems to not care what I do, or do not put in it. Therefore, this is the closest I can get tonight. Additionally, I swapped over from AttackerDamageHandler because CustomReasonDamageHandler actually updated death text on screen versus AttackerDamageHandler (at least from testing). Any feedback is helpful, thanks

Undid-Iridium:
I would like some feedback on the cctor of GenericDamageHandler:

Does it make sense to have each type separated in the way it currently is, or to just have everything be a CustomReasonDamageHandler  If you want some effects (like particle disruptor) then you need to specify the way I did; otherwise, if we just care about damage, then CustomReasonDamageHandler can be one line of ```base = CustomReasonDamageHandler...```

I believe it is fine as is, because it allows specific unique outcomes (aka particle disruptor body decaying effect). If anyone has anything they want to add, I would appreciate it. Thanks

Exiled.API/Features/GenericDamageHandler.cs
```diff
@@ -0,0 +1,219 @@
...
+    using Footprinting;
+
+    using PlayerStatsSystem;
+
+    using Subtitles;
+
+    /// <summary>
+    /// Allows generic damage to player.
+    /// </summary>
+    internal class GenericDamageHandler : PlayerStatsSystem.CustomReasonDamageHandler
```
IRacle1, line 24: This should be in `Exiled.API\Features\DamageHandlers` directory

Exiled.Events/Patches/Events/Server/KillPlayer.cs
```diff
@@ -0,0 +1,52 @@
...
+    using Mirror;
+
+    using PlayerStatsSystem;
+#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
+
+    /// <summary>
+    /// Prefix of KillPlayer action.
+    /// </summary>
+    [HarmonyPatch(typeof(PlayerStats), nameof(PlayerStats.KillPlayer))]
+    internal class KillPlayer
```
IRacle1, line 29: You also need to move it to the `Fixes` directory

IRacle1 (commented):
I can't say anything about the implementation, I haven't studied the exact logic of Exiled's DamageHandlers, and besides, it's a temporary solution

Exiled.Events/Patches/Events/Server/KillPlayer.cs
IRacle1: or in `Generic`

Undid-Iridium: Done

Jesus-QC (approved):

NaoUnderscore (changes requested):
Mark the old `Player::Hurt` method as obsolete as, even if it's not currently working, removing it is considered a breaking change.

Undid-Iridium:
Wait, I'm blind, I'm using DamageHandler cassie, adding another method

IRacle1:
> Not sure what you mean, the method isn't removed, I just added an optional parameter - their method should work exactly the same

Even if the parameter you added is optional, the method will still be visible to the assembly as a different one, and the plugins that called the old method will break

Undid-Iridium:
> > Not sure what you mean, the method isn't removed, I just added an optional parameter - their method should work exactly the same
> 
> Even if the parameter you added is optional, the method will still be visible to the assembly as a different one, and the plugins that called the old method will break

Am I being dumb? If you update to 5.3, your assembly reference will get updated automatically, 
```Player attacker, float amount, DamageType damageType = DamageType.Unknown``` is the minimum required to identify the function, any player that adds
```Player attacker, float amount, DamageType damageType = DamageType.Unknown, Cassie..``` even before, called now, should be fine (as you can see above)

**Comment by Undid-Iridium** [comment]
Update: 
IRacle has educated my tiny brain, my understanding of complication and symbol resolution was wrong

http://web.archive.org/web/20160530082900/https://msdn.microsoft.com/en-us/library/aa691331(VS.71).aspx
http://web.archive.org/web/20160527145711/https://msdn.microsoft.com/en-us/library/aa691356(VS.71).aspx

File: Exiled.API/Features/DamageHandlers/GenericDamageHandler.cs [hunk 3]
```diff
@@ -0,0 +1,219 @@
...
+                    break;
+                case DamageType.Scp:
+                case DamageType.Scp018:
+                case DamageType.Scp207:
+                case DamageType.Scp049:
+                case DamageType.Scp173:
+                case DamageType.Scp939:
+                case DamageType.Scp0492:
+                case DamageType.Scp106:
+                    Base = new ScpDamageHandler(attacker.ReferenceHub, damage, DeathTranslations.Unknown);
```
louis1706, line 140: Why DeathTranslation Unknown?

Exiled.API/Features/DamageHandlers/GenericDamageHandler.cs
```diff
@@ -0,0 +1,219 @@
...
+                case DamageType.FemurBreaker:
+                case DamageType.PocketDimension:
+                case DamageType.FriendlyFireDetector:
+                case DamageType.SeveredHands:
+                    Base = new CustomReasonDamageHandler(damageText ?? genericDamageText, damage, cassieAnnouncement.Announcement);
+                    break;
+                case DamageType.Warhead:
+                case DamageType.Decontamination:
+                case DamageType.Tesla:
+                    Base = new CustomReasonDamageHandler(damageText ?? genericEnvironmentDamageText, damage, cassieAnnouncement.Announcement);
```
louis1706, line 78: Why custom again ?

louis1706 (commented):
Just want to know why

Exiled.API/Features/DamageHandlers/GenericDamageHandler.cs
```diff
@@ -0,0 +1,219 @@
...
+                case DamageType.Hypothermia:
+                case DamageType.Asphyxiation:
+                case DamageType.Poison:
+                case DamageType.Bleeding:
+                case DamageType.Crushed:
+                case DamageType.FemurBreaker:
+                case DamageType.PocketDimension:
+                case DamageType.FriendlyFireDetector:
+                case DamageType.SeveredHands:
+                    Base = new CustomReasonDamageHandler(damageText ?? genericDamageText, damage, cassieAnnouncement.Announcement);
```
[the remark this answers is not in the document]
Undid-Iridium: Because these may cause death, and honestly, I don't know whether to add a specific type here but CustomReason allows you specify death reason, so that's why it's here

Exiled.API/Features/DamageHandlers/GenericDamageHandler.cs
Undid-Iridium: Fear/Laziness - I didn't want to specify for each type, and I didn't know whether we were going to keep this. I will add each individual type if you want

Exiled.API/Features/DamageHandlers/GenericDamageHandler.cs
Undid-Iridium: Because these may cause death but are more environmental than damage affects

Undid-Iridium: Updated tesla

Exiled.Events/Patches/Fixes/KillPlayer.cs
```diff
@@ -0,0 +1,62 @@
...
+    internal class KillPlayer
+    {
+        /// <summary>
+        /// Handles processing of kill player to cast to correct damage handler.
+        /// </summary>
+        /// <param name="__instance"> PlayerStat instance. </param>
+        /// <param name="handler"> DamageHandlerBase instance. </param>
+        /// <returns> Always returns true as we want to run original function with correct DamageHandler. </returns>
+        [HarmonyPrefix]
+        public static bool KillPlayerPrefix(PlayerStats __instance, ref DamageHandlerBase handler)
```
BuildBoy12, line 42: Can be made private to remove the need for doc, named Prefix to remove attribute, and can be made return type void since it always returns true to run the method

Exiled.Events/Patches/Fixes/KillPlayer.cs
```diff
@@ -0,0 +1,62 @@
...
+                    KillingPlayerEventArgs ev = new KillingPlayerEventArgs(Player.Get(__instance._hub), ref handler);
+                    Handlers.Player.OnKillPlayer(ev);
+                    handler = ev.Handler;
+                }
+            }
+
+            return true;
+        }
+    }
+#pragma warning restore SA1313 // Parameter names should begin with lower-case letter
```
BuildBoy12, line 61: Restore at EOF seems unnecessary

Exiled.Events/Patches/Fixes/KillPlayer.cs
```diff
@@ -0,0 +1,62 @@
...
+    using Exiled.Events.Extensions;
+
+    using HarmonyLib;
+
+    using Mirror;
+
+    using PlayerStatsSystem;
+
+    using static Exiled.Events.Events;
+#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
```
BuildBoy12, line 27: Pragma disabling should occur before usings to match style

Exiled.API/Features/DamageHandlers/GenericDamageHandler.cs
```diff
@@ -0,0 +1,258 @@
...
+        public override float Damage { get; set; }
+
+        /// <inheritdoc />
+        public override string ServerLogsText { get; }
+
+        /// <summary>
+        /// Process damage for this custom damage source.
+        /// </summary>
+        /// <param name="ply"> Current player reference hub. </param>
+        public override void ProcessDamage(ReferenceHub ply)
```
BuildBoy12, line 214: Override is unnecessary

BuildBoy12 (changes requested):

Exiled.Events/Patches/Fixes/KillPlayer.cs
Undid-Iridium: @BuildBoy12 Done

Undid-Iridium:
I would like to transform this into a transpiler, but I am tired right now. Unless someone wants to do it immediately, it can wait for a post-merge change. Alternatively, someone could hijack the DyingEventArg (Died.cs) logic to add this behavior as the first to occur (since both are in KillPlayer func)

PR #1296, opened 2022-06-25
Exiled 6 NoSpammingSomeEvent

louis1706:


louis1706:
i will do it for Tantrum and making the Prefix (and ask help to make it Transpiler)

IRacle1:
Is this based on an SL update?

louis1706:
> Is this based on an SL update?

Yes

louis1706:
I let nao doing it

PR #1301, opened 2022-06-28
`InteractingGeneratorEvents` to transpiler

IRacle1:
Yep, I already tested and everything seems to be working.
Waiting for a feedback 
~~It was difficult to test when the dev branch does not support the new update~~


IRacle1:
@NaoUnderscore , can you run workflows again?

NaoUnderscore (changes requested):
That's over-commented. We do have to keep everything well documented, but still, they are patches. It's not even a huge transpiler, so we could just end up by removing all the comments, but that's not what I'm saying. Keep the most important ones, comments are made to understand what's going on for us, people who know how to deal with IL, it shouldn't be a tutorial for all the other devs or something. If you want to look at the patches' code, you should already know how IL works. That's why we are not commenting every single line of code in C# API.

PR #1305, opened 2022-07-01
`ScotopiaPatch` fix and other corrections

IRacle1:
Don't blame me Nao😭


Exiled.API/Features/Roles/Scp106Role.cs
```diff
@@ -28,7 +28,7 @@ public class Scp106Role : Role
         /// <summary>
         /// Gets the <see cref="Scp106PlayerScript"/> script for the role.
         /// </summary>
-        public Scp106PlayerScript Script => script ??= Owner.ReferenceHub.scp106PlayerScript;
+        public Scp106PlayerScript Script => script ? script : script = Owner.ReferenceHub.scp106PlayerScript;
```
NaoUnderscore, line 31: Why

Exiled.API/Features/Roles/Scp079Role.cs
```diff
@@ -28,7 +28,7 @@ public class Scp079Role : Role
         /// <summary>
         /// Gets the <see cref="Scp079PlayerScript"/> script for the role.
         /// </summary>
-        public Scp079PlayerScript Script => script ??= Owner.ReferenceHub.scp079PlayerScript;
+        public Scp079PlayerScript Script => script ? script : script = Owner.ReferenceHub.scp079PlayerScript;
```
NaoUnderscore, line 31: No

Exiled.API/Features/Roles/Scp0492Role.cs
```diff
@@ -30,7 +30,7 @@ internal Scp0492Role(Player player)
         /// <summary>
         /// Gets the <see cref="Scp049_2PlayerScript"/> for this role.
         /// </summary>
-        public Scp049_2PlayerScript Script => script ??= Owner.ReferenceHub.characterClassManager.Scp0492;
+        public Scp049_2PlayerScript Script => script ? script : script = Owner.ReferenceHub.characterClassManager.Scp0492;
```
NaoUnderscore, line 33: No

Exiled.API/Features/Player.cs
```diff
@@ -594,35 +594,35 @@ public bool NoClipEnabled
...
         /// Equivalent to checking the player's <see cref="Team"/>.
         /// </summary>
-        public bool IsScp => Role?.Team == Team.SCP;
+        public bool IsScp => Role?.Team is Team.SCP;
 
         /// <summary>
         /// Gets a value indicating whether or not the player's <see cref="RoleType"/> is any human rank.
         /// </summary>
-        public bool IsHuman => Role is not null && Role.Is(out HumanRole _);
+        public bool IsHuman => Role?.Team is not Team.SCP and not Team.RIP;
```
NaoUnderscore, line 620: No

NaoUnderscore (changes requested):

Exiled.API/Features/Roles/Scp106Role.cs
IRacle1: That check for unity objects, problem can be in caching objects that need to save not only in current scene(scp scripts dont need to save, but everything can be, if you sure that it won't lead to any errors - I revert it)

Exiled.API/Features/Player.cs
IRacle1: Why, type-cast is faster that 2 get-methods?

NaoUnderscore: ` Gets a value indicating whether or not the player's <see cref="RoleType"/> is any human rank.`
Human rank is defined by `HumanRole`, and even if it's faster than two get methods, we can't just assume that NW won't add more non-human roles. "We can add a `and not Team.NewNonHumanTeam` if that's the case", yes, but it goes against the maintainability and it leads to more issues.

Exiled.API/Features/Roles/Scp106Role.cs
NaoUnderscore: I don't really get it.
`script ??= Owner.ReferenceHub.scp106PlayerScript` is the same as `script ? script : script = Owner.ReferenceHub.scp106PlayerScript`.

IRacle1: No, check IL code

Exiled.API/Features/Player.cs
IRacle1: Okay, i'll revert it

Exiled.API/Features/Roles/Scp106Role.cs
IRacle1: Or ask Yamato, I already told him about diff

NaoUnderscore: Yeah whatever, it's the same thing in this case 🤷🏼‍♂️

PR #1306, opened 2022-07-01
Little Item API rework

IRacle1:
@Undid-Iridium here we go
My main idea was to add to the API methods `OnAdded` and `OnRemoved`, which would not be like in `Firearm(Firearm itemBase)::ctor`, this can be improved and finalized, I will try to do it, but so far this is a partial implementation of the idea
(also maybe that will fix `ChangingDurability` event, idk)


IRacle1:
Okay, so tomorrow i pr that to Yamato's pr, and add some new stuff

PR #1308, opened 2022-07-02
little change on MessageTranslated

louis1706:
- i thinks it's would be better like this but not sure


Thundermaker300 (approved):

IRacle1 (commented):
okay, why `size=0` is better? and why you're remove `alpha=#00` only in MirrorExtensions?

IRacle1:
But, you dont answer on first question
> Why size=0 is better?

IRacle1:
okay, seems logic

IRacle1 (approved):

Jesus-QC (approved):

PR #1309, opened 2022-07-03
CustomRole to CustomRole friendly fire damage extenstion (See discord link as to why)

Undid-Iridium:
https://discord.com/channels/656673194693885975/668963275790221346/992910603469463572

https://discord.com/channels/656673194693885975/668963275790221346/992571334188081293

tl;dr custom role to custom role friendly fire rules otherwise you get lazy approach 

I am very tempted to consolidate all of this to just the Dict<string, Pair< string, float>> to just avoid all of this code

### DIFF
--- a/Exiled.API/Features/FriendlyFireStatusCode.cs
+++ b/Exiled.API/Features/FriendlyFireStatusCode.cs
@@ -5,10 +5,12 @@
 // </copyright>
 // -----------------------------------------------------------------------
 
-namespace Exiled.API.Features
+namespace Exiled.API.Enums
 {
+    using Exiled.API.Features;
+
     /// <summary>
-    /// Status code for adding FriendlyFire rules.
+    /// Status code for adding FriendlyFire rules. <seealso cref="Player.FriendlyFireMultiplier"/>
     /// </summary>
     public enum FriendlyFireStatusCode
     {
--- a/Exiled.API/Features/FriendlyFireStatusCode.cs
+++ b/Exiled.API/Features/FriendlyFireStatusCode.cs
@@ -18,18 +18,13 @@ namespace Exiled.API.Features
         UnableToAdd,
 
         /// <summary>
-        /// Current role successfully added
-        /// </summary>
-        AddedRole,
-
-        /// <summary>
         /// Partial roles were added, cannot guarantee all.
         /// </summary>
-        AddedSomeRoles,
+        PartialSuccess,
 
         /// <summary>
         /// All roles were successfully added
         /// </summary>
-        AddedAllRoles,
+        Success,
     }
 }
--- a/Exiled.API/Features/FriendlyFireStatusCode.cs
+++ b/Exiled.API/Features/FriendlyFireStatusCode.cs
@@ -1,0 +1,35 @@
+// -----------------------------------------------------------------------
+// <copyright file="FriendlyFireStatusCode.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Features
+{
+    /// <summary>
+    /// Status code for adding FriendlyFire rules.
+    /// </summary>
+    public enum FriendlyFireStatusCode
+    {
+        /// <summary>
+        /// Adding specific role failed, unable to be added.
+        /// </summary>
+        UnableToAdd,
+
+        /// <summary>
+        /// Current role successfully added
+        /// </summary>
+        AddedRole,
+
+        /// <summary>
+        /// Partial roles were added, cannot guarantee all.
+        /// </summary>
+        AddedSomeRoles,
+
+        /// <summary>
+        /// All roles were successfully added
+        /// </summary>
+        AddedAllRoles,
+    }
+}
--- a/Exiled.API/Features/FriendlyFireStatusCode.cs
+++ b/Exiled.API/Features/FriendlyFireStatusCode.cs
@@ -10,8 +10,11 @@ namespace Exiled.API.Enums
     using Exiled.API.Features;
 
     /// <summary>
-    /// Status code for adding FriendlyFire rules. <seealso cref="Player.FriendlyFireMultiplier"/>
+    /// Status code for adding FriendlyFire rules.
     /// </summary>
+    /// <seealso cref="Player.FriendlyFireMultiplier"/>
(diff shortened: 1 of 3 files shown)


Undid-Iridium:
image

tl;dr

CustomRoles can now damage based on the specific rules set per player, meaning (as shown)

Both are Scp035; however, one has rules where their damage to Scp035's is 0, and the other has rules specifying they can do damage. In the picture you can see that only one of two can be damage in this scenario (The one whose attacker has damage activated), and the other has taken no damage

This also applies if you set the damage to 0, and any other roles added would have same rules.

NaoUnderscore:
I'd just close this PR as we are gonna re-make the CustomRole API soon.
Anyway, I'd say that's a bad idea since the word "CustomRole" shouldn't even be in the base API, that's a design pattern I, we, would like to keep.

However, you could make a component using the Exiled component system to make it as a temp solution, which is fine imo.

**Note: I'm not gonna close this PR nor telling you to close it, ask to all the other devs about their thoughts on this PR.**

Undid-Iridium:
@NaoUnderscore Sorry, I use the word "CustomRole" as a place holder for "CustomUniqueRole". The FF system does not support Unique Role to Unique Role damage currently; therefore, this is to fix that. This also includes CustomRole as well; however, this is specifically for people looking for FF rules for CustomUniqueRole (and subsequently CustomRole).

NaoUnderscore:
My bad then. Gotcha 👍🏼

Exiled.API/Features/FriendlyFireStatusCode.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Adding specific role failed, unable to be added.
+        /// </summary>
+        UnableToAdd,
+
+        /// <summary>
+        /// Current role successfully added
+        /// </summary>
+        AddedRole,
```
Thundermaker300, line 23: Rename to "Success"

Exiled.API/Features/FriendlyFireStatusCode.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Partial roles were added, cannot guarantee all.
+        /// </summary>
+        AddedSomeRoles,
```
Thundermaker300, line 28: Rename to "PartialSuccess" or something of the sort (AddedSomeRoles doesn't feel formal)

Exiled.API/Features/FriendlyFireStatusCode.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// All roles were successfully added
+        /// </summary>
+        AddedAllRoles,
```
Thundermaker300, line 33: Not necessary - just use "Success" instead. Documentation can just be "All roles were successfully added"

Exiled.API/Features/FriendlyFireStatusCode.cs
```diff
@@ -0,0 +1,35 @@
...
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Features
+{
+    /// <summary>
+    /// Status code for adding FriendlyFire rules.
+    /// </summary>
```
Thundermaker300, line 12: Having some <seealso> tags that link to the API would be cool, and generate nicely on the website

Exiled.API/Features/FriendlyFireStatusCode.cs
Undid-Iridium: Done

Exiled.API/Features/FriendlyFireStatusCode.cs
Undid-Iridium: Done

Exiled.API/Features/FriendlyFireStatusCode.cs
Undid-Iridium: Done, thanks

Exiled.API/Features/FriendlyFireStatusCode.cs
Undid-Iridium: Does not seem like I can because they're in different modules
image

Thundermaker300: Seealso tags go outside/after summary tags

Undid-Iridium: I don't think you understand - it does not matter if I put it there, or after the summary tags. Exiled.API is a dependency that other modules build upon, making it a transitive dependency at the Exiled.Events level. To solve this, I would have to have Exiled.API import from Exiled.Events; however, that would create a cyclic dependency which would not work. The dependency inversion principle makes it clear that a higher level class should not depend on a lower level class; therefore, the correct solution would be to have both depend on an abstraction of the co-shared data (Class, Functions, etc). This would require a significant amount of rework to allow interface abstractions. Therefore, this is not possible at the current time.

See:

image

IRacle1: First:
image

IRacle1: Second: it shouldn't be placed in `API.Features`

Undid-Iridium: > Second: it shouldn't be placed in `API.Features`

Where should it be placed?

Undid-Iridium: > First: image (If that's what you mean)

Ok, Apparently I am brain dead lmao

Undid-Iridium: Using my brain, moved to Exiled.Enums

Undid-Iridium: @Thundermaker300 Sorry for my incorrect statements, I assume you want the see also to each enum xml doc, or to the enum parent doc only?

Undid-Iridium: For now I've only added the seealso to the parent doc, let me know if that's what you wanted

Jesus-QC (changes requested):
Resolve conflicts

Undid-Iridium:
> Resolve conflicts

Done

Undid-Iridium:
Closing due to inactivity from reviewers. If someone wants this, re-open or create new PR.

PR #1314, opened 2022-07-06
Fix item sync

steven4547466:
Attachment setting/removing currently does not sync automatically when using the firearm abstracted methods. This should change that, it also properly disables flashlight flags on flashlight removal.


steven4547466:
Idk what the hell the other commits are for... As far as I can see they affect nothing. Github moment?

Jesus-QC (approved):

PR #1315, opened 2022-07-06
Fix null error when player is null

louis1706:
- Fix https://discord.com/channels/656673194693885975/664263236048388097/994301452434559089

### DIFF
--- a/Exiled.Events/EventArgs/DamagingShootingTargetEventArgs.cs
+++ b/Exiled.Events/EventArgs/DamagingShootingTargetEventArgs.cs
@@ -38,7 +38,7 @@ namespace Exiled.Events.EventArgs
             Amount = damage;
             Distance = distance;
             ShootingTarget = ShootingTargetToy.Get(shootingTarget);
-            Item = player.CurrentItem;
+            Item = player?.CurrentItem;
             DamageHandler = (AttackerDamageHandler)damageHandler;
             HitLocation = hitLocation;
             IsAllowed = isAllowed;


Thundermaker300 (approved):

IRacle1:
If this event doesn't trigger only from a shot, what's the point of the `Item` property anyway? 
After all, you would think that in the case of an exploding grenade, the `Item` property would contain a grenade? But it is not an item, it is Pickup, in general, I would delete it, this property raises more questions for me.
People can guess by themselves that if the damage came from the shot, they need to take the player's current item.

louis1706:
> If this event doesn't trigger only from a shot, what's the point of the `Item` property anyway? 
> After all, you would think that in the case of an exploding grenade, the `Item` property would contain a grenade? But it is not an item, it is Pickup, in general, I would delete it, this property raises more questions for me.
> People can guess by themselves that if the damage came from the shot, they need to take the player's current item.

Idk who adding this i didn't look it but if you change it it's would be a breaking change and yes i thinks it's not necessary

IRacle1:
I would add an obsolete, with a note on how it really works (takes the item in the player's hands)

louis1706:
> I would add an obsolete, with a note on how it really works (takes the item in the player's hands)

Okay i will add this later and deleted it in exiled 6

IRacle1:
yea, thanks

zhaguanyang (approved):

Jesus-QC (approved):

Exiled.Events/EventArgs/DamagingShootingTargetEventArgs.cs
```diff
@@ -57,7 +56,7 @@ public DamagingShootingTargetEventArgs(Player player, float damage, float distan
         /// <summary>
         /// Gets the item which is being used to deal the damage.
         /// </summary>
-        public Item Item { get; }
+        public Item Item => Player?.CurrentItem;
```
Jesus-QC, line 59: Shouldnt it just be = and not =>

louis1706:
> Please, change that

I have changed this because nao requested it

Jesus-QC:
> > Please, change that
> 
> I have changed this because nao requested it
> 
> Before i have wan't to make it obsolete because it's not necessary Look how it's was before nao requests please i would prefer it like that ^^ if it's okay for you

Where is Nao request?

zhaguanyang:
> Where is Nao request?

https://discord.com/channels/656673194693885975/668651891944587264/995785753168715956

louis1706:
> > > Please, change that
> > 
> > I have changed this because nao requested it
> > 
> > Before i have wan't to make it obsolete because it's not necessary Look how it's was before nao requests please i would prefer it like that ^^ if it's okay for you
> 
> Where is Nao request?

You have a commit called nao request you can look what is deleted

Jesus-QC (approved):

PR #1317, opened 2022-07-07
InteractingDoor event with KeycardThrow

IRacle1:
~~The only problem is that the interaction with the collision occurs several times, and the event is also invoked several times (an average of 3 times), but I do not think that this is a big problem~~
fixed

### DIFF
--- a/Exiled.Events/Config.cs
+++ b/Exiled.Events/Config.cs
@@ -57,8 +57,8 @@ namespace Exiled.Events
         /// <summary>
         /// Gets or sets a value indicating whether keycard throw can affect basic doors.
         /// </summary>
-        [Description("Indicates whether keycard throw can affect on basic doors(door with empty permissions)")]
-        public bool CanKeycardThrowAffectOnBasicDoors { get; set; } = true;
+        [Description("Indicates whether thrown keycards can affect doors that don't require any permissions")]
+        public bool CanKeycardThrowAffectDoors { get; set; } = true;
 
         /// <summary>
         /// Gets or sets a value indicating whether configs has to be reloaded every time a round restarts.
--- a/Exiled.Events/Config.cs
+++ b/Exiled.Events/Config.cs
@@ -55,6 +55,12 @@ namespace Exiled.Events
         public bool CanSpawnBlood { get; set; } = true;
 
         /// <summary>
+        /// Gets or sets a value indicating whether keycard throw can affect basic doors.
+        /// </summary>
+        [Description("Indicates whether keycard throw can affect on basic doors(door with empty permissions)")]
+        public bool CanKeycardThrowAffectOnBasicDoors { get; set; } = true;
+
+        /// <summary>
         /// Gets or sets a value indicating whether configs has to be reloaded every time a round restarts.
         /// </summary>
         [Description("Indicates whether configs have to be reloaded every round restart")]
--- a/Exiled.Events/EventArgs/ThrowKeycardInteractingEventArgs.cs
+++ b/Exiled.Events/EventArgs/ThrowKeycardInteractingEventArgs.cs
@@ -42,7 +42,7 @@ namespace Exiled.Events.EventArgs
         public Pickup Pickup { get; }
 
         /// <summary>
-        /// Gets the player who's interacting with the door.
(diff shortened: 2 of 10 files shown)


louis1706:
I can look for make it not trigger multiple time

louis1706:
> I can imagine the "trigger multiple times" would be a problem for plugins such as discord game logging plugins, which would send a log multiple times.

I already know how fixed it ^^ do not worry

louis1706 (approved):
it's all fine

Exiled.Events/EventArgs/ThrowKeycardInteractingEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+    using Exiled.API.Features.Items;
+
+    using Interactables.Interobjects.DoorUtils;
+
+    using InventorySystem.Items.Keycards;
+
+    /// <summary>
+    /// Contains all informations before a keycard interacts with a door.
+    /// </summary>
+    public class ThrowKeycardInteractingEventArgs : EventArgs
```
NaoUnderscore, line 22: I'd rename it to `ThrowingInteractableKeycard` or `KeycardInteracting`.

Exiled.Events/Patches/Events/Item/InteractingDoorWithThrow.cs
```diff
@@ -0,0 +1,154 @@
...
+    using UnityEngine;
+
+    using static HarmonyLib.AccessTools;
+
+    /// <summary>
+    /// Patches <see cref="KeycardPickup.ProcessCollision(Collision)"/>.
+    /// Adds the <see cref="Handlers.Player.InteractingDoor"/> event.
+    /// </summary>
+    [HarmonyPatch(typeof(KeycardPickup), nameof(KeycardPickup.ProcessCollision))]
+    internal static class InteractingDoorWithThrow
```
NaoUnderscore, line 37: Keep the same name as the event if possible.

Exiled.Events/Config.cs
```diff
@@ -54,6 +54,12 @@ public sealed class Config : IConfig
         [Description("Indicates whether the blood can be spawned or not")]
         public bool CanSpawnBlood { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether keycard throw can affect basic doors.
+        /// </summary>
+        [Description("Indicates whether keycard throw can affect on basic doors(door with empty permissions)")]
+        public bool CanKeycardThrowAffectOnBasicDoors { get; set; } = true;
```
NaoUnderscore, line 61: `CanKeycardThrowAffectDoors`

Exiled.Events/Config.cs
```diff
@@ -54,6 +54,12 @@ public sealed class Config : IConfig
         [Description("Indicates whether the blood can be spawned or not")]
         public bool CanSpawnBlood { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether keycard throw can affect basic doors.
+        /// </summary>
+        [Description("Indicates whether keycard throw can affect on basic doors(door with empty permissions)")]
```
NaoUnderscore, line 60: I'd change it to: _Indicates whether thrown keycards can affect doors that don't require any permissions_

NaoUnderscore (changes requested):

BuildBoy12:
While it does fire only once when the interaction is allowed, it still fires multiple times when it is not.

LczArmory Door:
image

Checkpoint:
image

IRacle1:
my problem is logic, I'll fix it tomorrow

louis1706:
> While it does fire only once when the interaction is allowed, it still fires multiple times when it is not.
> 
> LczArmory Door:
> image
> 
> Checkpoint:
> image
> 

Is not in base game already like that i don't see the problem with this ? It's pretty normal i thinks

IRacle1:
It's still a duplicate event, lots of reasons why it's bad

louis1706:
> It's still a duplicate event, lots of reasons why it's bad

I will try something later for see if it's do not break something

Jesus-QC (approved):

PR #1326, opened 2022-07-14
Call finishing recall earlier

rayzerbrain:
This calls OnFinishingRecall even if the person being recalled isn't a spectator, and it just sets IsAllowed to false if this is the case.

### DIFF
--- a/Exiled.Events/Patches/Events/Scp049/StartingAndFinishingRecall.cs
+++ b/Exiled.Events/Patches/Events/Scp049/StartingAndFinishingRecall.cs
@@ -35,7 +35,7 @@ namespace Exiled.Events.Patches.Events.Scp049
         {
             List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Shared.Rent(instructions);
 
-            int offset = -4;
+            const int offset = -4;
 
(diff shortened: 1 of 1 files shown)


louis1706:
I really like thinks like that :)

Exiled.Events/Patches/Events/Scp049/StartingAndFinishingRecall.cs
```diff
@@ -60,23 +62,28 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
                 new(OpCodes.Brfalse_S, ret),
             });
 
-            offset = -2;
-            index = newInstructions.FindLastIndex(instruction => instruction.opcode == OpCodes.Ldc_I4_S) + offset;
+            offset = 0;
```
BuildBoy12, line 65: If the required offset is 0 then `offset` can be set to a `const` of -4 and this can be removed.

BuildBoy12 (changes requested):

rayzerbrain:
Done

BuildBoy12 (approved):

BuildBoy12 (changes requested):
Fix build errors.

PR #1331, opened 2022-07-18
EnvironmentalHazard fixes

IRacle1:
That's will fix `ev.IsAllowed` ignore


IRacle1:
Dont megre until i commit `OnExit` fix

IRacle1:
~Ctrl C Ctrl V from exiled-dev-talk~
1. Now `OnEnter` in sinkhole class enables effect permanently, and removing only in `OnExit`(instead of stupid `EnableEffect` in `OnStay` ,  I remove that)
2. Now, if I`sAllowed == false` in Sinkhole class in `OnExit` , RefHub will ALWAYS be removed from `AffectedPlayers`, `IsAllowed` will now only affect the removing effect(If RefHub is not removed from this list, `OnExit` will be called every frame, attempting to remove)
3. `IsAllowed` in `OnStay` event removed(I wrote before why)

IRacle1:
@NaoUnderscore , in theory, that all, I finished everything I wanted to do.
I've already checked everything a 1000 times, but it would be nice if @louis1706 and @Undid-Iridium also checked it

Undid-Iridium:
> @NaoUnderscore , in theory, that all, I finished everything I wanted to do. I've already checked everything a 1000 times, but it would be nice if @louis1706 and @Undid-Iridium also check it

I'll take a look later

Exiled.Events/Patches/Events/Player/ExitingSinkholeEnvironmentalHazard.cs
```diff
@@ -0,0 +1,85 @@
...
+                new(OpCodes.Brtrue_S, cnt),
+
+                // exit effect for 1 second, cause we disable effect in OnStay
+                new(OpCodes.Ldarg_1),
+                new(OpCodes.Ldfld, Field(typeof(ReferenceHub), nameof(ReferenceHub.playerEffectsController))),
+                new(OpCodes.Ldc_R4, 1f),
+                new(OpCodes.Ldc_I4_0),
+                new(OpCodes.Callvirt, Method(typeof(PlayerEffectsController), nameof(PlayerEffectsController.EnableEffect), new[] { typeof(float), typeof(bool) }, new[] { typeof(SinkHole) })),
+
+                new CodeInstruction(OpCodes.Nop).WithLabels(cnt),
```
Undid-Iridium, line 76: How about? (I assume the last instruction is a ret)

newInstructions[newInstructions.Count - 1].withLabels(cnt)

NaoUnderscore: Yeah I  agree, `Nop` isn't needed.

Exiled.Events/EventArgs/WalkingOnTantrumEventArgs.cs
```diff
@@ -14,6 +14,7 @@ namespace Exiled.Events.EventArgs
     /// <summary>
     /// Contains all information before a player walks over a tantrum.
     /// </summary>
+    [Obsolete("Use StayingOnEnvironmentalHazardEventArgs event instead.")]
```
NaoUnderscore, line 17: Same as above.

Exiled.Events/Handlers/Player.cs
```diff
@@ -828,6 +828,7 @@ public static class Player
         /// Called before a <see cref="API.Features.Player"/> walks on a sinkhole.
         /// </summary>
         /// /// <param name="ev">The <see cref="WalkingOnSinkholeEventArgs"/> instance.</param>
+        [Obsolete("Use OnStayingOnEnvironmentalHazard instead.")]
```
NaoUnderscore, line 831: Same as above.

Exiled.Events/Handlers/Player.cs
```diff
@@ -882,6 +883,7 @@ public static class Player
         /// Called before a <see cref="API.Features.Player"/> walks on a tantrum.
         /// </summary>
         /// /// <param name="ev">The <see cref="WalkingOnTantrumEventArgs"/> instance.</param>
+        [Obsolete("Use OnStayingOnEnvironmentalHazard instead.")]
```
NaoUnderscore, line 886: Same as above.

Exiled.Events/EventArgs/WalkingOnSinkholeEventArgs.cs
```diff
@@ -14,6 +14,7 @@ namespace Exiled.Events.EventArgs
     /// <summary>
     /// Contains all information before a player walks over a sinkhole.
     /// </summary>
+    [Obsolete("Use StayingOnEnvironmentalHazardEventArgs event instead.")]
```
NaoUnderscore, line 17: Use `ObsoleteAttribute(string, bool)` where `bool : true`.

Exiled.Events/Patches/Events/Player/ExitingSinkholeEnvironmentalHazard.cs
IRacle1: There was a little different logic here, and Nop was needed, but yes, it is not needed now, I will remove it tomorrow

Exiled.Events/EventArgs/StayingOnEnvironmentalHazardEventArgs.cs
```diff
@@ -42,6 +40,7 @@ public StayingOnEnvironmentalHazardEventArgs(Player player, EnvironmentalHazard
         /// <summary>
         /// Gets or sets a value indicating whether or not the player should be affected by the environmental hazard.
         /// </summary>
+        [Obsolete("Not used(I'm too lazy to come up with a clever message here :( )")]
```
NaoUnderscore, line 43: Use `ObsoleteAttribute(string, bool)` where `string : "IsAllowed has been deprecated, use EnteringEnvironmentalHazardEventArgs::IsAllowed instead."` & `bool : true`.

NaoUnderscore (changes requested):
Remove the old patches too, if any.

Exiled.Events/Patches/Fixes/SinkholeEffectFix.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            newInstructions.InsertRange(0, new[]
+            {
+                new CodeInstruction(OpCodes.Ldarg_0),
+                new(OpCodes.Ldarg_1),
+                new(OpCodes.Call, Method(typeof(EnvironmentalHazard), nameof(EnvironmentalHazard.OnEnter))),
+            });
+
+            int offset = -2;
+            int index = newInstructions.FindLastIndex(i => i.Calls(Method(typeof(EnvironmentalHazard), nameof(EnvironmentalHazard.OnEnter)))) + offset;
```
Undid-Iridium, line 41: Just to be clear, as I read this, you're moving the OnEnter to be called immediately, and deleting the original OnEnter later on. Further, you also are not enabling the SinkHole effect OnEnter anymore. Is this an accurate understanding?

IRacle1: You're right about `OnEnter` moving, but i still enable effect, I just set duration to 0, cause thats infinity duration

Undid-Iridium: Gotcha, thanks for clarifying

Undid-Iridium (approved):

IRacle1:
Thx❤

NaoUnderscore (approved):

Jesus-QC (approved):

PR #1332, opened 2022-07-18
[Breaking change] Fixes `uint/int` casts

IRacle1:


joker-119:
There's no reason for the uint -> int changes to custom items and roles.

Thundermaker300 (approved):

PR #1339, opened 2022-07-26
Item Get(serial) and List

louis1706:
- Add Item::List
- Add Item::Get(ushort Serial)
- Add Pickup::List
- Add Pickup::Get(ushort Serial)


louis1706:
I going to merged this In Pickup API branch

PR #1138, opened 2022-03-11
Shouldn't it be Ended

Jesus-QC:

PR #1340, opened 2022-07-27
Fix Round.IsStarted sense, added Round.InProgress

RunicLib:
image
image

Old `IsStarted` have a wrong sense. That property indicated `InProgress` status, instead of `Started` status.
Maybe that change isn't necessary, but sense of `IsStarted` not compare with `InProgress`, imho.

### DIFF
--- a/Exiled.API/Features/Round.cs
+++ b/Exiled.API/Features/Round.cs
@@ -37,7 +37,12 @@ namespace Exiled.API.Features
         /// <summary>
         /// Gets a value indicating whether the round is started or not.
         /// </summary>
-        public static bool IsStarted => ReferenceHub.LocalHub is not null && RoundSummary.RoundInProgress();
+        public static bool IsStarted => ReferenceHub.LocalHub?.characterClassManager.RoundStarted ?? false;
+
+        /// <summary>
+        /// Gets a value indicating whether the round in progress or not.
+        /// </summary>
+        public static bool InProgress => ReferenceHub.LocalHub is not null && RoundSummary.RoundInProgress();
 
         /// <summary>
         /// Gets a value indicating whether the round is ended or not.


Soft0101 (approved):

louis1706:
undirectly Breaking change

louis1706:
https://github.com/Exiled-Team/EXILED/pull/1138 it"s look same at this

RunicLib:
> undirectly Breaking change

yes, that change is breaking
but, original sense of `IsStarted` is breaking some systems too

RunicLib:
> #1138 it"s look same at the comment in this PR

that's funny, yea

louis1706:
> > undirectly Breaking change
> 
> yes, that change is breaking but, original sense of `IsStarted` is breaking some systems too

I rembere already discuss about this but i don't remember all the discussion

RunicLib:
> > > undirectly Breaking change
> > 
> > 
> > yes, that change is breaking but, original sense of `IsStarted` is breaking some systems too
> 
> I rembere already discuss about this but i don't remember all the discussion

As I see, our Pull Requests are little different. I have nothing to say, our requests are really different, or I not understand what you want to convey

IRacle1:
The point is that it changes the logic of the property, it will not return the same value as before

RunicLib:
> The point is that it changes the logic of the property, it will not return the same value as before

no care, mistakes are mistakes, but current logic of that property is wrong

RunicLib:
ok, I remembered what property `IsStarted` is being used in some EXILED
methods. I will change it to new and commit soon.

чт, 28 июл. 2022 г., 21:19 IRacle ***@***.***>:

RunicLib:
As I checked, new property value is not need to be changed in EXILED methods.
But, yea, this is breaking change, and if this pull request will be accepted, we need to notify dev's about this change.

Jesus-QC (approved):

NaoUnderscore:
If that's a breaking change, and we're gonna keep it for the next major update, we can just close it right now as NW is going to rework the `CharacterClassManager`. Aside from that, why do you mark it as breaking change? The signature is still the same 🤔

RunicLib:
> If that's a breaking change, and we're gonna keep it for the next major update, we can just close it right now as NW is going to rework the `CharacterClassManager`. Aside from that, why do you mark it as breaking change? The signature is still the same 🤔
 
I marked that as breaking change because value of the property is been changed. It means what property will not return old value

NaoUnderscore:
It doesn't really affect code logic though. If the round is in progress, it implicitly means the round is started.

RunicLib:
> It doesn't really affect code logic though. If the round is in progress, it implicitly means the round is started.

ok, so I will remove the breaking tag

No738 (approved):
lgtm

PR #1342, opened 2022-07-28
Fix component not being reset

louis1706:
- Thanks ArseeniK to report it

### DIFF
--- a/Exiled.API/Features/Scp914.cs
+++ b/Exiled.API/Features/Scp914.cs
@@ -19,8 +19,6 @@ namespace Exiled.API.Features
     /// </summary>
     public static class Scp914
     {
-        private static Scp914Controller scp914Controller;
-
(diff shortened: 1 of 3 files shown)


IRacle1:
I think this is unnecessary here, in my opinion the problem is in the conditional check for null, if you write it with a check gameObject to null, I think there will be no problem

IRacle1:
Perhaps this will solve all the problems of the cache

IRacle1:
`scp914Controller ? scp914Controller : scp914Controller = Object::FindObjectOfType`

louis1706:
> `scp914Controller ? scp914Controller : scp914Controller = Object::FindObjectOfType`

Ir's the than `scp914Controller ??= Object::FindObjectOfType`

Jesus-QC (approved):

PR #1349, opened 2022-08-01
Player events

Ficus-x:


louis1706:
VerifiedEventArgs and DestroyingEventArgs

rus-skill-issue:
Bruh moment pull

IRacle1:
I'm 90% sure that `IsAllowed = false` in these events will lead to things like server crash, I also want to know why they are needed at all

IRacle1:
As Yamato said, `Verified` and `Destroying` events exist

PR #1352, opened 2022-08-01
New overload for Player::Heal

Ficus-x:


Exiled.API/Features/Player.cs
```diff
@@ -1912,6 +1897,36 @@ public void Heal(float amount, bool overrideMaxHealth = false)
                 Health += amount;
         }
 
+        /// <summary>
+        /// Heals the player.
```
Thundermaker300, line 1901: "Heals the player by applying an item effect."?

Thundermaker300 (approved):

Exiled.API/Features/Player.cs
```diff
@@ -14,49 +14,34 @@ namespace Exiled.API.Features
...
     using PlayableScps;
-
     using PlayerStatsSystem;
-
     using RemoteAdmin;
-
     using RoundRestarting;
-
     using UnityEngine;
-
```
PintTheDragon, line 59: Please don't change the using style.

Ficus-x: Solved

Thundermaker300:
My only concern with this is Northwood changing item effect = method being outdated. I would recommend closing this, as #1355 does the same thing (+ with more items) and accounts for northwood code changes.

PR #1355, opened 2022-08-03
UseItem

louis1706:
- instead of https://github.com/Exiled-Team/EXILED/pull/1352

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1916,7 +1916,16 @@ namespace Exiled.API.Features
         /// Used an item for giving effect of it.
         /// </summary>
         /// <param name="usableItem">The item to be used.</param>
-        public void UseItem(ItemType usableItem) => (Item.Create(usableItem, this) as Usable)?.Base.ServerOnUsingCompleted();
+        public void UseItem(ItemType usableItem)
+        {
+            if (Item.Create(usableItem, this) is not Usable item)
+                throw new Exception($"The provided item [{usableItem}] is not a usable item.");
+
+            item.Base.Owner = referenceHub;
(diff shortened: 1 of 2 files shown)


louis1706:
Need test for all the folowiing itemtype
ItemType.Scp500 (or any other consumable item)
ItemType.Scp330
ItemType.Scp244
ItemType.Scp268

Exiled.API/Features/Player.cs
```diff
@@ -1912,6 +1912,21 @@ public void Heal(float amount, bool overrideMaxHealth = false)
                 Health += amount;
         }
 
+        /// <summary>
+        /// Used an item for giving effect of it.
```
Ficus-x, line 1916: Maybe "Uses an item for giving its effect."? I think the method name should be also Heal

louis1706: no because it's can use other things who are not heal like Scp244

Jesus-QC (approved):

Exiled.API/Features/Player.cs
```diff
@@ -1912,6 +1912,21 @@ public void Heal(float amount, bool overrideMaxHealth = false)
...
         }
 
+        /// <summary>
+        /// Uses an item for giving its effect.
+        /// </summary>
+        /// <param name="usableItem">The item to be used.</param>
+        public void UseItem(ItemType usableItem)
+        {
+            if (Item.Create(usableItem, this) is not Usable item)
+                throw new Exception($"The provided item [{usableItem}] is not a usable item.");
```
Ficus-x, line 1922: Maybe throw **new InvalidOperationException**? Also "The provided item [{usableItem}] is not a**n** usalbe item."

PR #1360, opened 2022-08-05
EnteringKillerCollision event

Ficus-x:


Monaldcry7788:
Idk if devs will accept this, it's better do a transpiler for the patch.

Thundermaker300:
> Idk if devs will accept this, it's better do a transpiler for the patch.

iirc they accept non-transpiler patches and convert them to transpilers later on. may be wrong though

louis1706 (approved):

IRacle1 (approved):

rus-skill-issue (approved):

PintTheDragon:
What does this allow that the Killing event doesn't?

Ficus-x:
Killing event does not fire if the player has God mode

PintTheDragon:
> Killing event does not fire if the player has God mode

So the killing by collision event will fire when a player won't actually be killed by collision?

Ficus-x:
I understood what you mean that is why i renamed this event

Jesus-QC (approved):

PR #1364, opened 2022-08-06
Clone methods

Thundermaker300:


Ficus-x:
Does cline methods make sense? 🤔

Thundermaker300:
> Does cline methods make sense? 🤔

very simple API yes, but I'm sure someone somewhere out there could use it lol

PintTheDragon:
> > Does cline methods make sense? thinking
> 
> very simple API yes, but I'm sure someone somewhere out there could use it lol

We shouldn't add things just to add things. I'm not completely sure what these would be used for.

Thundermaker300:
> > > Does cline methods make sense? thinking
> > 
> > 
> > very simple API yes, but I'm sure someone somewhere out there could use it lol
> 
> We shouldn't add things just to add things. I'm not completely sure what these would be used for.

Fair point

PR #1367, opened 2022-08-08
Add item, spawnpoint, and enums to Player::Teleport(object)

Thundermaker300:


Michal78900 (approved):

joker-119:
Items are already handled by Pickup, since items in a player's inventory may/may not have their transforms set properly, if you want to teleport to an item (which is only ever in a player's inventory) teleport to the item's owner, otherwise use Pickup instead.

Thundermaker300:
> Items are already handled by Pickup, since items in a player's inventory may/may not have their transforms set properly, if you want to teleport to an item (which is only ever in a player's inventory) teleport to the item's owner, otherwise use Pickup instead.

Done (I think)

Jesus-QC (approved):

PR #1372, opened 2022-08-10
`OnTogglingOverwatch`

Jesus-QC:
Adds new event.


Exiled.Events/EventArgs/TogglingOverwatchEventArgs.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Gets the player that is toggling overwatch.
+        /// </summary>
+        public Player Player { get; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether overwatch will be enabled or not.
+        /// </summary>
+        public bool NewValue { get; set; }
```
NaoUnderscore, line 37: Rename it to `IsEnabled` instead.

NaoUnderscore (changes requested):

PR #1374, opened 2022-08-13
Scp106Role.Contain null check

Thundermaker300:


IRacle1:
Why? Contain Server::Host player dont will lead to errors?

Exiled.API/Features/Roles/Scp106Role.cs
```diff
@@ -78,6 +78,6 @@ public float CaptureCooldown
         /// Contains SCP-106.
         /// </summary>
         /// <param name="container">The player who recontained SCP-106.</param>
-        public void Contain(Player container) => Script.Contain(container.Footprint);
+        public void Contain(Player container) => Script.Contain(container?.Footprint ?? Server.Host.Footprint);
```
Jesus-QC, line 81: Instead of giving it the servers hosts footprint, add a proper null check to avoid code execution + avoid weird issues

Jesus-QC (changes requested):

rus-skill-issue (approved):

Exiled.API/Features/Roles/Scp106Role.cs
Thundermaker300: If it's null, throw an exception or something?

Jesus-QC (approved):

PintTheDragon (approved):
I'm going to hold off on merging this (for 5.3.0-rc.1) because it isn't quite a bug fix, although it isn't really an addition either.

Thundermaker300:
> I'm going to hold off on merging this (for 5.3.0-rc.1) because it isn't quite a bug fix, although it isn't really an addition either.

ok

PR #1378, opened 2022-08-14
Make Side.None the default value

rayzerbrain:


PintTheDragon:
This is a breaking change due to the ways enums work. Any code references to Side.Tutorial will become Side.None, and vice versa.

rayzerbrain:
You're right, didn't think about that. Probably not worth it then

PR #1386, opened 2022-08-17
Metrics

joker-119:
Keeps track of various information about servers.
The endpoint will be an asp.net app that will keep records of various information about each server and allow the public to search through it to identify trends, such as the most popular plugins, avg tps of servers, the team that wins the most rounds, the avg player count/max player count of servers, etc.

More data can be added onto these metrics in the future.

All of this is with server anonymity in mind. Since you need the Server IP and Port hashed together before correlating a server to it's data, it will not be possible to identify from the database alone what servers have what stats, except via a command in the metrics plugin that I will add at a future time.

I will likely replace this server hash identifier with a unique api key stored in the plugin config, that way only those with access to the api key may be able to correlate the data at all (in it's current form, you could hash a server's ip and port yourself in the same fashion used here, to get it's stats)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,6 +134,7 @@ $tf/
 _ReSharper*/
 *.[Rr]e[Ss]harper
 *.DotSettings.user
+*.DotSettings
 
 # TeamCity is a build add-in
 _TeamCity*
--- a/EXILED.sln
+++ b/EXILED.sln
@@ -27,6 +27,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Exiled.CreditTags", "Exiled
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Exiled.CustomRoles", "Exiled.CustomRoles\Exiled.CustomRoles.csproj", "{417C3309-8B93-4218-A1D1-D4BB7B09BE0F}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Exiled.Metrics", "Exiled.Metrics\Exiled.Metrics.csproj", "{4F73D1A2-3101-4785-9564-3ECAEDB1AFD1}"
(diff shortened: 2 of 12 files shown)


PintTheDragon:
What happens if I send a request without a server? What if I pretend to be a hosting platform and send data bound to hundreds of ports? Is there any mechanism to make sure that a server actually exists? Is there any mechanism to make sure a server has sent a request in the past few days (this should be fairly simple, adding it here just in case)?

These are all questions to keep in mind with a system like this where there is clear gain to exploiting it. Here is how SCPStats handles it, which also has the aforementioned issue (maybe weakness is a better word), specifically with people gaining false stats:
1. Verify that a server actually exists on the list. Everything is tied to the server's list ID, and if servers switch out their SCPStats ID, the stats on the old SCPStats ID get "unverified" and no longer count towards global stats.
2. Servers are verified through the server list. This means that the server needs to have a spot on the list, and it needs to have permission to edit it. I can go through the actual verification flow if you'd like, but it basically prevents a server from lying about who they are.
3. For server counting, which seems to be what you're trying to achieve, I only count servers which are verified (using the method above) and have verified themselves in the past 30 days. This prevents stale servers from being included and only counts "real" servers.

joker-119:
I know I need to think about security here, and this current system has pretty much none at the moment. The asp.net app that'll handle these requests can be found [Here](https://github.com/Exiled-Team/ExiledMetricsWeb) and also has basically no security on it's end yet. 

This is a very early draft, but I'd certainly be willing to accept PR's to either side of this system that'll help improve it, especially it's security and server verification process.

PintTheDragon:
Ok, I'm working on adding the server verification that Zabszk made for us. The plugin is done (so long as it works), and I'm currently working on the website.

PintTheDragon:
Everything other than testing and fixing bugs is now complete. There might be a few more things to look into like input validation, but it isn't super exploitable (we should definitely do it anyway so that, at the very least, the data isn't invalid).

PR #1387, opened 2022-08-17
Add Usable::Use(Player)

Thundermaker300:


NaoUnderscore (approved):

Exiled.API/Features/Items/Usable.cs
```diff
@@ -97,5 +97,18 @@ public float RemainingCooldown
             get => Base.RemainingCooldown;
             set => Base.RemainingCooldown = value;
         }
+
+        /// <summary>
+        /// Uses the item.
+        /// </summary>
+        /// <param name="owner">The player to use the item.</param>
+        /// <exception cref="System.ArgumentNullException">Owner cannot be a null value.</exception>
+        public virtual void Use(Player owner)
```
joker-119, line 106: Wouldn't this technically allow a plugin to use a usable item from someones inventory on another player?

If that's the case 'owner' should be renamed to 'target' or something like that.

If it always uses the item on whoever actually owns it instead (wouldn't make sense but NW money code), having the owner parameter here would be pointless.

joker-119 (changes requested):

Exiled.API/Features/Items/Usable.cs
NaoUnderscore: `Owner` refers to the player who's using the item. The `Owner` is defined in the exact moment in which the item is being used, claiming the `Owner` as the delegate user.

However, this is a method in the `Item` class, which treats the `Owner` as an abstract figure, and sometimes can just be dynamic over logical considerations rather than over computational resolution.

PR #1394, opened 2022-08-25
Few docs fixes

IRacle1:
- fixed readme CI link(I remember jesus already fixed it, lol why)
- fixed readme link to new website
- fixed /Installation/Intro page, I don't think there should be two partitions with the same name, and 'The old way.' text for manual install


IRacle1:
Jesus reworked site

PR #1398, opened 2022-08-28
[Breaking] Role properties

Thundermaker300:


IRacle1:
these props also exists in `Player` class, so need to fix the duplicate

Thundermaker300:
> these props also exists in `Player` class, so need to fix the duplicate

Oh right, I thought they were removed. Forgot to check lol

rus-skill-issue (approved):

joker-119:
I'm like 99.999% certain they were purposefully left in the player class for ease of use.
``ev.Player.IsNtf`` is easier to type/remember than ``ev.Player.Role.IsNtf``

Thundermaker300:
> I'm like 99.999% certain they were purposefully left in the player class for ease of use. `ev.Player.IsNtf` is easier to type/remember than `ev.Player.Role.IsNtf`

Ease of use yeah that makes sense, but it doesn't make sense logistically. yes, whether or not a player is an SCP is a player property (and hell, we can leave the properties on the player if we really want), they are also role information which should be available on the role class itself

NaoUnderscore (approved):

joker-119:
> > I'm like 99.999% certain they were purposefully left in the player class for ease of use. `ev.Player.IsNtf` is easier to type/remember than `ev.Player.Role.IsNtf`
> 
> Ease of use yeah that makes sense, but it doesn't make sense logistically. yes, whether or not a player is an SCP is a player property (and hell, we can leave the properties on the player if we really want), they are also role information which should be available on the role class itself

Reasonable, and honestly I agree with you and these changes.
I just wanted to point out these were deliberately put here by someone at some point.

joker-119:
Also fix brokey 2.0

Thundermaker300:
idk how to fix the transpiler

Michal78900 (approved):

Michal78900 (changes requested):
Fix build errors please

Thundermaker300:
I do not know how to transpiler

Thundermaker300:
Resigned

PR #1402, opened 2022-09-03
New Player property `KeepInvisibilityOnInteracting`.

microsievert:
This property prevents game from removing invisibility effect for player that interacted with something. Added new property and new patch.


Ficus-x (approved):

NaoUnderscore:
I don't really find it so useful, as you can use the relative event to deny the original code behavior.
This applies to any properties which replicate the denial behavior of an event.

NaoUnderscore:
As I'm not able to find my previous way to achieve the same result using Exiled API only, I'm gonna consider it as a valid API addition as people may want to prevent players from losing their invisibility status effect regardless their interacting actions.

Exiled.API/Features/Player.cs
```diff
@@ -1083,6 +1083,11 @@ public bool IsInPocketDimension
         /// </summary>
         public bool IsUsingStamina { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether or not the player invisibility effect should not be removed when interacting.
+        /// </summary>
+        public bool InvisibilityInteractionProtect { get; set; } = false;
```
NaoUnderscore, line 1089: `KeepInvisibilityOnInteracting` should be better or something.

NaoUnderscore (changes requested):

Exiled.API/Features/Player.cs
microsievert: Solved

Exiled.API/Features/Player.cs
```diff
@@ -1083,6 +1083,11 @@ public bool IsInPocketDimension
         /// </summary>
         public bool IsUsingStamina { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether or not the player invisibility effect should not be removed when interacting.
+        /// </summary>
+        public bool KeepInvisibilityOnInteracting { get; set; } = false;
```
NaoUnderscore, line 1089: Just `Gets or sets a value indicating whether the invisibility effect should be removed when interacting.`

NaoUnderscore, line 1089: `= false` is not needed.

Exiled.Events/Patches/Generic/InvisibleInteractionPatch.cs
```diff
@@ -0,0 +1,107 @@
...
+            for (int z = 0; z < newInstructions.Count; z++)
+                yield return newInstructions[z];
+
+            ListPool<CodeInstruction>.Shared.Return(newInstructions);
+        }
+    }
+
+    /// <summary>
+    /// Patches <see cref="PlayerInteract.OnInteract"/>.
+    /// Implements <see cref="Player.KeepInvisibilityOnInteracting"/> property logic.
```
NaoUnderscore, line 71: Same as above.

Exiled.Events/Patches/Generic/InvisibleInteractionPatch.cs
```diff
@@ -0,0 +1,107 @@
...
+    using Interactables;
+    using Interactables.Verification;
+
+    using NorthwoodLib.Pools;
+
+    using static HarmonyLib.AccessTools;
+
+    /// <summary>
+    /// Patches <see cref="StandardDistanceVerification.ServerCanInteract(ReferenceHub, InteractableCollider)"/>.
+    /// Implements <see cref="Player.KeepInvisibilityOnInteracting"/> property logic.
```
NaoUnderscore, line 30: Remove `property logic.`

NaoUnderscore (changes requested):

Exiled.Events/Patches/Generic/InvisibleInteractionPatch.cs
```diff
@@ -0,0 +1,107 @@
...
+    using NorthwoodLib.Pools;
+
+    using static HarmonyLib.AccessTools;
+
+    /// <summary>
+    /// Patches <see cref="StandardDistanceVerification.ServerCanInteract(ReferenceHub, InteractableCollider)"/>.
+    /// Implements <see cref="Player.KeepInvisibilityOnInteracting"/>.
+    /// </summary>
+    [HarmonyPatch(typeof(StandardDistanceVerification), nameof(StandardDistanceVerification.ServerCanInteract))]
+    internal static class InvisibleInteractionPatch
```
NaoUnderscore, line 33: Rename it as `ServerCanInteractPatch`.

Exiled.Events/Patches/Generic/InvisibleInteractionPatch.cs
```diff
@@ -0,0 +1,107 @@
...
+            ListPool<CodeInstruction>.Shared.Return(newInstructions);
+        }
+    }
+
+    /// <summary>
+    /// Patches <see cref="PlayerInteract.OnInteract"/>.
+    /// Implements <see cref="Player.KeepInvisibilityOnInteracting"/>.
+    /// </summary>
+    [HarmonyPatch(typeof(PlayerInteract), nameof(PlayerInteract.OnInteract))]
+    internal static class InvisibleInteractionPlayerInteractPatch
```
NaoUnderscore, line 74: Rename it as `PlayerInteractPatch`.

NaoUnderscore (changes requested):

PR #1409, opened 2022-09-12
Warhead EventArgs renamed

ClaudioPanConQueso:


ClaudioPanConQueso:
For exiled 6

louis1706 (approved):

joker-119:
Since these are already in the Warhead namespace, is this even necessary? 
afaik there aren't even other events that share the same names, right?

Exiled.Events/Handlers/Scp079.cs
```diff
@@ -143,7 +143,7 @@ public static void OnStartingSpeaker(StartingSpeakerEventArgs ev)
         /// <summary>
         ///     Called before SCP-079 finishes using a speaker.
         /// </summary>
-        /// <param name="ev">The <see cref="StoppingEventArgs" /> instance.</param>
+        /// <param name="ev">The <see cref="StoppingWarheadEventArgs" /> instance.</param>
         public static void OnStoppingSpeaker(StoppingSpeakerEventArgs ev)
```
IRacle1, line 147: xD, fix it

IRacle1 (changes requested):

IRacle1:
> Since these are already in the Warhead namespace, is this even necessary? 
> afaik there aren't even other events that share the same names, right?

Yes, but most of people search events by EventArg name, it will be easier for them and for understanding that event + also many event are named with StoppedXXX

ClaudioPanConQueso:
First time i was coding, i didn't know that StartingEventArgs and Stopping was from the warhead.

Exiled.Events/Handlers/Scp079.cs
ClaudioPanConQueso: What thing? XD

ClaudioPanConQueso: fixed

Thundermaker300:
It is very much necessary

Michal78900 (approved):

xNexus-ACS (approved):

zhaguanyang:
I would say the idea is fine, but you forgot to rename it in patches. You should actually build it before trying to PR.

Michal78900 (changes requested):
Fix build issues please

ClaudioPanConQueso:
> Fix build issues please

ok

ClaudioPanConQueso:
done

PR #1417, opened 2022-09-20
Move Preference and change BaseCodeEnum to Uint

louis1706:


joker-119:
The preferences are specific to a player and specific firearm, why move those things out into the player class when they require a firearm to be useful anyways?

louis1706:
> The preferences are specific to a player and specific firearm, why move those things out into the player class when they require a firearm to be useful anyways?

Because having Static into the Item::Firearm is pretty bad i thinks to find it when it's for a player and not really about the item it's self

NaoUnderscore:
Base game handles them through ItemBase/Firearm class, hence we'll keep using the same pattern.

PR #1418, opened 2022-09-20
Exiled::Loader::VersionName

louis1706:


Exiled.Loader/Loader.cs
```diff
@@ -90,6 +90,11 @@ static Loader()
         /// </summary>
         public static Random Random { get; } = new();
 
+        /// <summary>
+        /// Gets the NameVersion of the assembly.
```
zhaguanyang, line 94: Just call it ``VersionName`` instead of ``NameVersion``?

PintTheDragon:
Why?

louis1706:
> this [il y a 4 jours](#event-7511225528)

because it's help people to get the version name of the actual exiled branch ?

PR #1419, opened 2022-09-20
Little fixes

IRacle1:


Exiled.API/Features/Player.cs
```diff
@@ -79,27 +79,20 @@ public class Player
...
         private CustomHealthStat healthStat;
         private Role role;
-        private HashSet<EActor> components = new();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Player"/> class.
         /// </summary>
         /// <param name="referenceHub">The <see cref="global::ReferenceHub"/> of the player to be encapsulated.</param>
-        public Player(ReferenceHub referenceHub)
+        internal Player(ReferenceHub referenceHub)
```
Ficus-x, line 95: no

Exiled.API/Extensions/ItemExtensions.cs
```diff
@@ -197,12 +197,7 @@ public static byte GetMaxAmmo(this ItemType item)
         /// </summary>
         /// <param name="items">The items to convert.</param>
         /// <returns>A new <see cref="List{T}"/> of <see cref="ItemType"/>s.</returns>
-        public static IEnumerable<ItemType> GetItemTypes(this IEnumerable<Item> items)
-        {
-            Item[] arr = items.ToArray();
-            for (int i = 0; i < arr.Length; i++)
-                yield return arr[i].Type;
-        }
+        public static IEnumerable<ItemType> GetItemTypes(this IEnumerable<Item> items) => items.Select(item => item.Type);
```
Ficus-x, line 200: i think it's useless method.

Ficus-x (changes requested):

Exiled.API/Features/Player.cs
louis1706: Wtf why no ? use Player.Get()

IRacle1: ???

Exiled.API/Extensions/ItemExtensions.cs
IRacle1: Maybe

Exiled.API/Features/Player.cs
IRacle1: That ctor should be non-public, cuz it can broke Player.List

louis1706: > That ctor should be non-public, cuz it can broke Player.List

Exiled.API.Publicized ;)

Exiled.API/Extensions/ItemExtensions.cs
joker-119: Would not
```
foreach (Item item in items)
    yield return item.Type;``` not be more time efficient than linq, here?

IRacle1: Linq should do the same thing, but with predicates. It won't give such performance boost

Jesus-QC (approved):

PR #1424, opened 2022-09-24
Fix the GenerateRandom not being called | Also fix Ev.IsAllowed = false killing player

louis1706:


NaoUnderscore:
Master branch already implements this fix, we're gonna close this PR.

PR #1427, opened 2022-09-27
Convert Stamina::ProcessStamina Prefix to Transpiler

louis1706:


IRacle1:
#1425

IRacle1:
Also, you just copied IL code from decompiler think about why your implementation won't work at all

PR #1436, opened 2022-10-16
Changed the default value of CanFlashbangsAffectThrower to `false`

Michal78900:
In the base-game the thrower of the flashbangs does not get flashed no matter what (FF settings for example). Therefore, this config option should match base-game behaviour.

### DIFF
--- a/Exiled.Events/Config.cs
+++ b/Exiled.Events/Config.cs
@@ -22,13 +22,13 @@ namespace Exiled.Events
         /// Gets or sets a value indicating whether SCP-173 can be blocked or not by the tutorial.
         /// </summary>
         [Description("Indicates whether SCP-173 can be blocked or not by the tutorial")]
-        public bool CanTutorialBlockScp173 { get; set; } = true;
+        public bool CanTutorialBlockScp173 { get; set; } = false;
 
         /// <summary>
         /// Gets or sets a value indicating whether SCP-096 can be triggered or not by the tutorial.
         /// </summary>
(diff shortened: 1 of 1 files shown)


zhaguanyang (approved):

Jesus-QC (approved):

Michal78900:
I forgot about other things that alter base-game bevaiour

Exiled.Events/Config.cs
```diff
@@ -22,13 +22,13 @@ public sealed class Config : IConfig
         /// Gets or sets a value indicating whether SCP-173 can be blocked or not by the tutorial.
         /// </summary>
         [Description("Indicates whether SCP-173 can be blocked or not by the tutorial")]
-        public bool CanTutorialBlockScp173 { get; set; } = true;
```
joker-119, line 25: Both of these should be true to emulate base game behavior, unless this was recently changes by Northwood.

Michal78900: Tutorial blocks SCP-173 as well as he triggers SCP-096 on a server with no EXILED installed. So I think my changes are correct and will emulate base game behaviour.

joker-119: Re-read the names of those configs and try again. If these are both true, it means tutorials can stop 173 and trigger 096.

Michal78900: Oh, okay my bad.

Michal78900:
> I forgot about other things that alter base-game bevaiour

Nevermind

PR #1439, opened 2022-10-30
`Player.OnAimingDownSight` event fix

IRacle1:
yea, now event works for both states, but i think in Exiled 6.0 we need change event name to something like `ChangingAimStatus`, and I don't see the point of two props `AdsIn` and `AdsOut` instead of one, for example `AdsStatus`


IRacle1:
image
image

Michal78900 (dismissed):

xNexus-ACS (approved):

Michal78900 (approved):

xNexus-ACS (approved):

PR #1449, opened 2022-11-30
Fix Teleport Method

louis1706:
Using NW method is not bad :) it's does what we have needed here also fix people teleporting in wall or in object or in the void


Exiled.API/Features/Player.cs
```diff
@@ -2778,7 +2778,7 @@ public void Reconnect(ushort newPort = 0, float delay = 5, bool reconnect = true
         /// Teleports the player to the given <see cref="Vector3"/> coordinates.
         /// </summary>
         /// <param name="position">The <see cref="Vector3"/> coordinates to move the player to.</param>
-        public void Teleport(Vector3 position) => Position = position;
+        public void Teleport(Vector3 position) => ReferenceHub.playerMovementSync.OverridePosition(position);
```
IRacle1, line 2781: Better to just edit `Position` setter

louis1706: no :)

Exiled.API/Features/Player.cs
```diff
@@ -2778,7 +2778,12 @@ public void Reconnect(ushort newPort = 0, float delay = 5, bool reconnect = true
         /// Teleports the player to the given <see cref="Vector3"/> coordinates.
         /// </summary>
         /// <param name="position">The <see cref="Vector3"/> coordinates to move the player to.</param>
-        public void Teleport(Vector3 position) => Position = position;
+        public void Teleport(Vector3 position)
+        {
+            if (!PlayerMovementSync.FindSafePosition(position, out Vector3 safepos, true, true))
+                safepos = position;
+            ReferenceHub.playerMovementSync.OverridePosition(safepos);
```
iopietro, line 2785:
```suggestion
            if (!PlayerMovementSync.FindSafePosition(position, out Vector3 safepos, true, true))
                safepos = position;

            ReferenceHub.playerMovementSync.OverridePosition(safepos);
```

iopietro (changes requested):
Should we add the `isSafe` optional parameter to safely or not safely spawn the player at the chosen coordinates?

Thundermaker300:
Target dev branch.

louis1706:
I close this PR Because FindSafePosition no longer exist and nothing similar than this exist

PR #1450, opened 2022-12-01
Fix Error Item::Clone()

louis1706:
error: You cannot change the firerate of non-automatic weapons


IRacle1:
Now it is useless, already was implemented in Exiled-EA

PR #1460, opened 2023-01-05
ServerNamePatch to transpiler

Ficus-x:


iopietro:
Yeah but that's a single opcode.

Thundermaker300 (approved):

iopietro:
I don't think that we need a transpiler for this event, what do you think @Thundermaker300, @IRacle1?

Thundermaker300:
i mean it doesn't hurt

IRacle1:
Honestly, I think that such simple patches would be much clearer to leave as is

Thundermaker300:
your guys's call

PR #1461, opened 2023-01-05
Implement IsUsingStamina

Ficus-x:


IRacle1 (changes requested):
Not working, patch errors + literally base game dont use that method to update stamina, check `FpcStateProcessor::UpdateMovementState`

Ficus-x:
Okay, will fix it, ty

Exiled.Events/Patches/Generic/StaminaUsage.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            newInstructions.InsertRange(0, new CodeInstruction[]
+            {
+                new(OpCodes.Ldfld, Field(typeof(StaminaStat), nameof(StaminaStat.Hub))),
+                new(OpCodes.Call, Method(typeof(Player), nameof(Player.Get), new[] { typeof(ReferenceHub) })),
+                new(OpCodes.Dup),
+                new(OpCodes.Brfalse_S, skipLabel),
+                new(OpCodes.Callvirt, PropertyGetter(typeof(Player), nameof(Player.IsUsingStamina))),
+                new(OpCodes.Brfalse_S, returnLabel),
+                new CodeInstruction(OpCodes.Nop).WithLabels(skipLabel),
```
iopietro, line 41: Add comments like in other transpilers.

iopietro (changes requested):

Exiled.Events/Patches/Generic/StaminaUsage.cs
```diff
@@ -0,0 +1,52 @@
...
+        private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions, ILGenerator generator)
+        {
+            List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Shared.Rent(instructions);
+
+            Label returnLabel = generator.DefineLabel();
+            Label skipLabel = generator.DefineLabel();
+
+            newInstructions.InsertRange(0, new CodeInstruction[]
+            {
+                new(OpCodes.Ldfld, Field(typeof(StaminaStat), nameof(StaminaStat.Hub))),
```
IRacle1, line 35:
```suggestion
                new(OpCodes.Callvirt, PropertyGetter(typeof(StaminaStat), nameof(StaminaStat.Hub))),
```

Exiled.Events/Patches/Generic/StaminaUsage.cs
IRacle1, line 41: Stack error
call Player::Get `[player]`
dup `[player, player]`
brfalse `[player, player]` (if player == null, goto return label)
...
return:
ret `[player]`

IRacle1 (dismissed):

Exiled.Events/Patches/Generic/StaminaUsage.cs
```diff
@@ -0,0 +1,52 @@
...
+    using HarmonyLib;
+    using NorthwoodLib.Pools;
+    using PlayerStatsSystem;
+
+    using static HarmonyLib.AccessTools;
+
+    /// <summary>
+    /// Patches <see cref="StaminaStat.CurValue"/>.
+    /// </summary>
+    [HarmonyPatch(typeof(StaminaStat), nameof(StaminaStat.CurValue), MethodType.Setter)]
```
IRacle1, line 23: Also, i dont think that it is a good idea to prevent all sets.
Like, by default float value == 0f, and stamina always be empty

iopietro:
Wait, have you tested this transpiler? @Ficus-x

Exiled.Events/Patches/Generic/StaminaUsage.cs
iopietro: Also we can't patch `StaminaStat.CurValue` since `CurValue` doesn't get overridden by `StaminaStat`.

Exiled.Events/Patches/Generic/StaminaUsage.cs
iopietro, line 35: `StaminaStat.Hub` is not static, so you've to load its instance which in this case is represented by the `ldarg.0` opcode.

Ficus-x:
Now i tested it and it works.
But it was very hard to me to translate it to transpiler, then i made a Prefix.

iopietro:
We should convert it to a transpiler before merging it.

iopietro (approved):

Thundermaker300:
I don't see a problem with merging it now, although it is a complicated method and making it a transpiler should be a priority

Thundermaker300 (approved):

PR #1470, opened 2023-01-05
Inline NoClip Permited

louis1706:


Exiled.API/Features/Player.cs
```diff
@@ -381,13 +381,7 @@ public bool IsOverwatchEnabled
...
         {
             get => FpcNoclip.IsPermitted(ReferenceHub);
-            set
-            {
-                if (value && !FpcNoclip.PermittedPlayers.Contains(ReferenceHub.netId))
-                    FpcNoclip.PermitPlayer(ReferenceHub);
-                else if (!value && FpcNoclip.PermittedPlayers.Contains(ReferenceHub.netId))
-                    FpcNoclip.UnpermitPlayer(ReferenceHub);
-            }
+            set => _ = value ? FpcNoclip.PermitPlayer(ReferenceHub) : FpcNoclip.UnpermitPlayer(ReferenceHub)
```
iopietro, line 384:
```suggestion
            set => value ? FpcNoclip.PermitPlayer(ReferenceHub) : FpcNoclip.UnpermitPlayer(ReferenceHub)
```

iopietro (changes requested):

Exiled.API/Features/Player.cs
IRacle1: It wont work compile without `_ =`, C# operators moment

iopietro: Yes, because you can't assign a void value to the discard operator `_`.

Thundermaker300:
Does this *need* to be inline? I think it's fine as is

IRacle1:
> Does this _need_ to be inline? I think it's fine as is

Yeah, i dont see anything sus in it

iopietro:
It cannot be done inline, so no.

louis1706:
> It cannot be done inline, so no.

for me is work pefectly fine it's just need to use _ = who are equal to use OpCodes.Pop in Transpiler

Exiled.API/Features/Player.cs
louis1706: the problem is from a missing ;

Thundermaker300 (approved):

iopietro (approved):

PR #1480, opened 2023-01-07
[BUG] Wahead.Controller._alreadyDetonated is true after round restart

FakeMan2332:


louis1706:
image
Bruh NW Issue they never set it to false

Exiled.Events/Handlers/Internal/Round.cs
```diff
@@ -34,6 +34,7 @@ public static void OnWaitingForPlayers()
                 TranslationManager.Reload();
 
             RoundSummary.RoundLock = false;
```
louis1706, line 36: can you remove this ? it's should not be here

IRacle1:
Uh, warhead controller should changes between scenes, it is exiled caching problem

IRacle1:
#1481

IRacle1:
fixed

Thundermaker300:
Caching issue, which will be fixed in #1481.

PR #1483, opened 2023-01-08
Fix of Scp2176

Undid-Iridium:
image

@Thundermaker300


Undid-Iridium:
@louis1706 and @IRacle1 should evaluate this as I am just guessing off their code

Undid-Iridium:
IRacle fixed in separate PR

PR #1495, opened 2023-01-13
Create IWrapper interface

Thundermaker300:
Introduce a new interface: `IWrapper`. This interface is used on all wrapper classes that use a `Base` property (figuring out how to support this for `Player`). The goal is to force wrapper classes to all use a similar Base naming scheme, while also providing inherited docs (since all wrapper classes use very similar docs for the `Base` property). I will also be using this in part 2 of the pool API coming later...

### DIFF
--- a/Exiled.API/Features/Camera.cs
+++ b/Exiled.API/Features/Camera.cs
@@ -12,6 +12,7 @@ namespace Exiled.API.Features
     using System.Linq;
 
     using Enums;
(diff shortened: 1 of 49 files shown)


iopietro (approved):

louis1706 (approved):

louis1706:
I aprove but i thinks this should be added in Next SL breaking change

Because this will **EVERY** plugin

PR #1497, opened 2023-01-13
Move Map.FindParentRoom to Room class, add Room.Get(RelativePosition)

Thundermaker300:


Exiled.API/Features/Map.cs
```diff
@@ -186,6 +186,22 @@ public static Room FindParentRoom(GameObject objectInRoom)
             return room ?? Room.Get(objectInRoom.transform.position) ?? default;
         }
 
+        /// <summary>
+        /// Tries to find a room at the given <see cref="Vector3">position</see> using raycasting.
+        /// </summary>
+        /// <param name="position">The position to look for a room.</param>
+        /// <returns>The <see cref="Room"/> that the <see cref="Vector3"/> is located inside. Can be <see langword="null"/>.</returns>
+        public static Room FindParentRoom(Vector3 position)
+            => Room.Get(position);
```
iopietro, line 195:
```suggestion
        public static Room FindParentRoom(Vector3 position) => Room.Get(position);
```

Exiled.API/Features/Map.cs
```diff
@@ -186,6 +186,22 @@ public static Room FindParentRoom(GameObject objectInRoom)
...
+
+        /// <summary>
+        /// Tries to find a room at the given <see cref="RelativePosition">position</see> using raycasting.
+        /// </summary>
+        /// <param name="position">The position to look for a room.</param>
+        /// <returns>The <see cref="Room"/> that the <see cref="RelativePosition"/> is located inside. Can be <see langword="null"/>.</returns>
+        public static Room FindParentRoom(RelativePosition position)
+            => Room.Get(position.Position);
```
iopietro, line 203:
```suggestion
        public static Room FindParentRoom(RelativePosition position) => Room.Get(position.Position);
```

iopietro (changes requested):

iopietro (approved):

louis1706:
> I'm wondering if we should rename these to `FindRoom`, as `FindParentRoom` is kind of misleading for Vector3s that don't have a "parent" room.
your  FindRoom() are useless here they should be all In Room api (also Map.FindParentRoom because it's a Room so should be in Room)

Thundermaker300:
Moving `Map.FindParentRoom` is a breaking change. I added these because Room.Get() isn't exactly as obvious. Having them doesn't hurt anything

louis1706:
> Moving `Map.FindParentRoom` is a breaking change. I added these because Room.Get() isn't exactly as obvious. Having them doesn't hurt anything

I know it's breaking change it's why i don't want it to be moved for now but i thinks everything related to Room should be kept in Room

IRacle1 (approved):

louis1706 (approved):

PR #1506, opened 2023-01-14
Convert pools to use basegame pools

Thundermaker300:
Convert pools to use basegame pools under the hood. Keep all of our API while using northwood pools so that there's only one pool internally.


iopietro (dismissed):
Put all one-liners `=>` in the same line of the method declaration.

iopietro (approved):

louis1706 (approved):

PR #1519, opened 2023-01-16
Fix IsInPocketDimension

louis1706:


Exiled.API/Features/Player.cs
```diff
@@ -959,7 +959,7 @@ public bool BadgeHidden
         /// <summary>
         /// Gets a value indicating whether or not the player is in the pocket dimension.
         /// </summary>
-        public bool IsInPocketDimension => IsEffectActive<Corroding>() || Map.FindParentRoom(GameObject)?.Type == RoomType.Pocket;
+        public bool IsInPocketDimension => IsEffectActive<Corroding>() && Map.FindParentRoom(GameObject)?.Type == RoomType.Pocket;
```
Misaka-ZeroTwo, line 962:
```suggestion
        public bool IsInPocketDimension => Map.FindParentRoom(GameObject)?.Type == RoomType.Pocket;
```

I don't see the need to make a check on that effect, considering there isn't much need, and player could still be in pocket dimension without having the effect

louis1706: i thinks it's more about getting player getting normally in this room so people without effect like Teleporting player will not be considering in the Room could be good for some plugin to have this

Misaka-ZeroTwo: My point was plugin could teleport the player to the pocket dimension and they won't have the effect.

Thundermaker300: This should return true if they are physically in the pocket dimension, regardless of whether they have the effect.

Thundermaker300 (approved):

Thundermaker300:
merge conflict

Exiled.API/Features/Player.cs
```diff
@@ -964,7 +964,7 @@ public bool BadgeHidden
         /// <summary>
         /// Gets a value indicating whether or not the player is in the pocket dimension.
         /// </summary>
-        public bool IsInPocketDimension => IsEffectActive<Corroding>() || Room.FindParentRoom(GameObject)?.Type == RoomType.Pocket;
+        public bool IsInPocketDimension => CurrentRoom.Type == RoomType.Pocket;
```
Thundermaker300, line 967:
```suggestion
        public bool IsInPocketDimension => CurrentRoom?.Type == RoomType.Pocket;
```

louis1706: No it's already have null check

iopietro (approved):

PR #1525, opened 2023-01-16
Update UpgradingPlayer.cs

louis1706:


Thundermaker300 (approved):

Misaka-ZeroTwo (changes requested):
The transpiler needs to be fixed

IRacle1:
Why is it exists, when transpiler and so on fixed in #1527?

Misaka-ZeroTwo:
> Why is it exists, when transpiler and so on fixed in #1527?

Cuz Yamato thought the transpiler works.

PR #1532, opened 2023-01-17
Remove Patch if one of them failed to load

louis1706:
it's will prevent many bug when patch fail occurs


Misaka-ZeroTwo:
I suggest adding a try-catch for patching. So even if one failed to patch, then the worst thing that could have happened would just be certain event(s) didn't fire instead of everything just break.

louis1706:
> I suggest adding a try-catch for patching. So even if one failed to patch, then the worst thing that could have happened would just be certain event(s) didn't fire instead of everything just break.

they already contain Try-Catch everywhere

PR #1555, opened 2023-01-21
Add spawn protection API to Respawn class

Thundermaker300:


louis1706 (commented):
Can you add the SpawnProtected.Teams ?

louis1706 (approved):

Misaka-ZeroTwo (approved):

joker-119 (approved):
Personally I feel like these should be under the server API

But aside from that lgtm, if other devs think it belongs in respawn that's fine

iopietro:
If we move them to the `Respawn` class, we should remove the `Spawn` prefix, so we don't have to type `Respawn.Spawn[...]` to call them.

Thundermaker300:
I think it makes sense to put these in the Respawn class. Since, you know, they're all used solely for Respawning. With that said, I'll make pietro's change now.

iopietro (approved):

PR #1559, opened 2023-01-23
Readding Exiled.Bootstrap

louis1706:


iopietro:
Why?

louis1706:
> Why?

Make possible to install Exiled with LocalAdmin plugin installer 

Remove Exiled dependency from NWdependency file

(Permitting a new way to update Exiled automatically)

joker-119:
LA plugin installer can't install dependencies? :hmm:

joker-119:
Why do we keep trying to reads this?this makes no sense.

PR #1561, opened 2023-01-25
Adding Role.Set overload

Misaka-ZeroTwo:


Exiled.API/Features/Player.cs
```diff
@@ -1839,6 +1839,37 @@ public int RemoveItem(Func<Item, bool> predicate, bool destroy = true)
...
+                if (keepPosition)
+                {
+                    roleSpawnFlags = RoleSpawnFlags.AssignInventory;
+                }
+
+                if (keepInventory)
+                {
+                    roleSpawnFlags = RoleSpawnFlags.UseSpawnpoint;
+                }
+            }
```
iopietro, line 1868:
```suggestion
            if (keepPosition && keepInventory)
            {
                roleSpawnFlags = RoleSpawnFlags.None;
            }
            else
            {
                if (keepPosition)
                    roleSpawnFlags = RoleSpawnFlags.AssignInventory;

                if (keepInventory)
                    roleSpawnFlags = RoleSpawnFlags.UseSpawnpoint;
            }
```

iopietro:
We already have `Player::Role::Set`

iopietro (changes requested):

Misaka-ZeroTwo:
I could move it to Role class, would that be cool?

Misaka-ZeroTwo:
I miss the time when I could still do ``Role.Type = RoleTypeId``. Would cause a breaking change if I add that now.

Exiled.API/Features/Roles/Role.cs
```diff
@@ -202,6 +202,33 @@ protected Role(PlayerRoleBase baseRole)
...
+                roleSpawnFlags = RoleSpawnFlags.None;
+            }
+            else
+            {
+                if (keepPosition)
+                    roleSpawnFlags = RoleSpawnFlags.AssignInventory;
+
+                if (keepInventory)
+                    roleSpawnFlags = RoleSpawnFlags.UseSpawnpoint;
+            }
```
louis1706, line 227: This thing will not work

Also use it like a flag not like an enum

louis1706 (changes requested):

louis1706:
Also SpawnFlag is already pretty self explanatory i don't see the point to make this method actually

Exiled.API/Features/Roles/Role.cs
Misaka-ZeroTwo: Unfortunately it can't be used like a flag. The value doesn't sum up. Regarding if it works or not, I would check it later.

Misaka-ZeroTwo:
> Also SpawnFlag is already pretty self explanatory i don't see the point to make this method actually 

It's. I just felt that flag isn't obvious, in terms of it determines whether position and inventory will be saved, without checking it out.

louis1706 (approved):

iopietro (approved):

Exiled.API/Features/Roles/Role.cs
```diff
@@ -193,6 +193,14 @@ protected Role(PlayerRoleBase baseRole)
         public virtual void Set(RoleTypeId newRole, SpawnReason reason = Enums.SpawnReason.ForceClass) =>
             Set(newRole, reason, RoleSpawnFlags.All);
```
iopietro, line 194:
```suggestion
        public virtual void Set(RoleTypeId newRole, SpawnReason reason = Enums.SpawnReason.ForceClass) => Set(newRole, reason, RoleSpawnFlags.All);
```

Exiled.API/Features/Roles/Role.cs
```diff
@@ -193,6 +193,14 @@ protected Role(PlayerRoleBase baseRole)
...
 
+        /// <summary>
+        /// Sets the player's <see cref="RoleTypeId"/>.
+        /// </summary>
+        /// <param name="newRole">The new <see cref="RoleTypeId"/> to be set.</param>
+        /// <param name="spawnFlags">The <see cref="RoleSpawnFlags"/> defining player spawn logic.</param>
+        public virtual void Set(RoleTypeId newRole, RoleSpawnFlags spawnFlags) =>
+            Owner.RoleManager.ServerSetRole(newRole, (RoleChangeReason)Enums.SpawnReason.ForceClass, spawnFlags);
```
iopietro, line 202:
```suggestion
        public virtual void Set(RoleTypeId newRole, RoleSpawnFlags spawnFlags) => Owner.RoleManager.ServerSetRole(newRole, (RoleChangeReason)Enums.SpawnReason.ForceClass, spawnFlags);
```

iopietro (changes requested):

iopietro (approved):

PR #1566, opened 2023-01-25
``Pickup::Get(GameObject)``

louis1706:


Exiled.API/Features/Pickups/Pickup.cs
```diff
@@ -309,6 +309,13 @@ public static Pickup Get(ItemPickupBase pickupBase)
         /// <returns>return the Pickup with Serial choose.</returns>
         public static Pickup Get(ushort serial) => List.SingleOrDefault(x => x.Serial == serial);
 
+        /// <summary>
+        /// Gets the <see cref="Pickup"/> with the given <see cref="UnityEngine.GameObject"/>.
+        /// </summary>
+        /// <param name="gameObject"> The serial of the Pickup you search.</param>
+        /// <returns>return the Pickup with gameObject choose.</returns>
+        public static Pickup Get(GameObject gameObject) => Get(gameObject?.GetComponent<ItemPickupBase>());
```
iopietro, line 317:
```suggestion
        public static Pickup Get(GameObject gameObject) => Get(gameObject == null ? null : gameObject.GetComponent<ItemPickupBase>());
```

`?.` can't be used with Unity objects.

`if (pickupBase is null)` above should also be changed to `if (pickupBase == null)`.

iopietro (changes requested):

Exiled.API/Features/Pickups/Pickup.cs
```diff
@@ -309,6 +309,13 @@ public static Pickup Get(ItemPickupBase pickupBase)
         /// <returns>return the Pickup with Serial choose.</returns>
         public static Pickup Get(ushort serial) => List.SingleOrDefault(x => x.Serial == serial);
 
+        /// <summary>
+        /// Gets the <see cref="Pickup"/> with the given <see cref="UnityEngine.GameObject"/>.
+        /// </summary>
+        /// <param name="gameObject"> The serial of the Pickup you search.</param>
```
Thundermaker300, line 315: Fix doc :trollface:

iopietro (approved):

PR #1578, opened 2023-01-28
RoleExtensions

louis1706:


louis1706:
IDK Why Exiled event is modify i will remove the change later
(same for Exiled.Example it's was just for testing the output)

joker-119:
Just a question, why not move these extensions to just be methods in the role api class instead?

Role.GetSpawnPoint(RoleTypeId) for ex

Or an abstract method in the Role class that the other role classes implement in their instance

So you could just Role.GetSpawnPoint() and Scp173 role overrides it to return his Spawn point, 049 to return his, etc.

So if you already have an instance of the role (for ex on the player object) you can just Player.Role.GetSpawnPoint()

louis1706:
> Just a question, why not move these extensions to just be methods in the role api class instead?
> 
> Role.GetSpawnPoint(RoleTypeId) for ex
> 
> Or an abstract method in the Role class that the other role classes implement in their instance
> 
> So you could just Role.GetSpawnPoint() and Scp173 role overrides it to return his Spawn point, 049 to return his, etc.
> 
> So if you already have an instance of the role (for ex on the player object) you can just Player.Role.GetSpawnPoint()
> 
> We could do the same with a lot of other extension methods as well. 

Yeah it's the reason of why i have kept it in Draft

I would like to fix actual bug before modifying this

Thundermaker300:
I like the idea, but what if you want to get spawnpoints of a role and you don't have the actual object?

PR #1582, opened 2023-01-29
Fix Flashlight IL Error && Fix infinite connect && Ammo drop in ChangingRole && EndingRound NWPluginAPI Fix

louis1706:

PR #1583, opened 2023-01-31
Fix null player on the spawned event

xNexus-ACS:


xNexus-ACS:
this PR depends of the changes maded in the #1582 ChangingRolePatch

Exiled.Events/EventArgs/Player/SpawnedEventArgs.cs
```diff
@@ -27,9 +27,9 @@ public class SpawnedEventArgs : IPlayerEvent
         /// <param name="oldRole">the spawned player's old <see cref="PlayerRoleBase">role</see>.</param>
         /// <param name="roleChangeReason">the player's spawned with <see cref="RoleChangeReason" />.</param>
         /// <param name="roleSpawnFlags">the player's spawned with <see cref="RoleSpawnFlags" />.</param>
-        public SpawnedEventArgs(ReferenceHub player, PlayerRoleBase oldRole, RoleChangeReason roleChangeReason, RoleSpawnFlags roleSpawnFlags)
+        public SpawnedEventArgs(Player player, PlayerRoleBase oldRole, RoleChangeReason roleChangeReason, RoleSpawnFlags roleSpawnFlags)
         {
-            Player = Player.Get(player);
+            Player = Player.Get(player.ReferenceHub);
```
louis1706, line 32: Wtf you take Player.Referencehub inside a Player.Get() lol

louis1706 (approved):

louis1706 (changes requested):
Miss Click lol

Exiled.Events/EventArgs/Player/SpawnedEventArgs.cs
xNexus-ACS: because the event parameters has been changed to `Player player` instead of `ReferenceHub player`
and if I try to take something that is not the referencehub it says that the player is null

xNexus-ACS: I created a null check but it says that the player is always null

xNexus-ACS: So I decided to make it like that, at least player is not null now

xNexus-ACS: oh right, I didn't see what do you mean

iopietro, line 32:
```suggestion
            Player = player;
```

iopietro (changes requested):

louis1706:
Nexus we have no change in this pr ? 
Why ?

xNexus-ACS:
Decided to close this bcs its seems it didn't changed anything for now

PR #1591, opened 2023-02-03
Temporarily disable dict pool

Thundermaker300:
I believe the returned dictionary may always be the same, resulting in some broken cases with players. If someone wants to fix I'd be more than happy to add that, I just don't know how to fix atm :)


louis1706:
what is the actual issue of this ?

Thundermaker300:
> what is the actual issue of this ?

Returned Dictionary appears to be the same object, causing every player to have the same session variables

PR #1612, opened 2023-02-09
Makes CustomItems Spawning at the player position when escaping

xNexus-ACS:


xNexus-ACS:
this has been suggested in the suggestion channel

Thundermaker300 (approved):

PR #1614, opened 2023-02-10
Nickname Issue Fix

xNexus-ACS:


louis1706 (approved):
I don't see an other way to fix this

xNexus-ACS:
This is a temporary fix until I or someone else found another way to fix it

PR #1618, opened 2023-02-11
Add Repsawn.NtfHelicopter and Respawn.ChaosVan (GameObjects)

Thundermaker300:


Exiled.API/Features/Respawn.cs
```diff
@@ -15,12 +15,26 @@ namespace Exiled.API.Features
...
     /// </summary>
     public static class Respawn
     {
+        private static GameObject ntfHelicopterGameObject;
+        private static GameObject chaosCarGameObject;
+
+        /// <summary>
+        /// Gets the NTF Helicopter's <see cref="GameObject"/>.
+        /// </summary>
+        public static GameObject NtfHelicopter => ntfHelicopterGameObject ??= GameObject.Find("Chopper");
```
iopietro, line 31: `??=` should be avoided when working with Unity objects.

iopietro (changes requested):

Exiled.API/Features/Respawn.cs
Thundermaker300: omg I took this out and then added it back I'm so silly. I'll fix today

Thundermaker300: I forgor...

louis1706 (approved):

PR #1623, opened 2023-02-14
More SCP Events (106)

xNexus-ACS:


xNexus-ACS:
I have tested all changed that I made, including all properties in the new event and all is working correctly

xNexus-ACS:
image

louis1706:
Same than 049 PR put TODO:// Rework Transpiler

louis1706:
> image

it's better and less work to test it with the DevTools plugin from Sanye :3

xNexus-ACS:
> Same than 049 PR put TODO:// Rework Transpiler

already did

xNexus-ACS:
> > image
> 
> it's better and less work to test it with the DevTools plugin from Sanye :3

true, but I don't want my console spammed XD

louis1706:
it's will be good to keep TeleportingEventArgs instead of UsingAtlas ? i don't disagree with it  

but it's will make an other break change for dev

xNexus-ACS:
> it's will be good to keep TeleportingEventArgs instead of UsingAtlas ? i don't disagree with it
> 
> but it's will make an other break change for dev

The change to UsingAtlas is currently easier to understand since there is currently no teleport method as such like in 11.0, but now to teleport the player uses the Atlas

louis1706 (approved):

IRacle1 (approved):

PR #1627, opened 2023-02-15
FixNWSerialBeing 0

louis1706:


iopietro (approved):

joker-119:
This can't be fixed in the getter like this.

If for ex, the base info serial is 0, this will generate a new serial for the return, but it does not update the base info to use the new serial, it remains 0 and thus will get another new serial the next time this getter is called, or when basegame assigns it one normally.

You need to also update the base info serial anytime you generate a new serial for the property return

louis1706:
> This can't be fixed in the getter like this.
> 
> If for ex, the base info serial is 0, this will generate a new serial for the return, but it does not update the base info to use the new serial, it remains 0 and thus will get another new serial the next time this getter is called, or when basegame assigns it one normally.
> 
> You need to also update the base info serial anytime you generate a new serial for the property return

I use Serial setter so it's updated it ???

IRacle1:
Serial equals to 0 is not a bug, its a valid state for non properly spawned items

Thundermaker300:
Won't this make it so unspawned items will also have a serial?

louis1706:
> Won't this make it so unspawned items will also have a serial?

that will never cause any issue

Thundermaker300 (approved):

PR #1630, opened 2023-02-18
Make CustomRole appears above the name

louis1706:


Thundermaker300 (approved):

louis1706:
Do not merge i will make a better change for this according to an actual suggestions on discord

IRacle1:
It will break all plugins that changing players nicknames

louis1706:
> It will break all plugins that changing players nicknames

only if they have CustomRole

IRacle1:
> > It will break all plugins that changing players nicknames
> 
> only if they have CustomRole and if they change name after the Role Has been set

Yes, but it is the problem

Thundermaker300:
We shouldn't be trying to judge what is more or less of a breaking change. If you make a change that changes a previously working feature to no longer work, it's a breaking change. That's all there is to it.

IRacle1:
> We shouldn't be trying to judge what is more or less of a breaking change. If you make a change that changes a previously working feature to no longer work, it's a breaking change. Breaking changes are no longer allowed. That's all there is to it.

That pr is not like a breaking change, it breaks the displayed nickname system with custom roles at all, and we 2 ways:
- Write a complex patches/api/other things for supporting that
- Accept the fact that we wont be able to move CustomInfo😭

Personally I think this change is *too* insignificant to try implement it

PR #1631, opened 2023-02-18
Player Get can now return Server.Host

louis1706:
- Fixing than Player.IsCuffed not being true when cuffed by Server


joker-119:
This is a bad way of doing it, as it could break things like attacker being null when taking damage from the server (plugin/command) etc.
I would specifically fix it for the IsCuffed property rather than making Player.Get able to return the server.

PR #1632, opened 2023-02-19
EscapeScenario for Escaping Event

louis1706:


louis1706:
Tested with DevTools on Windows Server

joker-119:
Tbh I'm not sure why this is even necessary

louis1706:
> Tbh I'm not sure why this is even necessary

the important change it's about than it's will be trigger with Ev.IsAllowed = false for 
other roletype than Scientist and Dclass

Thundermaker300 (approved):

IRacle1:
> > Tbh I'm not sure why this is even necessary
> 
> the important change it's about than it's will be trigger with Ev.IsAllowed = false for 
> other roletype than Scientist and Dclass

Yeah, i happy that now NW make it possible, but it will be breaking change

louis1706:
> > > Tbh I'm not sure why this is even necessary
> > 
> > 
> > the important change it's about than it's will be trigger with Ev.IsAllowed = false for
> > other roletype than Scientist and Dclass
> 
> Yeah, i happy that now NW make it possible, but it will be breaking change

it's not breaking ??

Thundermaker300:
It's only breaking if somebody *relied* on the classd/scientist only escaping behavior, which is kind of a weird thing to rely on

IRacle1:
> It's only breaking if somebody _relied_ on the classd/scientist only escaping behavior, which is kind of a weird thing to rely on

Maybe, but fact is fact -> event now calls for a few more situations

louis1706:
> event now calls for a few more situations

still that not breaking change :3

joker-119:
Technically anything that changes the behavior of the event qualifies as a breaking change, even if the plugins themselves can still use the event without recompile, because it requires code changes to use the event as it was intended in the plugin.

PR #1637, opened 2023-02-20
Starting Recall Event Fix

xNexus-ACS:


xNexus-ACS:
the event was being called every second when the 049 is starting to revive the player bcs the ServerValidateAny method is used in the Updated one, so I changed the target patch method and recoded the patch (I know it works like a Prefix or something)

xNexus-ACS:
I marked it as a draft bcs I messed something

xNexus-ACS:
All working without errors and bugs

louis1706 (approved):

Thundermaker300 (approved):

PR #1638, opened 2023-02-21
Fix late join+remove spam log info

louis1706:


Exiled.Events/Events.cs
```diff
@@ -115,14 +116,14 @@ public override void OnDisabled()
             Unpatch();
 
             DisabledPatchesHashSet.Clear();
-
             SceneManager.sceneUnloaded -= Handlers.Internal.SceneUnloaded.OnSceneUnloaded;
```
IRacle1, line 119: why

IRacle1 (changes requested):

Exiled.Events/Events.cs
louis1706: Idk lol

IRacle1 (approved):

Thundermaker300 (approved):

PR #1647, opened 2023-02-27
Adding Role::Type setter

VALERA771:


xNexus-ACS:
I recommend you testing this PR a lot, bcs changing a lot of properties used in patches can cause bugs

VALERA771:
> I recommend you testing this PR a lot, bcs changing a lot of properties used in patches can cause bugs, also I don't think why you need a Role::Type setter when you have Role::Set lmao, I can't find the utility of this PR

I understand you, but for me it'll be useful. Also can be useful in updating old plugins cuz in that exiled version it has a setter

xNexus-ACS:
> > I recommend you testing this PR a lot, bcs changing a lot of properties used in patches can cause bugs, also I don't think why you need a Role::Type setter when you have Role::Set lmao, I can't find the utility of this PR
> 
> I understand you, but for me it'll be useful. Also can be useful in updating old plugins cuz in that exiled version it has a setter

yeah, I can understand what you're saying, but why not use `Role.Set(Role.Type, SpawnFlags)` instead of making a setter that can't be overriden

VALERA771:
> > > I recommend you testing this PR a lot, bcs changing a lot of properties used in patches can cause bugs, also I don't think why you need a Role::Type setter when you have Role::Set lmao, I can't find the utility of this PR
> > 
> > 
> > I understand you, but for me it'll be useful. Also can be useful in updating old plugins cuz in that exiled version it has a setter
> 
> yeah, I can understand what you're saying, but why not use `Role.Set(Role.Type, SpawnFlags)` instead of making a setter that will not be overriden

Okay. Seems that it won't be useful or useful only in few cases

PR #1649, opened 2023-02-28
RelativePosition

louis1706:


IRacle1 (approved):

xNexus-ACS (approved):

Exiled.API/Features/Player.cs
```diff
@@ -490,6 +490,15 @@ public Vector3 Position
             set => ReferenceHub.TryOverridePosition(value, Vector3.zero);
         }
 
+        /// <summary>
+        /// Gets or sets the relative player's position.
```
Thundermaker300, line 494: Add a remark that the value will be 0,0,0,0 if they are not an FPC role.

Thundermaker300 (changes requested):

Exiled.API/Features/Player.cs
louis1706: Okay

louis1706: done

Exiled.API/Features/Player.cs
```diff
@@ -490,6 +490,16 @@ public Vector3 Position
...
+        /// </summary>
+        /// <remarks>value will be default if the player it's not an <see cref="FpcRole"/>.</remarks>
```
Thundermaker300, line 496:
```suggestion
        /// <remarks>The value will be default if the player's role is not an <see cref="FpcRole"/>.</remarks>
```

Thundermaker300 (changes requested):

Thundermaker300 (approved):

PR #1654, opened 2023-03-01
Fix Spamming Ability

xNexus-ACS:
Someone found a issue with the Call event args (049) that can be spammed to give zombies infinite shield lmao


IRacle1 (approved):

louis1706 (changes requested):
Make ev.IsAllowed = callAbility._serverTriggered || !callAbility.Cooldown.IsReady

louis1706 (approved):

PR #1657, opened 2023-03-01
DoorType::AirLock

louis1706:


Thundermaker300:
Add something like IsNonInteractable since you're adding api related to it :trollface:

Thundermaker300 (approved):

IRacle1 (approved):

PR #1676, opened 2023-03-07
EffectType::SpawnProtected

louis1706:


Thundermaker300 (approved):
I'm actually kinda surprised this wasn't already there.. I swear we had it. Huh.

IRacle1 (approved):

PR #1680, opened 2023-03-11
Change NewRole Setter

louis1706:


joker-119 (commented):
You're going to be calling the ChangingRole event from inside the changing role event. 

Imagine someone has this this code to swap all scientists and dbois in their server when a round starts:

`
if (ev.NewRole == RoleTypeId.Classd)
    ev.NewRole = RoleTypeId.Scientist;
else if (ev.NewRole == RoleTypeId.Scientist)
    ev.NewRole = RoleTypeId.ClassD`

This will create a StackOverflow-like condition where everyone who spawns as a class-d or scientist will just infinitely loop between the two roles forever.

PR #1683, opened 2023-03-13
Just some minor stuff.

joker-119:
Added lookup tables for CustomRole and CustomItem Get(string), Get(uint), Get(Type) methods.
Added CustomItem.Get(Type)
Added bool to CustomRole class that can be set to make the role not allow custom items in the inventory config to show the received item hint.
Support nullable types.


NaoUnderscore (changes requested):
Nullable types must be fully implemented, regardless how stylecop behaves.
Old C# patterns shoudn't be preferred over new C# >=`CUR_VER` ones.

PR #1689, opened 2023-03-21
Better can sense tutorial

xNexus-ACS:


xNexus-ACS:
not tested (do not merge)

xNexus-ACS:
I will implement the yamato's suggest (Taking another player as target is the tutorial or turnedplayers condition is met) in another PR bcs it needs another patch and I want to implement that first

xNexus-ACS:
Also tested and working

louis1706:
I would say finish before merging

PR #1697, opened 2023-03-24
Fix removing both custom roles when second one is added

Jon02:
If a player already has a custom role and gets another one both roles are removed.
TrackedPlayers should be called after setting the role so InternalChangingRole won't trigger.


joker-119:
Adding them to the tracker doesn't cause the changing role event to trigger.

Them changing roles causes it to trigger.

If you want someone to keep a role when assigned another one, the first one given needs to have `keep_role_on_changing_role` set to true.

PR #1698, opened 2023-03-24
Just some tidying up.

joker-119:
- Make Exiled.Installer return Code 1 when an exception is encountered during updating.
- Make Exiled.Updater not restart the server if the Installer returns code 1.
- Update Exiled.Installer to a .net6 app (from .netcore3) to support libssl3 on newer distros (like ubuntu 24.x+)


Michal78900 (approved):
Looks okay, however why stick to net6.0, can't we update to net7.0 since 8.0 is in the works?

joker-119:
We could, I just don't have net7 installed locally yet so instead of fucking about with that I just set it to 6 for now to test

PR #1699, opened 2023-03-26
Player::RemovePickup method

xNexus-ACS:


joker-119:
By definition, a player cannot have a pickup in their inventory.

Doesn't player.RemoveItem(ushort serial) already exist?

xNexus-ACS:
> By definition, a player cannot have a pickup in their inventory.
> 
> Doesn't player.RemoveItem(ushort serial) already exist?

We don't have any overload for RemoveItem that have ushort serial
I created that method for the thing that we can't get Pickups and remove them with this method bcs this method only have the Item parameter and the overload Func<Item, ...>

joker-119:
I'd much rather this just be an overload for RemoveItem that accepts a ushort for the serial.

Because like i said, by definition, you cannot have a pickup in your inventory, so if you have a pickup reference and someone picks it up, you're either doing something really janky to keep that pickup reference alive, or your pickup reference should be null, since there's no pickup on the server as it is now an item in their inventory.

xNexus-ACS:
> I'd much rather this just be an overload for RemoveItem that accepts a ushort for the serial.
> 
> Because like i said, by definition, you cannot have a pickup in your inventory, so if you have a pickup reference and someone picks it up, you're either doing something really janky to keep that pickup reference alive, or your pickup reference should be null, since there's no pickup on the server as it is now an item in their inventory.

sure, okay :)

xNexus-ACS:
im going to close this pr then

PR #1703, opened 2023-04-01
Throw Method for CustomGrenades

xNexus-ACS:
Not tested for now, gonna test it later


xNexus-ACS:
tested

xNexus-ACS:
> 

yeah can be done here, but the problem is, before Exiled-EA, we had a Throw method for CustomGrenades that has been removed bcs I don't know why someone doesn't know that RelativePosition property exists or smth in the PickupInfo, I can do the method in the GrenadePickup too

louis1706 (approved):

joker-119 (approved):

PR #1710, opened 2023-04-11
Lift::Players

louis1706:


Misaka-ZeroTwo (approved):

Misaka-ZeroTwo (approved):

Exiled.API/Features/Lift.cs
```diff
@@ -77,6 +77,11 @@ internal Lift(ElevatorChamber elevator)
         /// </summary>
         public IReadOnlyCollection<ElevatorDoor> Doors => internalDoorsList;
 
+        /// <summary>
+        /// Gets a <see cref="IEnumerable{T}"/> of <see cref="Player"/> in the <see cref="Room"/>.
+        /// </summary>
+        public IEnumerable<Player> Players => Player.List.Where(x => Bounds.Contains(Position));
```
Thundermaker300, line 83: Shouldn't this be `Bounds.Contains(x.Position)`? This will get every player otherwise, since the condition will always be `true`.

Thundermaker300: how did neither of you monkeys notice this? i got u

Thundermaker300 (approved):

PR #1717, opened 2023-04-12
TeslaGate NRE fix

VALERA771:
Just another null check


Exiled.API/Features/TeslaGate.cs
```diff
@@ -270,7 +270,12 @@ public void ForceTrigger()
         /// </summary>
         /// <param name="player">The <see cref="Player"/> to check.</param>
         /// <returns><see langword="true"/> if the given <see cref="Player"/> is in the idle range of the tesla gate; otherwise, <see langword="false"/>.</returns>
-        public bool IsPlayerInIdleRange(Player player) => Base.PlayerInIdleRange(player.ReferenceHub);
+        public bool IsPlayerInIdleRange(Player player)
+        {
+            if (player.ReferenceHub is null)
```
louis1706, line 275: ReferenceHub can't be null if something it's null it's player

louis1706 (changes requested):

louis1706 (changes requested):
Same than before

PR #1718, opened 2023-04-13
More Role API Methods

xNexus-ACS:


Exiled.API/Features/Roles/Scp173Role.cs
```diff
@@ -247,6 +258,23 @@ public GameObject Tantrum(bool failIfObserved = false, float cooldown = 0)
         /// <param name="targetPos">The Target Position.</param>
         public void Blink(Vector3 targetPos) => BlinkTimer.ServerBlink(targetPos);
 
+        /// <summary>
+        /// Snap a <see cref="Player"/> (Attack).
+        /// </summary>
+        /// <param name="player">The <see cref="Player"/>to snap.</param>
+        public void Snap(Player player)
```
joker-119, line 265: If we are going to have an Attack() method for every SCP, why are we not defining it as an abstract void in an IScpRole interface or such, that way it's called the same thing in all of them?

xNexus-ACS: Uhm good question, I didn't thonk about it, gonna do that then

xNexus-ACS: working on it

xNexus-ACS:
not finished, I need to test the attack method in 939, 0492 & 096

louis1706 (approved):

Exiled.API/Features/Roles/Scp106Role.cs
```diff
@@ -82,7 +82,7 @@ internal Scp106Role(Scp106GameRole baseRole)
         /// <summary>
         /// Gets the <see cref="Scp106Attack"/>.
         /// </summary>
-        public Scp106Attack Attack { get; }
+        public Scp106Attack AttackAbility { get; }
```
louis1706, line 85: There a breaking change here

louis1706 (changes requested):

Exiled.API/Features/Roles/Scp106Role.cs
xNexus-ACS: it makes conflict with the Attack Method

xNexus-ACS: that's why I changed it

louis1706 (approved):
BreakChange

PR #1722, opened 2023-04-16
Adding SendingCommand Event

souf-bruh:
yey


louis1706:
No Event with command will be accepted

PR #1723, opened 2023-04-17
Using const && Optimisation

louis1706:


Exiled.API/Enums/SpawnReason.cs
```diff
@@ -10,7 +10,7 @@ namespace Exiled.API.Enums
     /// <summary>
     /// Possible spawn reasons.
     /// </summary>
-    public enum SpawnReason : byte
+    public enum SpawnReason : byte // TOTO: Remove this file and use Basegame
```
joker-119, line 13: Why remove this and use basegame when basegame can change it and break plugins, while we can just update the values.

Exiled.API/Features/Generator.cs
```diff
@@ -224,10 +224,8 @@ public KeycardPermissions KeycardPermissions
         /// </summary>
         /// <param name="scp079Generator">The <see cref="Scp079Generator"/> instance.</param>
         /// <returns>A <see cref="Generator"/> or <see langword="null"/> if not found.</returns>
-        public static Generator Get(Scp079Generator scp079Generator) =>
-            Scp079GeneratorToGenerator.TryGetValue(scp079Generator, out Generator generator) ?
-            generator :
-            new(scp079Generator);
+        public static Generator Get(Scp079Generator scp079Generator) => scp079Generator == null ? null :
```
joker-119, line 227: Does TryGetValue() not just return null if it's given a null scp079Generator value?

louis1706: image
nope it's will return NRE

Exiled.API/Enums/SpawnReason.cs
louis1706: i was thinking they was useless to make since NW Have it they just make an other same enum  (for example RoleType we use the BaseGame one not an enum on exiled and so they don't need to be updated)

louis1706:
there no enhancement here

Misaka-ZeroTwo:
optimization = enhancement, no?

Exiled.API/Enums/DoorType.cs
```diff
@@ -59,6 +61,7 @@ public enum DoorType
         /// <summary>
         /// Represents the 106_BOTTOM door.
         /// </summary>
+        [Obsolete("This door do not Exist")]
```
Thundermaker300, line 64:
```suggestion
        [Obsolete("This door no longer exists, and this enum item will be removed soon.")]
```

xNexus-ACS:
tested

louis1706:
> tested, but I recommend to not remove the CallDelayed inside the ResetInventory method

it's useless now ? since nw correctly get it's packet in the right order

PR #1725, opened 2023-04-17
Remove Timing.CallDelayed if useless

louis1706:


Exiled.API/Features/Player.cs
```diff
@@ -2386,13 +2387,6 @@ public Item AddItem(ItemBase itemBase, Item item = null)
                     acquisitionConfirmationTrigger.AcquisitionAlreadyReceived = true;
                 }
 
-                // Dont care, didnt ask, ratio
-                Timing.CallDelayed(0.02f, () =>
-                {
-                    if (item.Type is ItemType.SCP330 && item.Base != null)
-                        ((Scp330)item).Base.ServerRefreshBag();
```
joker-119, line 2393: This is necessary for the client to get the contents of the bag properly.

Exiled.API/Features/Player.cs
```diff
@@ -2467,25 +2461,16 @@ public bool TryAddCandy(CandyKindID candyType)
         {
             if (Scp330Bag.TryGetBag(ReferenceHub, out Scp330Bag bag))
             {
-                bool flag = bag.TryAddSpecific(candyType);
-
-                if (flag)
-                    bag.ServerRefreshBag();
```
joker-119, line 2473: Same as above.

Exiled.API/Features/Player.cs
```diff
@@ -2467,25 +2461,16 @@ public bool TryAddCandy(CandyKindID candyType)
...
 
             if (Items.Count > 7)
                 return false;
 
             Scp330 scp330 = (Scp330)AddItem(ItemType.SCP330);
 
-            Timing.CallDelayed(0.02f, () =>
-            {
-                scp330.Base.Candies.Clear();
-                scp330.AddCandy(candyType);
```
joker-119, line 2486: 1-frame delay is necessary to give client time to create the bag object on it's end before trying to add something to it.

Exiled.API/Features/Player.cs
```diff
@@ -2497,16 +2482,8 @@ public void ResetInventory(IEnumerable<ItemType> newItems)
         {
             ClearInventory();
 
-            Timing.CallDelayed(
```
joker-119, line 2500: Player items are sent the frame after they are removed, therefore you need to wait at least 1 frame between removing the last one and adding a new one, or the client could hit a race condition where it's trying to add an item into a slot index that's already occupied.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -777,7 +777,7 @@ public virtual void Give(Player player, Item item, bool displayMessage = true)
                 if (displayMessage)
                     ShowPickedUpMessage(player);
 
-                Timing.CallDelayed(0.05f, () => OnAcquired(player));
```
joker-119, line 780: frame delay necessary to ensure the client has actually processed receiving the item before you try to do something with it in the OnAcquired override method.

Exiled.CustomItems/API/Features/CustomItem.cs
```diff
@@ -1086,17 +1086,21 @@ private void OnInternalOwnerEscaping(EscapingEventArgs ev)
                     continue;
 
                 ev.Player.RemoveItem(item);
-
                 TrackedSerials.Remove(item.Serial);
-
-                Timing.CallDelayed(1.5f, () => Spawn(ev.Player.Position, item, null));
+                ev.Player.SessionVariables.Add("CustomItem", item);
```
joker-119, line 1090: This is a novel idea, but will entirely break if the player has multiple custom items.

Exiled.CustomRoles/API/Features/CustomRole.cs
```diff
@@ -902,16 +902,8 @@ private void OnInternalChangingRole(ChangingRoleEventArgs ev)
                 if (Ammo.Count > 0)
                 {
                     ev.Ammo.Clear();
-                    Timing.CallDelayed(
```
joker-119, line 905: Delay necessary to preserve ammo values if they exceed the server's configured ammo limits. Doing it with ev.Ammo results in excess ammo being dropped on the ground below them, rather than overriding the ammo limits.

joker-119 (changes requested):

Exiled.CustomRoles/API/Features/CustomRole.cs
louis1706: that seems more about a needed fix in the code ?

Exiled.CustomItems/API/Features/CustomItem.cs
louis1706: true true i will change it in a better way "CustomItem-{Serial]"

Exiled.API/Features/Player.cs
louis1706: i will like to request someone to test it at least to be 100% sure

Exiled.API/Features/Player.cs
louis1706: same "i will like to request someone to test it at least to be 100% sure"

because they have change how Item was sync so i am pretty curious

Exiled.CustomItems/API/Features/CustomItem.cs
louis1706: i don't get why we need client to have it ?

PR #1729, opened 2023-04-18
VoiceChat patch.

Monaldcry7788:
Removed old voice chat patch from ChangingSpeakerStatusAndVoiceChatting.cs and created VoiceChatting.cs for use instead of VoiceModuleBase VoiceTransceiver.ServerReceiveMessage for get the actual message.


louis1706 (approved):

louis1706:
If you want this to be merge for Exiled 6 remove the breaking change

Monaldcry7788:
The voice message is a breaking, but it's the cause of the Pr, so i think i will wait untill exiled

Issue #1724, opened 2023-04-17
console/remoteadmin command Customroles info [Id] bug


Vladislav2437:
**Describe the bug**
When I use the command in the console Customroles info (customroleid)
the server shows this error:
 Command execution failed! Error: System.StackOverflowException: The requested operation caused a stack overflow
at Exiled.CustomRoles.API.Features.CustomRole.TryGet (System.UInt32 id, Exiled.CustomRoles.API.Features.CustomRole& customRole) <0x21b7c084b40 + 0x00008> in <1db9570839fa40d78d3e9b0a39d7b95e>:0
                                   at Exiled.CustomRoles.API.Features.CustomRole.TryGet (System.UInt32 id, Exiled.CustomRoles.API.Features.CustomRole& customRole) [0x00000] in <1db9570839fa40d78d3e9b0a39d7b95e>:0

sending this lines about one hundred times, instead if i use the same command in the remoteadmin it kick me out of the server independently if the customrole Id is valid or not

**To Reproduce**
1) Register a custom role,
2) try the command: CustomRoles info [id]
3)error

**Expected behavior**
it should show the info about that custom role

**Server logs**
Please include a pastebin of your localadmin log file (or both MA_log and SCP_log files if you use MultiAdmin) from the time in which the bug occured

**EXILED Version (6.1.0 beta 3):**

[2023-04-17 22:05:43.315 +02:00] Total number of plugins: 8
                                 Enabled plugins: 8
                                 Disabled plugins: 0

                                 Exiled.Events:
                                        - Author: Exiled Team
                                        - Version: 6.1.0.0
                                        - Required Exiled Version: 6.1.0.0
                                        - Prefix: exiled_events
                                        - Priority: Highest
                                 Exiled.Permissions:
                                        - Author: Exiled Team
                                        - Version: 6.1.0.0
                                        - Required Exiled Version: 6.1.0.0
                                        - Prefix: exiled_permissions
                                        - Priority: Medium
                                 Exiled.CreditTags:
                                        - Author: Babyboucher20 & iRebbok & Exiled Team
                                        - Version: 6.1.0.0
                                        - Required Exiled Version: 6.1.0.0
                                        - Prefix: exiled_credits
                                        - Priority: Medium
                                 Exiled.Updater:
                                        - Author: Exiled.Updater
                                        - Version: 3.1.2.0
                                        - Required Exiled Version: 6.1.0.0
                                        - Prefix: exiled_updater
[...]

PR #1732, opened 2023-04-18
Forgot about TryGet

VALERA771:
Edited a Get, but not TryGet

Fix for #1724


IRacle1 (approved):
It would be better if you move documentation to a non-obsolete method, and will inherit to a obsolete, but it's not that important

louis1706 (approved):

Issue #1739, opened 2023-04-20
[BUG] OnInternalExplodingGrenade (CustomItems NRE)


ITeMbI4:
```
[2023-04-01 23:41:27.524 +00:00] [ERROR] [Exiled.Events] Method "OnInternalExplodingGrenade" of the class "Exiled.CustomItems.API.Features.CustomGrenade" caused an exception when handling the event "Exiled.Events.Events+CustomEventHandler`1[[Exiled.Events.EventArgs.Map.ExplodingGrenadeEventArgs, Exiled.Events, Version=6.0.1.0, Culture=neutral, PublicKeyToken=null]]"
[2023-04-01 23:41:27.537 +00:00] [ERROR] [Exiled.Events] System.NullReferenceException: Object reference not set to an instance of an object
                                   at InventorySystem.Items.Pickups.ItemPickupBase.InfoReceived (InventorySystem.Items.Pickups.PickupSyncInfo oldInfo, InventorySystem.Items.Pickups.PickupSyncInfo newInfo) [0x00000] in <dbaa2e8a14064a9eaeb4b5505e16639e>:0 
                                   at InventorySystem.Items.Pickups.ItemPickupBase.set_NetworkInfo (InventorySystem.Items.Pickups.PickupSyncInfo value) [0x00060] in <dbaa2e8a14064a9eaeb4b5505e16639e>:0 
                                   at Exiled.API.Features.Pickups.Pickup.set_Info (InventorySystem.Items.Pickups.PickupSyncInfo value) [0x0001f] in <af51d992e06540ea87818ba18e9f6bb1>:0 
                                   at Exiled.API.Features.Pickups.Pickup.set_Serial (System.UInt16 value) [0x00018] in <af51d992e06540ea87818ba18e9f6bb1>:0 
                                   at Exiled.API.Features.Pickups.Pickup.get_Serial () [0x00019] in <af51d992e06540ea87818ba18e9f6bb1>:0 
                                   at Exiled.CustomItems.API.Features.CustomGrenade.Check (Exiled.API.Features.Pickups.Projectiles.Projectile grenade) [0x00006] in <7fa5030fac1d45f18b2b2dd720e50f12>:0 
                                   at Exiled.CustomItems.API.Features.CustomGrenade.OnInternalExplodingGrenade (Exiled.Events.EventArgs.Map.ExplodingGrenadeEventArgs ev) [0x00007] in <7fa5030fac1d45f18b2b2dd720e50f12>:0 
[...]

PR #1742, opened 2023-04-20
Fix For OnInternalExplodingGrenade & OnInternalThrowingRequest & CustomItems

xNexus-ACS:
Not tested, so I mark this as draft for now


xNexus-ACS:
#1739

Issue #1757, opened 2023-04-25
[BUG] RespawningTeam Event


Misaka-ZeroTwo:
**Describe the bug**
ev.Players returns all spawnable players.

**Expected behavior**
It should return players that will be spawned

**EXILED Version ("latest" is not a version):**
6.1.0-rc.1

Reported in https://discord.com/channels/656673194693885975/1100509030683791480

PR #1758, opened 2023-04-25
Respawing team event fix

VALERA771:
Returned wrong value because patch got list before removing values.

Fix for #1757


Misaka-ZeroTwo (changes requested):
You also need to handle the logic which when ``ev.MaximumRespawnAmount`` got set to a lower value, ``ev.Players`` needs to be adjusted.

Misaka-ZeroTwo (commented):
accidentally added a file?

VALERA771:
Oops

Misaka-ZeroTwo:
also it still doesn't solve the problem, because ev.Players only gets adjusted when the eventargs is constructed.

VALERA771:
Okay, now should be fine

VALERA771:
> I feel like it's better to just change the setter in the eventarg instead. cuz that way ev.Players will be the same.

Okay but I don't get what is wrong. ev.Players will be same but list will be adjusted anyway

Misaka-ZeroTwo:
For example
```
ev.MaximumRespawnAmount = 10; // smaller than current ev.Players
List<Player> spawnablePlayers = ev.Players; // this won't work if you are changing afterwards
```

VALERA771:
> For example
> 
> ```
> 
> ev.MaximumRespawnAmount = 10; // smaller than current ev.Players number
> 
> List<Player> spawnablePlayers = ev.Players; // this won't work if you are changing afterwards
> 
> ```

In this case yeah. And it'll be easier

VALERA771:
Tested

VALERA771:
Small thing, I think requires-testing label should be removed

Exiled.Events/EventArgs/Server/RespawningTeamEventArgs.cs
```diff
@@ -58,7 +59,22 @@ public RespawningTeamEventArgs(List<Player> players, int maxRespawn, SpawnableTe
...
         ///     Gets or sets the maximum amount of respawnable players.
         /// </summary>
-        public int MaximumRespawnAmount { get; set; }
+        public int MaximumRespawnAmount
+        {
+            get => maximumRespawnAmount;
+            set
+            {
+                if (value == maximumRespawnAmount)
+                    return;
```
louis1706, line 68: not required

louis1706 (changes requested):

PR #1761, opened 2023-04-27
Fixed Shot event for non buckshot guns

### LINKED ISSUE #1745: [BUG] Shot event doesn't fire
**Describe the bug**
Shot event doesn't fire if you don't hit a player

**Expected behavior**
Event should fire.

Reported by @TtroubleTT

Misaka-ZeroTwo:
Partially fixed #1745


Misaka-ZeroTwo:
The original patch for Shot is inserted after retrieving damage. so I inserted one right before PlaceBulletholeDecal.

image

louis1706 (approved):

PR #1765, opened 2023-05-03
`Map.StartDecontamination` does not work with `disable_decontamination` true

Ficus-x:
When `disable_decontamination` is true, you cant force decontamination. The fix replaces methods which ignore this.
Tested

### DIFF
--- a/Exiled.API/Features/Map.cs
+++ b/Exiled.API/Features/Map.cs
@@ -204,11 +204,7 @@ namespace Exiled.API.Features
         /// <summary>
         /// Starts the light containment zone decontamination process.
         /// </summary>
-        public static void StartDecontamination()
-        {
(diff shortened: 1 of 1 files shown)


louis1706:
Test confirmed by Fircurse

IRacle1 (approved):

Misaka-ZeroTwo (approved):

joker-119:
Imagine, disabling decontamination on ur server actually stops decontamination from happening....

louis1706 (approved):

louis1706:
> Imagine, disabling decontamination on ur server actually stops decontamination from happening....

Yes but at least Exiled should overide this and not looking about server config (when possible) since that will trigger by a dev

PR #1776, opened 2023-05-10
Fix remove cuff event

### LINKED ISSUE #1772: [BUG] RemovingHandcuffs event not firing
**Describe the bug**
Event doesn't fire while uncuffing

Misaka-ZeroTwo:
Fix #1772


Misaka-ZeroTwo:
> Why we need this check?

which check?

Misaka-ZeroTwo:
> For keycard and bypass mod in patch I think NW can do this cuz I've also worked on this error but github didn't opened my PR. So when I removed that check and leaved NW (in patch was only event) everything also worked normal

This is uncuffing event. Not InteractingDoor event.

VALERA771:
Oh I ficked smth up. Sorry

Issue #1771, opened 2023-05-10
[BUG] Enable nuke kick


Misaka-ZeroTwo:
**Describe the bug**
Start the nuke with the bypass mode result kick.

**Expected behavior**
Player shouldn't get kicked.

**Server logs**
```
[2023-05-10 13:05:30.894 +01:00] [STDOUT] Disconnecting connId=50 to prevent exploits from an Exception in MessageHandler: NullReferenceException Object reference not set to an instance of an object.
[2023-05-10 13:05:30.894 +01:00] [STDOUT]   at AlphaWarheadOutsitePanel.set_NetworkkeycardEntered (System.Boolean value) [0x00000] in <8b78f4bd8dad4c1aa34c3b5fb085e2f8>:0
[2023-05-10 13:05:30.894 +01:00] [STDOUT]   at (wrapper dynamic-method) PlayerInteract.PlayerInteract.UserCode_CmdSwitchAWButton_Patch0(PlayerInteract)
[2023-05-10 13:05:30.894 +01:00] [STDOUT]   at PlayerInteract.InvokeUserCode_CmdSwitchAWButton (Mirror.NetworkBehaviour obj, Mirror.NetworkReader reader, Mirror.NetworkConnectionToClient senderConnection) [0x00016] in <8b78f4bd8dad4c1aa34c3b5fb085e2f8>:0
[2023-05-10 13:05:30.894 +01:00] [STDOUT]   at Mirror.RemoteCalls.RemoteProcedureCalls.Invoke (System.UInt16 functionHash, Mirror.RemoteCalls.RemoteCallType remoteCallType, Mirror.NetworkReader reader, Mirror.NetworkBehaviour component, Mirror.NetworkConnectionToClient senderConnection) [0x00019] in <4a8a79ac7a7545e9a4c54fc7c7d028a0>:0
[2023-05-10 13:05:30.894 +01:00] [STDOUT]   at Mirror.NetworkIdentity.HandleRemoteCall (System.Byte componentIndex, System.UInt16 functionHash, Mirror.RemoteCalls.RemoteCallType remoteCallType, Mirror.NetworkReader reader, Mirror.NetworkConnectionToClient senderConnection) [0x00065] in <4a8a79ac7a7545e9a4c54fc7c7d028a0>:0
[...]

PR #1778, opened 2023-05-10
Activate warhead ev fix

VALERA771:
Idk why there is long check for access but if remove it everything works fine!

Fix for #1771


Misaka-ZeroTwo (changes requested):
By placing it before ``OnInteract``, you are making the event not fire when player doesn't have sufficient permission.

joker-119:
This 'fix' removes a core feature of the event for the reason stated in Misaka's comment, please change this so that it doesn't remove functionality from the event to fix the issue.

VALERA771:
Also tested

Misaka-ZeroTwo (commented):
Why not insert after ``if (!this.ChckDis(player.Position)``?

VALERA771:
> Why not insert after `if (!this.ChckDis(player.Position)`?

Cuz there is some troubles with labels that I couldn't fix

Exiled.Events/Patches/Events/Player/ActivatingWarheadPanel.cs
```diff
@@ -34,90 +31,82 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
-            newInstructions[newInstructions.Count - 1].labels.Add(retLabel);
+            LocalBuilder player = generator.DeclareLocal(typeof(Player));
+            LocalBuilder allowed = generator.DeclareLocal(typeof(bool));
+            LocalBuilder card = generator.DeclareLocal(typeof(Keycard));
+
+            int index = newInstructions.FindIndex(i => i.Is(OpCodes.Ldfld, Field(typeof(PlayerInteract), nameof(PlayerInteract._sr))));
+
+            newInstructions.RemoveRange(index, 17);
+
+            newInstructions[index].labels.Add(retLabel);
```
louis1706, line 46: Rename it continueLabel or anything you want but not ret since it's will not going to a return direclty

louis1706 (changes requested):

louis1706 (approved):

PR #1780, opened 2023-05-10
New Explode Method

louis1706:


Exiled.API/Features/Player.cs
```diff
@@ -3167,6 +3167,12 @@ public bool TryGetComponent<T>(Type type, out T component)
         public void GetCooldownItem(float time, ItemType itemType)
             => UsableItemsController.GetHandler(ReferenceHub).PersonalCooldowns[itemType] = Time.timeSinceLevelLoad + time;
 
+        /// <summary>
+        /// Explode the player.
+        /// </summary>
+        /// <param name="attacker">player who create the explosion.</param>
+        public void Explode(Player attacker = null) => ExplosionUtils.ServerExplode(attacker?.ReferenceHub ?? Server.Host.ReferenceHub);
```
joker-119, line 3174: ServerExplode()'s ReferenceHub parameter is the player who is exploding. If it's a null value it will throw a nullref, therefore since this is an instanced method, it should either be Player target = null that defaults to the player the instanced method is called from, or should just be paramaterless.

joker-119 (changes requested):

Exiled.API/Features/Player.cs
louis1706: oh wait yes i am dumbdumb lol it's do not need an argument at all finally lol 
I got tinybrain moment

Thundermaker300 (approved):

Thundermaker300 (approved):

louis1706:
I miss to rename Explode2 method

IRacle1 (approved):

PR #1783, opened 2023-05-11
Better error message when same command has already registered

VALERA771:


Misaka-ZeroTwo (commented):
That's a lot iteration. Use catch instead

VALERA771:
But what exception should I catch?

Misaka-ZeroTwo:
```
                    catch (ArgumentException e)
                    {
                        if (e.Message.StartsWith("An"))
                        {
                            Log.Error($"Command with same name has already registered! Command: {command.Command}");
                        }
                        else
                        {
                            Log.Error($"An error has occurred while registering a command: {e}");
                        }
                    }
```
will do the trick

Misaka-ZeroTwo (approved):

Thundermaker300 (approved):

PR #1785, opened 2023-05-11
Scp079Scanner

louis1706:


Exiled.API/Extensions/EnumExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        /// <summary>
+        /// Converting a Bool Array to an Enum.
+        /// </summary>
+        /// <typeparam name="T">The returned object type.</typeparam>
+        /// <param name="arr">The array of bool to convert.</param>
+        /// <returns>The <typeparamref name="T"/> module that was requested.</returns>
+        public static T ToFlagEnum<T>(this bool[] arr)
+            where T : struct
```
Thundermaker300, line 42:
```suggestion
            where T : Enum
```

VALERA771:
Why we can't just use Scp079ScannerSelectorAbility::SelectedZones which are already FacilityZone?

louis1706:
> Scp079ScannerSelectorAbility::

cause we use our own Enum value :3

NaoUnderscore (approved):

PR #1787, opened 2023-05-12
[Feature] Interacting door event

### LINKED ISSUE #1733: [BUG] InteractingDoorEventArgs ev.IsAllowed not working
**Describe the bug**
If door is locked player can't open it, even if  in InteractingDoorEventArgs property ev.IsAllowed set  to true

Reported by @ITeMbI4 in https://discord.com/channels/656673194693885975/1079489116875522098/1079489116875522098

Misaka-ZeroTwo:
Reopen for the next Exiled/SL major update to prevent unnecessary breaking to plugins.
New feature: Instead of setting ``ev.IsAllowed = false`` which makes the door do nothing, a new parameter allows permission denied to be shown.
``ev.IsAllowed = true`` will need to be changed to ``ev.InteractionResult = DoorBeepType.InteractionAllowed``


Thundermaker300 (approved):

louis1706 (approved):

louis1706:
That will need some plugin to be rewriten to not use ev.IsAllowed = false

Misaka-ZeroTwo:
> That will need some plugin to be rewriten to not use ev.IsAllowed = false

That's up to the dev. It still functional the same as before(do nothing).

Misaka-ZeroTwo:
tested with setting different ``InteractionResult``, and ``IsAllowed``. Worked as expected.

louis1706 (approved):

joker-119:
Fix merge conflict.

Also shouldn't we just have IsAllowed setter change to denied beep for us, so the dev doesn't have to do 2 lines of code for this?

Misaka-ZeroTwo:
> Fix merge conflict.
> 
> 
> 
> Also shouldn't we just have IsAllowed setter change to denied beep for us, so the dev doesn't have to do 2 lines of code for this?

It makes more sense to me that ev.IsAllowed simply reject the event. However, there's no need for 2 lines for each cases. If they want to have denied beep, set ev.InteractionResult to PermissionDenied would do the trick.

Misaka-ZeroTwo:
Closing this PR cuz it's a breaking change, however, will reopen for next major version as it seem to be a better approach.

louis1706:
> Closing this PR cuz it's a breaking change, however, will reopen for next major version as it seem to be a better approach.

you can keep it open with DoNotMerge.

Exiled.Events/Patches/Events/Player/InteractingDoor.cs
```diff
@@ -34,91 +45,127 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
...
 
             ListPool<CodeInstruction>.Pool.Return(newInstructions);
         }
 
-        private static bool CanStateChange(DoorVariant variant)
+        private static void Process(DoorVariant __instance, ReferenceHub ply, byte colliderId, InteractingDoorEventArgs ev)
         {
-            return !(variant.GetExactState() > 0f && variant.GetExactState() < 1f);
+            Handlers.Player.OnInteractingDoor(ev);
+            if (ev.IsAllowed)
```
louis1706, line 152: i am really not sure about this because mostly every plugin use this to just DeniedPermission

louis1706: and even in a breaking change update it's will be confusing for people

PR #1793, opened 2023-05-13
Remove non-existent doortype

Misaka-ZeroTwo:


Misaka-ZeroTwo:
Better merge it later since it will require all plugins using DoorType to recompile.

Thundermaker300:
You could obsolete the enum item instead of removing it. Then we can merge it now, and remove it later when we remove every obsolete member.

Misaka-ZeroTwo:
Duplicate with #1723 , closing

Issue #1749, opened 2023-04-21
[BUG] PreAuthenticating event does not fire when connecting to a server with whitelist enabled


Misaka-ZeroTwo:
**Describe the bug**
Per title

Reported by: @Ficus-x

joker-119:
While this is P1, the only way to fix it would be by re-adding a preauth patch for this event. Idk why it was removed in the first place but there's a lot that could go wrong with doing so, thus this P1 will not be considered blocking.

Misaka-ZeroTwo:
Re-open due to the fix being broken.

louis1706:
I will look at this after the next Release of NW since they will entirely rewrite this method

ALEXWARELLC:
@Misaka-ZeroTwo Hey!

After doing some testing, it seems `PreAuthenticating` is working with Whitelist enabled, so feel free to close this issue if you wish.

Thanks :)

PR #1799, opened 2023-05-15
Pre auth event transpiler

VALERA771:
Also I've tested that It's calling but cuz it's for #1749 requiers test in that case


louis1706 (approved):

VALERA771:
Tested in issue case

PR #1804, opened 2023-05-16
Better ExiledLoading

louis1706:
- why is better ?
 - because Exiled.API don't need to be In NW Dependency 
 
 
 (for testing make sure to verify than config is not reset or overwrited and if config is loaded correctly)


louis1706:
Not possible cause of NW checking everything inside the plugin

PR #1808, opened 2023-05-17
Make all texts in config be quoted

VALERA771:
Time has come


louis1706:
I would like to know if it's will break previous config ^^

VALERA771:
Now I have problem that it's also makes properties be quoted. Not only values. So now I'm trying to fix this and then check that

VALERA771:
Confirm that won't break configs

VALERA771:
Alsp tested

louis1706 (changes requested):
Fix Build Error

louis1706:
> > Fix Build Error
> 
> What build error? All checks passed

yup sorry certainly miss refresh moment

louis1706 (approved):

PR #1815, opened 2023-05-21
Allow editing tickets in Escaping event

### LINKED ISSUE #1810: [BUG] Escape event
How to reproduce: 
Be a class (in my instance a special scientist class) that on escape method isnt allowed to escape. If it goes into escape zone it gives all the tickets to their team. In my case mtf. 

Expected outcome: 
Does not give tickets if they cant actually escape. 

Video:  https://medal.tv/games/scp-secret-laboratory/clips/19BYbWhNbp1Vnv/d13373BlLEy9?invite=cr-MSwzRkssNjYyODkyMjgs

4310V343k:
also fixes #1810


Misaka-ZeroTwo:
Should be getting tickets, not editing tickets.

Exiled.Events/EventArgs/Player/EscapingEventArgs.cs
```diff
@@ -53,6 +87,11 @@ public EscapingEventArgs(Player player, RoleTypeId newRole, EscapeScenario escap
         /// </summary>
         public EscapeScenario EscapeScenario { get; set; }
 
+        /// <summary>
+        ///     Gets the TicketsToChange that will represent the amount of tickets granted after the player escapes.
+        /// </summary>
+        public Dictionary<SpawnableTeamType, float> TicketsToChange { get; }
```
Misaka-ZeroTwo, line 93: That's get only.

Misaka-ZeroTwo: Oh ok, then just do `get; set;`? Because that indicates the property can be edited.

louis1706 (changes requested):
also there no need to make a Dictionary and new argument for the ctor

louis1706:
this PR also conflict with #1807  (that not a requested change it's only informative)

louis1706 (approved):

VALERA771:
Why you've created 2 new constructors?

PR #1820, opened 2023-05-23
Scp914 Transpiler

louis1706:


louis1706:
> Tested. No errors on startup

i will need to tested it directly in game ^^

PR #1818, opened 2023-05-22
Event stuff

VALERA771:
* Convert some prefixes to transpilers cuz they're using NW event system which'll be updated in 13.0 NWAPI
* Adds Scanning event for 079

I'm also planning to convert some of prefixes with TODO to transpilers. _P.S. Consuming won't be reworked cuz I've already done this in #1720_

PR #1822, opened 2023-05-23
Kicking event prefix to transpiler

Misaka-ZeroTwo:


VALERA771:
I've alreasy done it in #1818

PR #1824, opened 2023-05-23
Scp049 Activating Sense Transpiler Reworked

louis1706:
Fix than people can't use Sense ability when a tutorial is near to him


VALERA771:
#1818 🙃

joker-119:
Does this change conflict with #1818 at all, cuz that ones already been merged.

VALERA771:
> Does this change conflict with #1818 at all, cuz that ones already been merged.

No conflicts. I've removed my rework

PR #1839, opened 2023-05-29
Exiled.Loader 6.9 & No more alias

louis1706:


VALERA771:
Exiled 6.9 has a chance to exist?

louis1706:
Technically 7.6.9 could exist

VALERA771:
Huh. Really

joker-119:
Why are we disabling aliases?

louis1706:
> Why are we disabling aliases?

Because it's just horrible
 List<object>
 - &0e1
 - 0e1

joker-119:
The change to the serializer also needs to be duplicated here: https://github.com/Exiled-Team/EXILED/blob/dev/Exiled.CustomRoles/CustomRoles.cs#L35

joker-119 (approved):

Misaka-ZeroTwo (approved):

PR #1845, opened 2023-06-01
Fix Player.IsConnected property.

o5zereth:


o5zereth:
> GameObject != null should work it's would be an issue only if it's will be GameObject is not null

Did you read what I just wrote?

Misaka-ZeroTwo (commented):
It still doesn't match with the docs. Player that joined but not yet verified is still true.

o5zereth:
Connected and verified are not the same thing. Connected means connected. Verified means verified. Theres a difference.

o5zereth:
A valid way to check for this fix would be something like this.

```cs
void PlayerLeft(PlayerLeftEventArgs ev)
{
    Timing.CallDelayed(1f, () =>
    {
        Log.Debug(ev.Player.IsConnected);
    }
}
```

PR #1863, opened 2023-06-12
Fix Player.Get(ServerHostPlayer)

louis1706:
Now Server.Host can trigger event and will not be null when attacking for example
(SpawnActive Grenade with Server.Host for example)


Michal78900 (approved):

joker-119:
This in itself is not technically a *breaking* change, however it will change expected behavior for plugins. Delaying merge for 7.1

louis1706:
> Now Server.Host can trigger event and will not be null when attacking for example (SpawnActive Grenade with Server.Host for example)

yeah that why i have make little precision in the description

louis1706:
Step to reproduce 
- use SR
- Forceclass yourself to any role but don't start the round
- Disconect
- Get Spammed of LocalAdmingLogs error

louis1706:
Do not merge actually causing bug

louis1706:
it's seems than the bug is gone with one of the NPC fix but i am actually not sure if there more bug with it

louis1706:
I have tested it today on my server it's seems fine now

PR #1870, opened 2023-06-15
Reverted back to original method of finding player by int as new meth…

NotIntense:
Reverted back to original method of finding player by int as new method uses basegame needs players to "Request Data" before trying to find by player ID


Misaka-ZeroTwo (changes requested):
need adjustments on other parts too, it won't build.

PR #1871, opened 2023-06-15
Reverted PlayerGet(ID)

NotIntense:
Reverted to the original method of getting players by ID


NotIntense:
omfg it didnt bring the actual thing i needed

NotIntense:
hold on

PR #1873, opened 2023-06-15
Reverted method of getting players by their ID 

NotIntense:
Reverted : https://github.com/Exiled-Team/EXILED/pull/1872

### DIFF
--- a/Exiled.Events/Patches/Events/Player/Joined.cs
+++ b/Exiled.Events/Patches/Events/Player/Joined.cs
@@ -47,6 +47,7 @@ namespace Exiled.Events.Patches.Events.Player
                 }
                 else
                 {
+                    Player.UnverifiedPlayers.Add(hub, player);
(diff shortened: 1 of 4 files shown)


Exiled.Events/Patches/Events/Player/Joined.cs
```diff
@@ -47,6 +47,7 @@ internal static void CallEvent(ReferenceHub hub, out Player player)
                 }
                 else
                 {
+                    Player.UnverifiedPlayers.Add(hub, player);
```
joker-119, line 50: Will bork things.

NotIntense: how so?

joker-119: cuz UnverifiedPlayers is a Dictionary<GameObject, Player> not <ReferenceHub, Player> now, and I don't see anywhere that's reverted that change.

PR #1877, opened 2023-06-16
Fix Scp049 Attack

louis1706:


Misaka-ZeroTwo (commented):
Add a bool in param for using sense ability or not?

louis1706:
> Add a bool in param for using sense ability or not?

That a break change

Thundermaker300 (approved):

PR #1878, opened 2023-06-16
If you merge this now you are really dumb :trollface: 

louis1706:
Need to be split in multiple PR

- [ ]  IDamageEvent / ITargetEvent / IUsableEvent (with adding them to every event where it's possible without break)
- [ ]  adding FlippingCoinEventArgs::Item 
- [ ]  adding ChangingRadioPresetEventArgs::Radio and ChangingRadioPresetEventArgs::Item
- [ ]  use  ``/// <inheritdoc />`` everywhere it's possible

breaking change or required Obsolete
-  [ ]  DamageHandler should use CustomDamageHandler
- [ ]  Remove every Inherited Event by an other Event


Thundermaker300:
im merging this now

PR #1903, opened 2023-06-25
funny fix

IRacle1:
- Fix for spamming `ReceivingStatusEffect` event
- Fix for compiling docs project???


louis1706 (approved):

louis1706 (approved):

louis1706:
@joker-119

joker-119:
Is this even a verified working change? What if something in the game or a plugin uses Intensity = 1 instead of ForceIntensity to change it? We should instead just see if value = curIntensity then return

PR #1914, opened 2023-06-28
Projectile fixes

IRacle1:
Fixes `Projectile::Create/CreateAndSpawn` methods.
Problem was in difference between projectile and normal pickups, meaning that grenade pickup(when you drop out grenade from your inventory) and grenade projectiles(when you throw grenade) is completely different, `Projectile::Create` method created normal grenade pickups and tried to cast it to projectiles.

### About many `activeSelf` checks
Projectiles prefabs are active by default, and if you instantiate then without turning off gameobject state, they will be collide and do things with something(it is bad for example for Scp018 and Scp2176), that logic has been exist in ThrowableItem for a long time

I tested it with that code 
```cs
ExplosionGrenadeProjectile pj = (ExplosionGrenadeProjectile)Projectile.CreateAndSpawn(ProjectileType.FragGrenade, ev.Player.Position, Quaternion.identity, false, ev.Player);
Pickup pickup = Pickup.CreateAndSpawn(ItemType.KeycardFacilityManager, ev.Player.Position, Quaternion.identity, ev.Player);

Timing.CallDelayed(5f, () =>
{
    pj.FuseTime = 2f;
    pj.IsActive = true;
});
```
and it worked perfectly, for more detailed debug ill check it later


louis1706 (approved):
Reminder for people it's shouldn't be needed to put delay

Exiled.API/Features/Pickups/Projectiles/Projectile.cs
```diff
@@ -35,9 +38,23 @@ internal Projectile(ThrownProjectile pickupBase)
         /// </summary>
         /// <param name="type">The <see cref="ItemType"/> of the pickup.</param>
         internal Projectile(ItemType type)
-            : base(type)
         {
-            Base = (ThrownProjectile)((Pickup)this).Base;
+            if (!InventoryItemLoader.AvailableItems.TryGetValue(type, out ItemBase itemBase) || itemBase is not ThrowableItem throwable)
+                return;
+
+            throwable.Projectile.gameObject.SetActive(false);
```
Misaka-ZeroTwo, line 45: Why do you SetActive to false and then true?

Misaka-ZeroTwo (approved):

louis1706 (approved):

louis1706 (approved):

PR #1915, opened 2023-06-29
Removed option to set NPC ID

NotIntense:
Removed the option to set the ID, and is now determined by player count and NPC count. Server handles it and can break when trying to set it manually

### DIFF
--- a/Exiled.API/Features/Npc.cs
+++ b/Exiled.API/Features/Npc.cs
@@ -138,11 +138,10 @@ namespace Exiled.API.Features
         /// </summary>
         /// <param name="name">The name of the NPC.</param>
         /// <param name="role">The role type ID of the NPC.</param>
-        /// <param name="id">The ID of the NPC.</param>
         /// <param name="userId">The user ID of the NPC.</param>
         /// <param name="position">The position to spawn the NPC.</param>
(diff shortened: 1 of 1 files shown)


Exiled.API/Features/Npc.cs
```diff
@@ -156,7 +155,7 @@ public static Npc Spawn(string name, RoleTypeId role, int id = 0, string userId
             }
 
             ReferenceHub hubPlayer = npc.ReferenceHub;
-            FakeConnection fakeConnection = new(id);
+            FakeConnection fakeConnection = new(Player.List.Count + Npc.List.Count);
```
Misaka-ZeroTwo, line 158: That would cause duplicate ids.

Misaka-ZeroTwo (changes requested):

Exiled.API/Features/Npc.cs
Misaka-ZeroTwo: Also breaking change

louis1706:
I closed the PR this is clearly not how it should be handled to add id for npc automatically

joker-119:
Fixed without breaking changes in [ee8becb](https://github.com/Exiled-Team/EXILED/commit/ee8becbbfaac397d9fde071711a748e54b55ac58)

PR #1916, opened 2023-06-29
Fix funny gate

Misaka-ZeroTwo:


Misaka-ZeroTwo:
> Seems strange about removing this part of code but since you tested it i will approved

Basically, this part is base-game checking is the door ready for another Interaction. We shouldn't allow Door to still be able to interact, otherwise, it would cause some funny things like https://discord.com/channels/656673194693885975/1123768569977643121/1123768569977643121

IRacle1 (approved):

PR #1923, opened 2023-06-30
Fix Door api for Gate

### LINKED ISSUE #1919: [BUG] Door.IsMoving does not work for the gate
**Describe the bug**
If Door is a gate, then Door.IsMoving is always false. 

**To Reproduce**
By way of example
`private void OnDoorInteract(InteractingDoorEventArgs ev)
{
    if (ev.Door.IsGate)
        Log.Info(ev.Door.IsMoving);
}`

**EXILED Version 7.1.0**

Misaka-ZeroTwo:
Fix #1919 

Gate doesn't have an implementation on ``GetExactState()``. The only way to tell if it's still moving is through its animation CD.


R2kip (approved):
thx

IRacle1 (approved):

Thundermaker300 (approved):

PR #1933, opened 2023-07-02
CustomRole::RegisterRoles(bool, object?, bool, Assembly?) fix

VALERA771:
inhernitAttributes now have a reason to exist


joker-119:
The completely breaks using the attribute at all if you set inheritAttributes to false.

VALERA771:
For me easier to override property instead of using attribute
So I thought that inheritAttributes parameter indicates should class has a CustomRole attribute

joker-119:
Yea that's not at all what it's for. It's for use in the reflection, to determine if the attribute should be inherited if you are making an interface class between the api and your custom roles.

VALERA771:
Huh. Ty

Misaka-ZeroTwo:
build error

VALERA771:
I think this PR should have enchantment label

NaoUnderscore:
I don't really get the point of this PR

VALERA771:
1. Allowing to register CustomRoles which define Type without attribute
2. Breaking changes. In CustomItems we have only 2 register methods and I don't see any reasons why we have here so much

VALERA771:
Closing cuz CR API will be reworked

PR #1938, opened 2023-07-03
RestartNextRound when Exiled update

louis1706:
wait for to be merged #1941


louis1706:
Would like someone to test it on other server than windows

Exiled.Updater/Updater.cs
```diff
@@ -261,88 +292,94 @@ where name.StartsWith("Exiled.", StringComparison.OrdinalIgnoreCase)
...
         #endregion
 
-        private void Update(HttpClient client, NewVersion newVersion)
+        private async Task Update(HttpClient client, NewVersion newVersion)
         {
             try
             {
                 Log.Info("Downloading installer...");
-                using (HttpResponseMessage installer = client.GetAsync(newVersion.Asset.BrowserDownloadUrl).ConfigureAwait(false).GetAwaiter().GetResult())
+                HttpResponseMessage installerResponse = client.GetAsync(newVersion.Asset.BrowserDownloadUrl).ConfigureAwait(false).GetAwaiter().GetResult();
```
louis1706, line 300: I didn't find how to make this one await without error

joker-119: ok but HttpResponseMessage is an IDisposable so where's my using directive so it's disposed properly?

And what was the error when using a normal await?

louis1706:
Let's rewrite this PR from 0 and doing it better

PR #1940, opened 2023-07-04
Allow for spawning Unverified NPCs (VSR Compliance stuff with NPC API)

creepycats:
Allows NPCs to be created without a UserID by leaving it blank instead of replacing it with `Dummy@localhost`, in the Unverified ClientInstanceMode. 

Should fix VSR compliance by removing the issue of NPCs counting as real players and taking up slots unless I'm a stupid goober.


louis1706:
#1944 will fix your issue

creepycats:
> #1944 will fix your issue
Oki, I'll prob close it since I don't know if it would be useful to have Unverified NPCs in that case.

creepycats:
Update: I just tested your changes ingame. Even if it might fix the issue of VSR, I'd like to see unverified NPCs be a feature. 

The NPCs using your method and the current method both display these NPCs on the RemoteAdmin and the Player list.
image
image
Also, it seems like player IDs get shifted over a little but I might just be stupid.

Using Unverified NPCs would allow server hosts and plugin devs to hide the NPCs from their lists as well as from certain events that might conflict. It basically just makes it look less inflated and also more easy to manage. Imagine having a plugin that spawns one NPC for each player. It would NOT be fun to manage with RemoteAdmin.

louis1706:
> Update: I just tested your changes ingame. Even if it might fix the issue of VSR, I'd like to see unverified NPCs be a feature.
> 
> The NPCs using your method and the current method both display these NPCs on the RemoteAdmin and the Player list.

can i contact you trough discord so we can talk easyer

creepycats:
Yea, `creepycats` is my tag. DM me and tell me where I'm wrong.

louis1706:
No Activity on this PR for more than 4 Month (No fix about the build)

PR #1949, opened 2023-07-05
Loader

Misaka-ZeroTwo:
Exiled.API is loaded from NW API dep. And NW API has such a check, so the check should be deleted.


louis1706:
I actually don't want this PR Merged since it's just put Exiled.API in Exiled.Loader and don't fixed the probleme than we still need to update two different file

Misaka-ZeroTwo:
> I actually don't want this PR Merged since it's just put Exiled.API in Exiled.Loader and don't fixed the probleme than we still need to update two different file

No, I don't think it's needed to put Exiled.API in exiled dep.

PR #1952, opened 2023-07-05
LittleChange

louis1706:
- Fix than when Item.Base is null it's don't work


joker-119:
How tf would Item.Base be null when Item itself is not and the Base is populated in the Item ctor?

louis1706:
> How tf would Item.Base be null when Item itself is not and the Base is populated in the Item ctor?

Don't ask i am actually don't getting it but someone told me than it's was working better bruh and would make no issue to having it like that

joker-119:
The only way I see this happening is if someone specifically stores an item before removing it from another player, then tries to add it to another player, instead of using Clone() which will result in bad behavior (IE taking a gun that's only got 1 round and making a new one with full clip), or if they very specifically bulled the Item.Base property.

In this case, I'm not really sure we **want** to implement this 'fix', as it doesn't really fix issues with the most likely scenario that would cause it, it just shifts the issue from being a clear exception saying something is wrong, to a weird bug that could take days to track down.

PR #1953, opened 2023-07-06
Public secured PatchAll

louis1706:


joker-119:
Ok but like... Why?

louis1706:
> Ok but like... Why?

Why other plugin shouldn't have access to this ?

louis1706:
> This should use GetCallingAssembly() not GetExecutingAssembly()

it's not the point of this PR if you want you can make an other PR

o5zereth:
> > This should use GetCallingAssembly() not GetExecutingAssembly()
> 
> it's not the point of this PR if you want you can make an other PR

Uh, if you use GetExecutingAssembly your commit is completely useless, and will only patch the current assembly.

It quite literally requires the caller to have any point in being public.

My point is, your PR is wrong. You fix it, not me.

Thundermaker300 (approved):

NaoUnderscore (changes requested):
Absolutely not.
`GlobalPatchProcessor` must not be marked as `static`, you must be able to inherit from it.

NaoUnderscore (approved):

PR #1954, opened 2023-07-06
DynamicPatching

louis1706:
The main purpose of this pr is to implement a new ```Exiled.Events``` feature, patching per event. Basically, event patches will only be patched if there are methods subscribed to the event they target. I think this is a necessary feature to lessen the performance and stability issues of requiring the many unused patches exiled has.

How this works is, all existing ```event```s in the ```Handlers``` namespace will be replaced with a wrapper ```Event``` (or ```Event<T>```), which handles subscribing methods in relatively the same way, to lessen the breaking changes. 

The difference is, on method subscription, the ```Patcher``` class will look for any patches that target that event, and patch ONLY them.

The main purpose of this pr is to implement a new ```Exiled.Events``` feature, patching per event. Basically, event patches will only be patched if there are methods subscribed to the event they target. I think this is a necessary feature to lessen the performance and stability issues of requiring the many unused patches exiled has.

~~How I envision it currently:~~
How it works currently:
```cs
public static class Player //handler class
{
    public static Event<ShootingEventArgs> Shooting { get; set; } = new(); // Property required for += op
    //other handlers here
}

[EventPatch(typeof(Player), nameof(Player.Shooting))] //Patcher class will look for this
[HarmonyPatch( /*same thing here*/ )]
public static class Shooting //patch
{
    private IEnumerator<float> Transpiler() {} //Patch functions being the same
}
```
This is configurable too, so you can choose to patch the old way as well, for debugging patches and such.

Any feedback or if this is completely useless for some reason, please lmk


joker-119:
This will be a breaking change. Marking as DNM until 8.x

Exiled.Events/Features/Event.cs
```diff
@@ -0,0 +1,107 @@
...
+                InnerEvent += handler;
+            }
+        }
+
+        /// <summary>
+        /// Unsubscribes a target <see cref="CustomEventHandler"/> from the inner event if the conditional is true.
+        /// </summary>
+        /// <param name="handler">The handler to add.</param>
+        /// <param name="shouldUnsubscribe">Whether the handler should be subscribed or not.</param>
+        public void Unsubscribe(CustomEventHandler handler, bool shouldUnsubscribe = true)
```
joker-119, line 82: Same question as above

Exiled.Events/Features/Patcher.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            failedPatch = 0;
+
+            try
+            {
+#if DEBUG
+                bool lastDebugStatus = Harmony.DEBUG;
+                Harmony.DEBUG = true;
+#endif
+                List<Type> toRemove = new();
```
joker-119, line 86: Should use a ListPool here

Exiled.Events/Features/Patcher.cs
```diff
@@ -0,0 +1,136 @@
...
+                IEnumerable<Type> toPatch = includeEvents ? UnpatchedTypes : UnpatchedTypes.Where((type) => !type.GetCustomAttributes<EventPatchAttribute>().Any());
+                foreach (Type patch in toPatch)
+                {
+                    try
+                    {
+                        Harmony.CreateClassProcessor(patch).Patch();
+                        toRemove.Add(patch);
```
joker-119, line 93: You populate this list here, then never do anything with it.

Exiled.Events/Handlers/Item.cs
```diff
@@ -9,44 +9,42 @@ namespace Exiled.Events.Handlers
...
 
-    using Extensions;
-
-    using static Events;
+    using Exiled.Events.Features;
 
     /// <summary>
     ///     Item related events.
     /// </summary>
     public static class Item
```
joker-119, line 17: The doc summaries for all of the events should be changed. Either by omitting the `Gets or sets` text, or by changing it to something that makes sense.
This would be a good example of selectively disabling pragma warnings, as the `gets or sets` text is pretty much irrelevant here.

Exiled.Events/Features/Patcher.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        /// <summary>
+        /// Patches all events that target a specific <see cref="IExiledEvent"/>.
+        /// </summary>
+        /// <param name="event">The <see cref="IExiledEvent"/> all matching patches should target.</param>
+        public void Patch(IExiledEvent @event)
+        {
+            try
+            {
+                List<Type> types = types = UnpatchedTypes.Where(x => x.GetCustomAttributes<EventPatchAttribute>().Any((epa) => epa.Event == @event)).ToList();
```
joker-119, line 56: Should use ListPool here.

joker-119 (changes requested):
Not horrible, a few small things need to be addressed, though.

VALERA771:
What about patches that add more than one event?

louis1706:
> What about patches that add more than one event?

You can already see them

Exiled.Events/Patches/Events/Map/ChangingIntoGrenade.cs
```diff
@@ -32,6 +32,8 @@ namespace Exiled.Events.Patches.Events.Map
     ///     Patches <see cref="TimedGrenadePickup.Update" />.
     ///     Adds the <see cref="Map.ChangingIntoGrenade" /> and <see cref="Map.ChangedIntoGrenade" /> events.
     /// </summary>
+    [EventPatch(typeof(Map), nameof(Map.ChangingIntoGrenade))]
+    [EventPatch(typeof(Map), nameof(Map.ChangedIntoGrenade))]
```
louis1706, line 36: @Misaka-ZeroTwo

NaoUnderscore (changes requested):
Before reviewing this PR, I'd like to let y'all know that a global patch processor already exists, although, due to breaking many breaking changes it hasn't been implementend entirely; however, the full implementation of it, should be able to handle events and patches as this dynamic patch system.

Not sure about the `Event` and `Event<T>` classes.
If we want to proceed in this way, I'd say to make our own multicast delegate/event implementation, but that's up to you to be honest.

We're basically wrapping as more things as possible to make everything easier, not sure whether it's a wise choice or not.

That being said, since we're gonna break everything, let's break more in order to add a proper support.
I'd like to make it as more versatile as possible in order to, eventually, make it easy to integrate more features and/or changes/edits.

What's currently missing is what the global patch processor, or its full implementation which we've never seen on Exiled due to the reasons I described before, handles or how it handles.

Firstly, let's make it compatible with patch groups, so that people can make their own patches better organized, such as if they were using folders.
Patch groups are useful to get, for example, all the patches in a specific range or context, like a plugin or, even more, Exiled itself (code sections).
I'd avoid relying on the name of the type, instead, I'd allow the user to display-name it as they prefer.

Exiled.Events/Commands/Show/Patches.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <inheritdoc/>
+        public string[] Aliases { get; } = { "patched" };
+
+        /// <inheritdoc/>
+        public string Description { get; } = "Returns information about all patches (whether they are patched or not)";
+
+        /// <inheritdoc/>
+        public bool Execute(ArraySegment<string> arguments, ICommandSender sender, out string response)
+        {
+            const string perm = "ee.showpatches";
```
Ficus-x, line 40: should be in class not method

Exiled.Events/Handlers/Item.cs
```diff
@@ -21,32 +21,32 @@ public static class Item
...
         ///     Invoked before item attachments are changed.
         /// </summary>
-        public static event CustomEventHandler<ChangingAttachmentsEventArgs> ChangingAttachments;
+        public static Event<ChangingAttachmentsEventArgs> ChangingAttachments { get; set; } = new();
 
         /// <summary>
         ///     Invoked before receiving a preference.
         /// </summary>
-        public static event CustomEventHandler<ReceivingPreferenceEventArgs> ReceivingPreference;
+        public static Event<ReceivingPreferenceEventArgs> ReceivingPreference { get; set; } = new();
```
Ficus-x, line 34: Isnt a breaking changes?

Exiled.Events/Features/Patcher.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// Patches all events that target a specific <see cref="IExiledEvent"/>.
+        /// </summary>
+        /// <param name="event">The <see cref="IExiledEvent"/> all matching patches should target.</param>
+        public void Patch(IExiledEvent @event)
+        {
+            try
+            {
+                List<Type> types = ListPool<Type>.Pool.Get(UnpatchedTypes.Where(x => x.GetCustomAttributes<EventPatchAttribute>().Any((epa) => epa.Event == @event)));
+
+                for (int i = 0; i < types.Count; i++)
```
Ficus-x, line 60: foreach

Ficus-x (changes requested):

Exiled.Events/Handlers/Item.cs
Ficus-x: I know, you added overload for + but anyways

louis1706: I don't care about breaking change

Exiled.Events/Commands/Show/Patches.cs
louis1706: It's will change nothing

louis1706:
I am actually using it on my server (has been test with DevTools) and other plugin of my server

NaoUnderscore (changes requested):
Code itself is fine, however, add a description about your changes.
I'd like to know, mainly for better organize the changelog and the API changes as well, what does this do, how it does it and why it should be merged.

louis1706:
image
❤️

NaoUnderscore (approved):

PR #1965, opened 2023-07-21
Update Npc's userid

skyyt15:
simply add a serial number for dummy to change the userid of an npc if it's default. like, the userId of an Npc or other Npc will be different


louis1706 (changes requested):
Fix Build Error

louis1706:
3 Month with no response

PR #1971, opened 2023-07-27
Fix Error

louis1706:
image


Misaka-ZeroTwo:
tf, why is it null

louis1706:
> tf, why is it null

There have check for creating the ctor of the event i show u on discord if u want

PR #1974, opened 2023-07-28
FixRunCommand

louis1706:


joker-119:
Why would we want to pass a null sender instead of Server.Host ?

louis1706:
> Why would we want to pass a null sender instead of Server.Host ?

image
simple cause Server.Host is not the same than ServerConsole.Scs and Server.Host don't have permission

PR #1984, opened 2023-08-09
Scp106ReworkTranspiller

louis1706:


Exiled.Events/EventArgs/Scp106/StalkingEventArgs.cs
```diff
@@ -22,26 +24,31 @@ public class StalkingEventArgs : IPlayerEvent, IDeniableEvent
...
         /// </summary>
-        public Scp106StalkAbility Scp106StalkAbility { get; }
+        [Obsolete("Use Scp106.StalkAbility instead of this")]
+        public Scp106StalkAbility Scp106StalkAbility => Scp106.StalkAbility;
 
         /// <summary>
         /// Gets or sets the current vigor when SCP-106 starts to stalk.
         /// </summary>
-        public float Vigor { get; set; }
+        public float Vigor
```
VALERA771, line 45: Why don't put obsolete here?

NaoUnderscore: This PR refers to a breaking change.

VALERA771: I'm about why not put obsolete exactly to property and not to getter and setter

louis1706: firstly is not a BreakingChange since people can still use it
secondly is because get and set are different 'method'

louis1706:
Don't merge for Exiled 8.0

Misaka-ZeroTwo (approved):

PR #1992, opened 2023-08-10
DroppingScp330Event Fix Candy

louis1706:


louis1706:
No idea how to fix this PR actually causing crash player when it's drop it's item

PR #1999, opened 2023-08-13
Revert DynalmicPatching Only

louis1706:


VALERA771 (approved):
Imagine release without this PR

PR #2003, opened 2023-08-17
Fix pickup bug + damage type.scp0492

louis1706:


Exiled.Events/Patches/Generic/InventoryControlPatch.cs
```diff
@@ -77,10 +77,10 @@ private static void AddItem(Player player, ItemBase itemBase, ItemPickupBase ite
         {
             Item item = Item.Get(itemBase);
             Pickup pickup = Pickup.Get(itemPickupBase);
-
+            /*
```
Misaka-ZeroTwo, line 80: Make a comment on why you commented on this part instead of removing it?

Exiled.Events/Patches/Generic/PickupControlPatch.cs
```diff
@@ -47,6 +47,7 @@ internal static class PickupControlPatch
 
             newInstructions.InsertRange(index, new CodeInstruction[]
             {
+                /*
```
Misaka-ZeroTwo, line 50: same as above

louis1706: i will fix it on next release

PR #2011, opened 2023-08-25
new CameraType

louis1706:
Hcz049Tunnel -> Hcz049Tunnel
Hcz049Armory -> DELETED
Added
        Hcz173Outside
        Hcz173Hallway
        HczCurve
        
Added / Fix incorect for RoomType.Scp049
        Lcz173Stairs -> Hcz173Stairs 
        Lcz173ContChamber -> Hcz173ContChamber
        
And More Debug


Exiled.API/Enums/CameraType.cs
```diff
@@ -44,12 +47,15 @@ public enum CameraType
...
         #region Hcz
+        [Obsolete("Has been removed from the game")]
         Hcz049Armory,
         Hcz049ContChamber,
         Hcz049ElevTop,
         Hcz049Hallway,
         Hcz049TopFloor,
+        [Obsolete("Has been renamed Hcz049Outside")]
         Hcz049Tunnel,
+        Hcz049Outside = Hcz049Tunnel,
```
Thundermaker300, line 58: why exactly are you using an obsolete enum item right after the line that makes it obsolete lol

louis1706: because they rename the name of the Camera so Hcz049Tunnel will be removed in futur and being replace by Hcz049Outside

Thundermaker300: ok I got that, but like. make the obsolete item use the non-obsolete enum, not vice versa

IRacle1 (approved):

PR #2018, opened 2023-08-29
ChangedItemEventArgs : IItemEvent

louis1706:


Thundermaker300 (approved):

VALERA771:
Why not move NewItem to obsolete?

joker-119:
It should be marked obsolete for the sake of removing it later.

louis1706:
> It should be marked obsolete for the sake of removing it later.
i thinks it's better to keep the one with NewItem even if it's the same than Item to keep than it's more intuitive

IRacle1 (approved):

joker-119:
I respectfully disagree, there's no reason for that kind of redundancy, not to mention if you use ev.Item, in this way, it'll be calling two property get methods instead of one, meaning it's objectively slower.

Exiled.Events/EventArgs/Player/ChangedItemEventArgs.cs
```diff
@@ -43,11 +45,17 @@ public ChangedItemEventArgs(Player player, ItemBase oldItem)
         /// <summary>
         ///     Gets the new item.
         /// </summary>
-        public Item NewItem { get; }
+        [Obsolete("Use ev.Item instead of this")]
+        public Item NewItem => NewItem;
```
VALERA771, line 49:
```suggestion
        public Item NewItem => Item;
```

VALERA771 (changes requested):

Misaka-ZeroTwo (approved):

PR #2028, opened 2023-08-31
No more TODO and obsoletes in API

VALERA771:


Exiled.API/Features/Items/Firearm.cs
```diff
@@ -624,7 +624,7 @@ internal override void ChangeOwner(Player oldOwner, Player newOwner)
                     new DisruptorHitreg(particleDisruptor, particleDisruptor.Owner, particleDisruptor._explosionSettings),
                 Revolver revolver =>
                     new SingleBulletHitreg(revolver, revolver.Owner),
-                _ => throw new NotImplementedException("Should never happend"),
```
louis1706, line 627: Why ?

louis1706 (changes requested):
that the only change than need to be reverted in my opinion everything else is fine

louis1706 (approved):

Misaka-ZeroTwo (approved):

Thundermaker300 (approved):

PR #2031, opened 2023-09-01
Timing event fixes

Vladislav-CS:
There is a problem, for an example, event Detonat**ed** is fired BEFORE warhead detonates.
It fixes:
1. GeneratorActivated
2. Clawed
3. Detonated

### DIFF
--- a/Exiled.Events/Patches/Events/Map/GeneratorActivated.cs
+++ b/Exiled.Events/Patches/Events/Map/GeneratorActivated.cs
@@ -35,8 +35,6 @@ namespace Exiled.Events.Patches.Events.Map
         {
             List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Pool.Get(instructions);
 
-            const int index = 0;
-
             Label retModLabel = generator.DefineLabel();
             Label returnLabel = generator.DefineLabel();
 
@@ -49,7 +47,7 @@ namespace Exiled.Events.Patches.Events.Map
             // if (!ev.IsAllowed)
             //   return;
             newInstructions.InsertRange(
-                index,
+                newInstructions.Count - 1,
                 new CodeInstruction[]
                 {
                     // this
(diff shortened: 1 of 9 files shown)


Exiled.Events/Patches/Events/Warhead/Detonated.cs
```diff
@@ -36,7 +36,7 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
             newInstructions[0].labels.Add(continueLabel);
 
             newInstructions.InsertRange(
-                0,
+                newInstructions.Count - 1,
```
VALERA771, line 39: This is detonating event and it's calling before warhead detonate

Exiled.Events/Patches/Events/Map/GeneratorActivated.cs
```diff
@@ -49,7 +47,7 @@ private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstructi
             // if (!ev.IsAllowed)
             //   return;
             newInstructions.InsertRange(
-                index,
```
IRacle1, line 52: idc about that, I'd like to simply rename the event and no to change patch
(if you can disable event by IsAllowed, then event is called *before* event)

Exiled.Events/Patches/Events/Warhead/Detonated.cs
Vladislav-CS: Yeah, i see, someone fixed it but i found another problem

IRacle1 (changes requested):

Exiled.Events/Patches/Events/Map/GeneratorActivated.cs
Vladislav-CS: Bruh, bad naming. It's fired before generator activates and has IsAllowed property but called Activat**ed**

louis1706: Oh i didn't seen than it's isallowed bruh

PR #2033, opened 2023-09-01
Jailbird Break fix

### LINKED ISSUE #1858: [BUG] Jailbird break method doesn't work
**Describe the bug**
No any errors, it just does nothing.
In master was this code and it works fine
```
        public void Break()
        {
            Base._broken = true;
            Base.SendRpc(JailbirdMessageType.Broken);
        }
```

**EXILED Version ("latest" is not a version):**
7.0.0-rc.5

Vladislav-CS:
#1858
Has to be tested


Exiled.API/Features/Items/Jailbird.cs
```diff
@@ -110,7 +110,10 @@ public JailbirdWearState WearState
         /// <summary>
         /// Breaks the Jailbird.
         /// </summary>
-        public void Break() => WearState = JailbirdWearState.Broken;
+        public void Break()
+        {
+            Base.SendRpc(JailbirdMessageType.Broken);
```
NaoUnderscore, line 115: Keep the lambda: `public void Break() => Base.SendRpc(JailbirdMessageType.Broken);`

NaoUnderscore (changes requested):

Exiled.API/Features/Items/Jailbird.cs
louis1706: actually i don't even see how the Item will be destroy it's just send Rpc to client

Thundermaker300:
Shouldn't this also have `Base._broken = true;`? It only sends RPC, doesn't actually break it

louis1706:
> Shouldn't this also have `Base._broken = true;`? It only sends RPC, doesn't actually break it

that what i said in nao Requested Change

louis1706 (approved):

louis1706 (commented):
I would just like than you test this i am not sure if the Item will imidiately disapear when destroy

PR #2037, opened 2023-09-01
Fix RoomLightControllers

louis1706:


louis1706:
i am actually not sure about the fix

IRacle1:
In general you fix setters, but in my pr i also fixes all `RoomLight` related things

PR #2048, opened 2023-09-02
Smart ammo limits

joker-119:
Allow Custom Roles to easily calculate the max ammo a player can carry of a specific type, and set their ammo to that, by setting the ammo property value to ushort.MaxValue.

ushort.MaxValue - 1 can still be used to give them basically the normal 65k possible max.

Also allows for getting CustomAbilities by their type, rather than name.

### DIFF
--- a/Exiled.CustomRoles/API/Features/CustomAbility.cs
+++ b/Exiled.CustomRoles/API/Features/CustomAbility.cs
@@ -60,18 +60,38 @@ namespace Exiled.CustomRoles.API.Features
         public string AbilityType { get; }
 
         /// <summary>
-        /// Gets a <see cref="CustomRole"/> by name.
+        /// Gets a <see cref="CustomAbility"/> by name.
         /// </summary>
-        /// <param name="name">The name of the role to get.</param>
-        /// <returns>The role, or <see langword="null"/> if it doesn't exist.</returns>
+        /// <param name="name">The name of the ability to get.</param>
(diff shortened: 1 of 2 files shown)


Thundermaker300 (approved):

Exiled.CustomRoles/API/Features/CustomAbility.cs
```diff
@@ -60,18 +60,38 @@ public CustomAbility()
...
+        /// Tries to get a <see cref="CustomAbility"/> by type.
+        /// </summary>
+        /// <param name="type">The type of the ability to get.</param>
+        /// <param name="customAbility">The custom ability.</param>
+        /// <returns>True if the ability exists, otherwise false.</returns>
+        public static bool TryGet(Type type, out CustomAbility? customAbility)
+        {
+            customAbility = Get(type);
+
+            return customAbility is not null;
```
louis1706, line 86: can't this be inlined ?

louis1706 (changes requested):

Exiled.CustomRoles/API/Features/CustomAbility.cs
joker-119: It *can* but there's no benefit of it, and doesn't follow the style consistency of the various existing TryGet methods in the API.

louis1706 (approved):

Thundermaker300 (approved):

PR #2050, opened 2023-09-03
Make Amount in DroppingAmmoEventArgs ushort instead of int

ghost:
DroppingAmmoEventArgs.Amount : int -> ushort

### DIFF
--- a/Exiled.Events/EventArgs/Player/DroppingAmmoEventArgs.cs
+++ b/Exiled.Events/EventArgs/Player/DroppingAmmoEventArgs.cs
@@ -52,7 +52,7 @@ namespace Exiled.Events.EventArgs.Player
         /// <summary>
         ///     Gets or sets the amount of ammo being dropped.
         /// </summary>
-        public int Amount { get; set; }
+        public ushort Amount { get; set; }
 
         /// <summary>
         ///     Gets or sets a value indicating whether or not the ammo can be dropped.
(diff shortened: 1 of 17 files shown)


NaoUnderscore (approved):

Exiled.Events/EventArgs/Player/DroppingAmmoEventArgs.cs
```diff
@@ -52,7 +52,7 @@ public DroppingAmmoEventArgs(Player player, AmmoType ammoType, ushort amount, bo
         /// <summary>
         ///     Gets or sets the amount of ammo being dropped.
         /// </summary>
-        public int Amount { get; set; }
```
louis1706, line 55: Use Obsolete instead of removing it and cast it to Ushort ?

louis1706 (changes requested):

Exiled.Events/EventArgs/Player/DroppingAmmoEventArgs.cs
NaoUnderscore: `ushort` is what the game uses, how is he going to cast it?

louis1706 (approved):
Nao said me than it's isn't a break change for Plugin (if it is i will blamed NAO for it)

Bonjemus:
This pull is breaking bad.

PR #2057, opened 2023-09-03
Remove Singleton from Exiled.Example

Misfiy:
crazy? i was crazy once


Misfiy:
i take it back

louis1706 (approved):

Ruemena:
imo i think that at the very least the constructor should be private

Exiled.Example/Example.cs
```diff
@@ -26,21 +24,18 @@ public class Example : Plugin<Config>
...
-        private Example()
-        {
-        }
-
         /// <summary>
-        /// Gets the only existing instance of this plugin.
+        /// The Instance of the plugin.
         /// </summary>
-        public static Example Instance => Singleton;
+        public static Example Instance;
```
Thundermaker300, line 30:
```suggestion
        public static Example Instance { get; set; }
```
Property

Thundermaker300 (changes requested):

Jesus-QC:
This shouldnt be approved, it showcases how if there ia already an existing static instance and the constructor is private it will take that one instance instead of making a new one.

PR #2068, opened 2023-09-04
CustomRoles improvements

Misfiy:
Might work, might not, who knowsd


NaoUnderscore:
I'm not against this change, however we're going to rework custom role API soon, we may consider the project "frozen" unless it's regarding some critical bugs.

Misfiy:
roger that 👍

PR #2075, opened 2023-09-05
Add `SoundtrackMute` to exceptions

NaoUnderscore:
Of course it is what it is, however unless we re-define the dictionary from which we get all effects, that's the only way to fix it.


louis1706:
i thinks we are going to need to make this bigger ^^'

NaoUnderscore:
> i thinks we are going to need to make this bigger ^^'
> 
> Amnesia Vision Amnesia Item

Take a look to the last commit.

louis1706 (approved):

PR #2077, opened 2023-09-05
`Scp173::BlinkingRequest` event

IRacle1:
there were a some reason to create that event:
 - if you disallow `Blinking` event, blink sound still be playing
 - although, before sending blinking request, blink position is unavailable, so we cant just edit `Blinking` event invoke


NaoUnderscore (approved):

louis1706:
> there were a some reason to create that event:
> 
> * if you disallow `Blinking` event, blink sound still be playing
> * although, before sending blinking request, blink position is unavailable, so we cant just edit `Blinking` event invoke

so we should remove the IsAllowed from Blinking and asking people to use BlinkingRequest ?

louis1706 (approved):

PR #2081, opened 2023-09-05
`Exiled.Updater` rework

NaoUnderscore:


Exiled.Loader/Config.cs
```diff
@@ -29,27 +30,27 @@ public sealed class Config : IConfig
...
+        [Description("Indicates whether outdated plugins should be loaded or not.")]
         public bool ShouldLoadOutdatedPlugins { get; set; } = true;
 
         /// <summary>
         /// Gets or sets the Exiled directory path from which plugins will be loaded.
         /// </summary>
-        [Description("The Exiled directory path from which plugins will be loaded")]
-        public string ExiledDirectoryPath { get; set; } = Path.Combine(Paths.AppData, "EXILED");
+        [Description("The Exiled directory path from which plugins will be loaded.")]
+        public string ExiledDirectoryPath { get; set; } = Path.Combine(Paths.AppData, "EXILED.");
```
louis1706, line 48: Eeeh this dot shouldn't be here i thinks ?

louis1706 (changes requested):
I also have an other idea 
For making it updating without needing to restart server

Exiled.Loader/Config.cs
NaoUnderscore: Oh my bad

PR #2084, opened 2023-09-06
FixGateIsFullyOpen

louis1706:


Exiled.API/Features/Doors/Gate.cs
```diff
@@ -40,6 +40,21 @@ public Gate(PryableDoor door, List<Room> room)
         /// </summary>
         public IEnumerable<Transform> PryPositions => Base.PryPositions;
 
+        /// <summary>
+        /// Gets a value indicating whether or not the door is fully closed.
+        /// </summary>
+        public new bool IsFullyClosed => base.IsFullyClosed && RemainingPryCooldown > 0;
```
IRacle1, line 46: it doesnt make any sense, those methods should be overridden from base with `override` keyword

Exiled.API/Features/Doors/Door.cs
```diff
@@ -101,17 +101,17 @@ internal Door(DoorVariant door, List<Room> rooms)
         /// <summary>
         /// Gets a value indicating whether or not the door is fully closed.
         /// </summary>
-        public bool IsFullyClosed => Is(out Gate gate) ? !IsOpen && gate.RemainingPryCooldown <= 0 : ExactState is 0;
+        public bool IsFullyClosed => ExactState is 0;
```
IRacle1, line 104: all of this should be `virtual`

IRacle1 (changes requested):

Exiled.API/Features/Doors/Gate.cs
louis1706: Isn't this would make them breaking change ?

Issue #1905, opened 2023-06-25
[BUG] No error when trying to parse duplicate keys in plugin config dictionary


Misaka-ZeroTwo:
When there is a Dictionary<TKey, TValue> for a plugin's config, and you put a duplicate key value in the plugin config file, EXILED will reset the plugin's config without saying anything (no WARN or ERROR messages are printed in console. A backup file is not created, it just wipes it silently.)

Exiled Version: 7.0.0-rc.5

Example:
```
    OneToOne:
    - original: Scp173
      new: Scp096
      chance: 100
      keep_inventory: true
    OneToOne:
    - original: Scp096
      new: Scp939
      chance: 100
      keep_inventory: true
    OneToOne:
    - original: Scp939
      new: Scp049
      chance: 100
      keep_inventory: true
    OneToOne:
    - original: Scp049
      new: Scp106
      chance: 100
      keep_inventory: true
  ```

ALEXWARELLC:
Ooo, maybe something I can actually do.

Do you want an error message or what?

Misaka-ZeroTwo:
> Ooo, maybe something I can actually do.
> 
> Do you want an error message or what?

There should be an error when it happens, and a backup file should be created.

ALEXWARELLC:
> There should be an error when it happens, and a backup file should be created.

OK. I'll get started on that. Thanks.

ALEXWARELLC:
#2087

Is this what you wanted?

Misaka-ZeroTwo:
> #2087
> 
> Is this what you wanted?

That will not solve the issue.

ALEXWARELLC:
Where is this happening as I'm not sure where to look as everything seems to do the same thing.

ALEXWARELLC:
Does this occur now?

PR #2087, opened 2023-09-07
Reworked Plugin Config Detection.

ALEXWARELLC:
Stopped plugin configs from loading duplicate keys.

It will ignore keys that have been duplicated after the original key has been loaded.

### DIFF
--- a/Exiled.Loader/ConfigManager.cs
+++ b/Exiled.Loader/ConfigManager.cs
@@ -44,11 +44,9 @@ namespace Exiled.Loader
                 foreach (IPlugin<IConfig> plugin in Loader.Plugins)
                 {
                     if (!deserializedConfigs.ContainsKey(plugin.Prefix))
-                    {
                         deserializedConfigs.Add(plugin.Prefix, plugin.LoadConfig(rawDeserializedConfigs));
-                    } else {
+                    else
                         Log.Warn($"{plugin.Prefix} already exists in {plugin.Name}'s configuration. {plugin.Prefix} has been skipped.");
-                    }
                 }
 
                 // Make sure that no keys in the config file were discarded. (Individual can ignore this since rawDeserializedConfigs is null)
@@ -58,7 +56,7 @@ namespace Exiled.Loader
                     File.WriteAllText(Paths.BackupConfig, rawConfigs);
                 }
 
-                Log.Info("All plugin configs have been loaded successfully!");
+                Log.Info("Plugin configs loaded successfully!");
 
                 DictionaryPool<string, object>.Pool.Return(rawDeserializedConfigs);
                 return deserializedConfigs;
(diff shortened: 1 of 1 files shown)


VALERA771:
Useless change. This is never happening and exception will be thrown if we will try to add duplicate key in dictionary

VALERA771 (changes requested):
Useless change. This is never happening and exception will be thrown if we will try to add duplicate key in dictionary

VALERA771:
In issue we have another case. You edit it to prevent adding configs of plugins with same prefix but we have a duplicate keys in dictionaries in plugin configs. So this change won't affect anything

moddedmcplayer:
> Useless change. This is never happening and exception will be thrown if we will try to add duplicate key in dictionary

#1905

Exiled.Loader/ConfigManager.cs
```diff
@@ -52,7 +58,7 @@ public static class ConfigManager
                     File.WriteAllText(Paths.BackupConfig, rawConfigs);
                 }
 
-                Log.Info("Plugin configs loaded successfully!");
+                Log.Info("All plugin configs have been loaded successfully!");
```
moddedmcplayer, line 61: Why?

NotIntense:
lmfao wtf

VALERA771:
> > Useless change. This is never happening and exception will be thrown if we will try to add duplicate key in dictionary
> 
> #1905

He didn't fix this issue. Just look at edited files

moddedmcplayer:
> > > Useless change. This is never happening and exception will be thrown if we will try to add duplicate key in dictionary
> > 
> > 
> > #1905
> 
> He didn't fix this issue. Just look at edited files

💀 mb

Exiled.Loader/ConfigManager.cs
```diff
@@ -40,9 +40,15 @@ public static class ConfigManager
...
                 SortedDictionary<string, IConfig> deserializedConfigs = new(StringComparer.Ordinal);
 
+                // Only allow unique keys to be loaded. Any duplicates in the plugin's config will be skipped.
                 foreach (IPlugin<IConfig> plugin in Loader.Plugins)
                 {
-                    deserializedConfigs.Add(plugin.Prefix, plugin.LoadConfig(rawDeserializedConfigs));
+                    if (!deserializedConfigs.ContainsKey(plugin.Prefix))
+                    {
+                        deserializedConfigs.Add(plugin.Prefix, plugin.LoadConfig(rawDeserializedConfigs));
+                    } else {
```
NaoUnderscore, line 49: immagine

This bracing style won't be accepted.

Exiled.Loader/ConfigManager.cs
NaoUnderscore, line 61: Not needed.

NaoUnderscore (changes requested):

Exiled.Loader/ConfigManager.cs
ALEXWARELLC: Reset back to orginal.

Exiled.Loader/ConfigManager.cs
ALEXWARELLC: Updated Bracing Style.

PR #2105, opened 2023-09-11
Make `Server.RunCommand` return response

ITeMbI4:


Exiled.API/Features/Server.cs
```diff
@@ -248,7 +248,7 @@ public static bool ShutdownRedirect(ushort redirectPort)
         /// </summary>
         /// <param name="command">The command to be run.</param>
         /// <param name="sender">The <see cref="CommandSender"/> running the command.</param>
-        public static void RunCommand(string command, CommandSender sender = null) => GameCore.Console.singleton.TypeCommand(command, sender);
+        public static string RunCommand(string command, CommandSender sender = null) => GameCore.Console.singleton.TypeCommand(command, sender);
```
louis1706, line 251: Breaking change make obsolete the previous one also can we get if the command has successfully be made or not ?

louis1706 (changes requested):

Exiled.API/Features/Server.cs
Misaka-ZeroTwo: > Breaking change make obsolete the previous one also can we get if the command has successfully be made or not ?

Compile Error CS0111 will be thrown if a class contains two member declarations with the same name and parameter types.

VALERA771: > > Breaking change make obsolete the previous one also can we get if the command has successfully be made or not ?
> 
> 
> 
> Compile Error CS0111 will be thrown if a class contains two member declarations with the same name and parameter types.

Can be renamed to ExecuteCommand or something else like this. It really will be a breaking change so

ITeMbI4: @Misaka-ZeroTwo @louis1706 check please

louis1706 (approved):

Misaka-ZeroTwo (approved):

Thundermaker300 (approved):

PR #2114, opened 2023-09-20
add TargetOffset

Ruemena:
this adds targetoffset, a way to change the number of targets displayed to scps. this allows you to implement npcs, custom roles, etc that aren't displayed as targets (among other things)
TargetOffset is added to the total number of targets - it doesn't allow you to directly set the number of targets directly, as that would break compatibility between plugins
UpdateTargetCount() forces an update in the number of targets. this is necessary because you may not want to immediately update the number of targets displayed.

tested and works fine (for real)

**THIS MAY STILL NEED SOME REFINING** as rn there's no way to not have an incredibly quick increase followed by a decrease or vice versa when modifying the targetoffset and spawning or killing a player. this could prob be fixed by sending a sync var on the spawning event, or something like that


louis1706 (changes requested):
the Patches should be ine Generic file not Fixed

Exiled.Events/Patches/Generic/TargetOffset.cs
```diff
@@ -0,0 +1,54 @@
+// -----------------------------------------------------------------------
+// <copyright file="TargetOffset.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.Events.Patches.Events.Player
```
louis1706, line 8: This is not the correct 'amespace

louis1706 (changes requested):

Exiled.Events/Patches/Generic/TargetOffset.cs
Ruemena: fixed

louis1706 (changes requested):
Fix Build Error

Exiled.API/Features/Round.cs
```diff
@@ -174,6 +174,23 @@ public static IEnumerable<Side> AliveSides
             }
         }
 
+        /// <summary>
+        /// Gets or sets a visual offset applied to the target counter for SCPs.
+        /// </summary>
+        /// <remarks>This does not update the target counter automatically. To do that, call Round.UpdateTargetCounter().</remarks>
+        public static int TargetOffset { get; set; } = 0;
```
louis1706, line 181:
```suggestion
        public static int TargetOffset { get; set; }
```

louis1706 (approved):

Exiled.API/Features/Round.cs
```diff
@@ -239,5 +254,16 @@ public static bool EndRound(bool forceEnd = false)
         /// Start the round.
         /// </summary>
         public static void Start() => CharacterClassManager.ForceRoundStart();
+
+        /// <summary>
+        /// Updates the target counter for SCPs.
+        /// </summary>
+        private static void UpdateTargetCounter()
```
NaoUnderscore, line 261: This method is called once, why making it a method then?

NaoUnderscore (changes requested):

Exiled.API/Features/Round.cs
```diff
@@ -248,5 +263,16 @@ public static bool EndRound(bool forceEnd = false)
...
+        {
+            foreach (Player player in Player.List)
+            {
+                player.SendFakeSyncVar(RoundSummary.singleton.netIdentity, typeof(RoundSummary), nameof(RoundSummary.Network_chaosTargetCount), RoundSummary.singleton._chaosTargetCount + TargetOffset);
+            }
+        }
```
louis1706, line 276:
```suggestion
```

Exiled.API/Features/Round.cs
```diff
@@ -183,6 +185,19 @@ public static IEnumerable<Side> AliveSides
...
+        /// <summary>
+        /// Gets or sets a visual offset applied to the target counter for SCPs.
+        /// </summary>
+        public static int TargetOffset
+        {
+            get => targetOffset;
+            set
+            {
+                targetOffset = value;
+                UpdateTargetCounter();
```
louis1706, line 197:
```suggestion
            foreach (Player player in Player.List)
            {
                player.SendFakeSyncVar(RoundSummary.singleton.netIdentity, typeof(RoundSummary), nameof(RoundSummary.Network_chaosTargetCount), RoundSummary.singleton._chaosTargetCount + TargetOffset);
            }
```

Exiled.API/Features/Round.cs
```diff
@@ -183,6 +185,22 @@ public static IEnumerable<Side> AliveSides
...
+        public static int TargetOffset
+        {
+            get => targetOffset;
+            set
+            {
+                targetOffset = value;
+            foreach (Player player in Player.List)
+            {
+                player.SendFakeSyncVar(RoundSummary.singleton.netIdentity, typeof(RoundSummary), nameof(RoundSummary.Network_chaosTargetCount), RoundSummary.singleton._chaosTargetCount + TargetOffset);
+            }
```
louis1706, line 200:
```suggestion
                targetOffset = value;
                foreach (Player player in Player.List)
                {
                    player.SendFakeSyncVar(RoundSummary.singleton.netIdentity, typeof(RoundSummary), nameof(RoundSummary.Network_chaosTargetCount), RoundSummary.singleton._chaosTargetCount + TargetOffset);
                }
```

PR #2124, opened 2023-10-01
Add LocalPositionOffset to Room

Mirecek0010:
Just a little addition to determine the LocalPosition offset based on World Position

Basically, if I have a Room and want to determine it's LocalPosition in order to, for example, spawn something at a certain place in a room no matter the rotation, I can do so using this.
I've been doing this before using MER and primitiveobjects which showed localposition, but I thought of adding it as a normal thing so it doesn't have to be this unnecessarily complicated

### DIFF
--- a/Exiled.API/Features/Room.cs
+++ b/Exiled.API/Features/Room.cs
@@ -308,6 +308,13 @@ namespace Exiled.API.Features
         public Vector3 WorldPosition(Vector3 offset) => Transform.InverseTransformPoint(offset);
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


louis1706:
image
it's just already exist ?!???

Mirecek0010:
It is the same thing but worded very differently. Since with WorldPosition, I'd not expect to be able to get the offset of Local Position, and without looking into the Exiled code itself, I'd have no idea it can be used for that.
Don't know if it's worth adding an extra thing, but I just figured it would be useful to have it worded like this, too

PR #2133, opened 2023-10-06
Yes

joker-119:


Exiled.CustomRoles/API/Features/CustomRole.cs
```diff
@@ -528,18 +534,18 @@ public virtual void AddRole(Player player)
...
-            if (position != Vector3.zero)
-            {
-                player.Position = position;
-            }
+                    Vector3 position = GetSpawnPosition();
+                    if (position != Vector3.zero)
+                    {
+                        player.Position = position;
+                    }
+                });
```
louis1706, line 548: you don't need to delay all this stuff

louis1706 (changes requested):

NaoUnderscore:
I'm sure we're going to rework it as well, but as long as we keep it as is, this should be fine.
Just take a look to what @louis1706 is barking about.

Exiled.CustomRoles/API/Features/KeypressActivator.cs
```diff
@@ -121,6 +99,34 @@ private IEnumerator<float> ProcessAltKey(Player player)
...
+
+            if (!coroutineTracker[ev.Player].IsRunning)
+                coroutineTracker[ev.Player] = Timing.RunCoroutine(ProcessAltKey(ev.Player));
+        }
+
+        private void OnEndingRound(EndingRoundEventArgs ev)
+        {
+            altTracker.Clear();
+            foreach (CoroutineHandle handle in coroutineTracker.Values)
+                Timing.KillCoroutines(handle);
```
NaoUnderscore, line 126: `Timing.KillCoroutines(coroutineTracker.Values.ToArray());`
May be better.

Exiled.CustomRoles/API/Features/CustomRole.cs
joker-119: Ok, I'll close my or, let you remove them, and say I told you so when everything breaks. Have fun.

Exiled.CustomRoles/API/Features/KeypressActivator.cs
louis1706: you just make me learn this BRUH

PR #2134, opened 2023-10-08
EnableEffect should have intensity argument

louis1706:
image


warden161:
Why would it have the intensity argument? ChangeEffectIntensity does this.

louis1706:
> Why would it have the intensity argument? ChangeEffectIntensity does this.

why needing two method to do the same things at once ?

NaoUnderscore (approved):

Thundermaker300 (approved):

PR #2136, opened 2023-10-09
NPC class adjustments + Server count fix

NotIntense:


NotIntense:
Server count fix is UNTESTED, cant get a verified server to test this on but it patched fine.

louis1706 (approved):

louis1706:
I would just like to wait a Test before merge

PR #2137, opened 2023-10-10
EventList

louis1706:


Exiled.Events/Features/Event.cs
```diff
@@ -23,17 +24,25 @@ namespace Exiled.Events.Features
     /// </summary>
     public class Event : IExiledEvent
     {
+        private static readonly List<Event> EventList = new();
```
NaoUnderscore, line 27: Rename it to `EventsValue`

NaoUnderscore (changes requested):

Exiled.Events/Features/Event.cs
louis1706: sure

PR #2142, opened 2023-10-11
Fix `Scp0492.ConsumingCorpse` event was called with `Scp049`

IRacle1:
Scp049 resurrect ability calls zombie event.
Idk how is can be, but seems like patching generic classes is weird.
tested and confirmed by me

### DIFF
--- a/Exiled.Events/EventArgs/Scp0492/ConsumingCorpseEventArgs.cs
+++ b/Exiled.Events/EventArgs/Scp0492/ConsumingCorpseEventArgs.cs
@@ -58,15 +58,8 @@ namespace Exiled.Events.EventArgs.Scp0492
         /// </summary>
         public bool IsAllowed
         {
-            get
-            {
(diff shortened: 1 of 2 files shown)


louis1706 (approved):

Exiled.Events/EventArgs/Scp0492/ConsumingCorpseEventArgs.cs
```diff
@@ -56,6 +56,17 @@ public ConsumingCorpseEventArgs(ReferenceHub player, BasicRagdoll ragDoll, Zombi
         /// <summary>
         ///     Gets or sets a value indicating whether 049-2 can consume a corpse.
         /// </summary>
-        public bool IsAllowed { get; set; }
+        public bool IsAllowed
```
NaoUnderscore, line 59: I'd personally remove braces.

PR #2145, opened 2023-10-14
DecontaminationFeatures

louis1706:


Exiled.API/Enums/DecontaminationState.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    using System;
+
+    using Features;
+
+    /// <summary>
+    /// Represents the state of a <see cref="LightContainmentZoneDecontamination.DecontaminationController"/>.
+    /// </summary>
+    /// <seealso cref="Map.DecontaminationState"/>
+    [Flags]
```
IRacle1, line 18: why is it flags?

louis1706: Oups

IRacle1 (approved):

PR #2151, opened 2023-10-18
We are not going to miss this (Remove Old Exiled.Updater)

louis1706:


Exiled.Loader/Loader.cs
```diff
@@ -120,6 +120,9 @@ public Loader()
         /// </summary>
         public static void LoadPlugins()
         {
+            // Required to remove the old ExiledUpdater than now integrated inside Exiled.Loader
+            File.Delete($"{Paths.Plugins}/Exiled.Updater.dll");
```
IRacle1, line 124: use `Path.Combine(Paths.Plugins, "Exiled.Updater.dll")`

separating symbol is different on Win and Unix

IRacle1 (changes requested):

Exiled.Loader/Loader.cs
louis1706: Oh okay yes

PR #2160, opened 2023-10-21
Add Whitelist && update ReservedSlot extensions

ITeMbI4:
This PR adds Whitelist && update ReservedSlot extensions in Player.cs


VALERA771 (commented):
Why we can't just use File.AppendAllLines() or smth else

louis1706 (changes requested):
Don't make Breaking Change please keep the previous method and just add the new one with new args

VALERA771 (approved):

Exiled.API/Features/Player.cs
```diff
@@ -1408,11 +1418,37 @@ public static Player Get(string args)
         /// <seealso cref="GiveReservedSlot()"/>
         public static bool AddReservedSlot(string userId) => ReservedSlot.Users.Add(userId);
 
+        /// <summary>
+        /// Adds a player's UserId to the list of reserved slots permanently. It will add a UserId to UserIDReservedSlots.txt file.
+        /// </summary>
+        /// <param name="userId">The UserId of the player to add.</param>
+        public static void PermanentlyAddReservedSlot(string userId) => ReservedSlots.Add(userId);
```
Thundermaker300, line 1425: Maybe instead of a different method, we can add a `bool permanent` to `AddReservedSlot`?

Thundermaker300: Oh, that was the original intention. Thanks, Yamato

louis1706 (approved):

PR #2164, opened 2023-10-22
MaxAmmo Setter && NightVision should also be true when NightVision

louis1706:
- Fix Custom Weapons can reload without going through the animation
- Added MaxAmmo Setter
- Added CanSeeThroughDark
- Added NightVisionEnabled


Exiled.API/Features/Items/Firearm.cs
```diff
@@ -118,9 +118,27 @@ public byte Ammo
...
+            {
+                switch (Base.AmmoManagerModule)
+                {
+                    case TubularMagazineAmmoManager tubularMagazineAmmoManager:
+                        tubularMagazineAmmoManager.MaxAmmo = value;
+                        break;
+                    case ClipLoadedInternalMagAmmoManager clipLoadedInternalMagAmmoManager:
+                        clipLoadedInternalMagAmmoManager.MaxAmmo = value;
+                        break;
+                    case AutomaticAmmoManager automaticAmmoManager:
```
Thundermaker300, line 136: This switch should have a default clause and throw an exception if it's a module that cannot be set.

Exiled.API/Features/Items/Firearm.cs
```diff
@@ -138,9 +156,10 @@ public byte Ammo
         public bool Aiming => Base.AdsModule.ServerAds;
 
         /// <summary>
-        /// Gets a value indicating whether the firearm's flashlight module is enabled.
+        /// Gets a value indicating whether the firearm's flashlight module is enabled or if playing see trough a NightVision.
         /// </summary>
-        public bool FlashlightEnabled => Base.Status.Flags.HasFlagFast(FirearmStatusFlags.FlashlightEnabled);
+        public bool FlashlightEnabled => Base.Status.Flags.HasFlagFast(FirearmStatusFlags.FlashlightEnabled) ||
```
Thundermaker300, line 161: Separate property for night vision

Exiled.API/Features/Items/Firearm.cs
louis1706: The only default would be with Disruptor than MaxAmmo is actually useless

Thundermaker300: I'm aware, which is why it should throw an error as opposed to not doing anything, as you cannot set the max ammo in this case.

Exiled.API/Features/Items/Firearm.cs
louis1706: How would you call the property that combined both ?

Thundermaker300: Separate property for Flashlight vs night vision

louis1706: > Separate property for Flashlight vs night vision

I still want a property than combine both of them how should i call it ?

Thundermaker300: IsEmittingLight would probably work

Exiled.API/Features/Items/Firearm.cs
```diff
@@ -143,6 +165,16 @@ public byte Ammo
         /// </summary>
         public bool FlashlightEnabled => Base.Status.Flags.HasFlagFast(FirearmStatusFlags.FlashlightEnabled);
 
+        /// <summary>
+        /// Gets a value indicating whether the firearm's flashlight module is enabled or if playing see trough a NightVision.
```
Misaka-ZeroTwo, line 169:
```suggestion
        /// Gets a value indicating whether the firearm's NightVision is being used.
```

Exiled.API/Features/Items/Firearm.cs
```diff
@@ -143,6 +165,16 @@ public byte Ammo
...
+        /// </summary>
+        public bool NightVisionEnabled => Aiming && Base.HasAdvantageFlag(AttachmentDescriptiveAdvantages.NightVision);
+
+        /// <summary>
+        /// Gets a value indicating whether the firearm's flashlight module is enabled or if playing see trough a NightVision.
```
Misaka-ZeroTwo, line 174:
```suggestion
        /// Gets a value indicating whether the firearm's flashlight module is enabled or NightVision is being used.
```

Exiled.API/Features/Items/Firearm.cs
```diff
@@ -143,6 +165,16 @@ public byte Ammo
...
+        /// </summary>
+        public bool IsEmittingLight => FlashlightEnabled || NightVisionEnabled;
```
louis1706, line 176:
```suggestion
        public bool CanSeeThroughDark => FlashlightEnabled || NightVisionEnabled;
```

Misaka-ZeroTwo:
Tested by trouble.

PR #2183, opened 2023-10-29
feat(events): implement EventExceptionLogger to capture event exceptions

Nekonyx:
No breaking changes.

I extracted error logging from `Exiled.Events.Features.Event.InvokeSafely()` into a separate public static class.

This can be useful for patching, logging errors not only to the console but also to services like Sentry, Hawk, DataDog, or any other destination of choice.

### DIFF
--- a/Exiled.Events/Features/Event.cs
+++ b/Exiled.Events/Features/Event.cs
@@ -109,7 +109,7 @@ namespace Exiled.Events.Features
                 }
                 catch (Exception ex)
                 {
-                    Log.Error($"Method \"{handler.Method.Name}\" of the class \"{handler.Method.ReflectedType.FullName}\" caused an exception when handling the event \"{GetType().FullName}\"\n{ex}");
+                    EventExceptionLogger.Capture(ex, GetType().FullName, handler.Method);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


IRacle1 (approved):

louis1706:
> No breaking changes.
> 
> I extracted error logging from `Exiled.Events.Features.Event.InvokeSafely()` into a separate public static class.
> 
> This can be useful for patching, logging errors not only to the console but also to services like Sentry, Hawk, DataDog, or any other destination of choice.

can you do the same PR but for DevBranch ? halloween branch will not be merged on DevBranch or MainBranch

louis1706 (changes requested):
Config inside Exiled.Event for enabling or disabling logging

Nekonyx:
@louis1706 
> can you do the same PR but for DevBranch ? halloween branch will not be merged on DevBranch or MainBranch
Okay
> Config inside Exiled.Event for enabling or disabling logging
I don't think that it is a good idea to disable error logs in configs. Logging event errors in the console is very necessary for everyone, they will never be redundant.

Nekonyx:
@louis1706 #2207

PR #2195, opened 2023-10-31
`CancellingItemUse` fix

IRacle1:


louis1706:
i made an other change instead of this (thanks for finding the cause of the problem)

PR #2197, opened 2023-10-31
Fix ChangingGroupEventArgs.NewGroup having no effect.

o5zereth:


o5zereth:
@Misaka-ZeroTwo yes

Misaka-ZeroTwo (dismissed):

VALERA771 (commented):
I think group should be changing after checking IsAllowed

louis1706 (changes requested):
Valera request is right it's should be done after the ev.IsAllowed

o5zereth:
why does it matter? i would have to make a new local var which jusy makes the pr more complicated

o5zereth:
> Valera request is right it's should be done after the ev.IsAllowed

why does it matter? i would have to make a new local var which just makes the pr more complicated.

its storing into the argument slot, which wont break anything. if it isnt allowed it will return anyways.

louis1706 (approved):

Misaka-ZeroTwo:
Sorry, please target to dev branch cuz Halloween is ending.

o5zereth:
Im tired of waiting, so im not wasting my time making a third pr. Do it yourself.

PR #2198, opened 2023-11-01
Fix Rpc and WriterExtensions dict

sanyae2439:
works fine™


louis1706:
We also need to merge this to devbranch Halloween will not be merge into dev

louis1706 (approved):

louis1706 (approved):

PR #2213, opened 2023-11-07
CommonExtensions::Chance

FUTURE-SL:


Exiled.API/Features/Spawn/DynamicSpawnPoint.cs
```diff
@@ -28,7 +28,7 @@ public class DynamicSpawnPoint : SpawnPoint
         public SpawnLocationType Location { get; set; }
 
         /// <inheritdoc/>
-        public override float Chance { get; set; }
```
Misaka-ZeroTwo, line 31: Why?

Misaka-ZeroTwo (dismissed):
There's no need to change float to int. People might want things to happen in rare occasions.

Exiled.API/Features/Spawn/DynamicSpawnPoint.cs
FUTURE-SL: My bad, i have updated my solution

Exiled.Loader/Loader.cs
NotIntense, line 1: Why would this needed to be removed? Lots of plugins depend on this and may not want to use GetRandomValue etc

NotIntense (changes requested):
I dont think this should be merged. While cleaning up the code by using `.GetRandomValue();` is good, removing a highly popular and used static Random method, that'll break hundreds of plugins for a single bool method that only accepts a float value.

Exiled.Loader/Loader.cs
FUTURE-SL: I listened to your opinion and updated the solution, thanks!

Exiled.API/Extensions/CommonExtensions.cs
```diff
@@ -61,5 +61,26 @@ public static AnimationCurve Add(this AnimationCurve curve, float amount)
 
             return curve;
         }
+
+        /// <summary>
+        /// Get a chance of a number falling out (as a percentage).
+        /// </summary>
+        /// <param name="chance">The value of chance.</param>
+        /// <returns>Whether the necessary chance has fallen out.</returns>
+        public static bool Chance(this double chance) => Chance(chance);
```
louis1706, line 70: this isn't double check but a float check i would prefer than you use Exiled.Loader.Random.NextDouble() so you will compare it with double and not use float

louis1706 (changes requested):
we want to keep the precision of double for random :)

Exiled.API/Extensions/CommonExtensions.cs
FUTURE-SL: I have updated my solution. Maybe it will be better

Exiled.API/Extensions/CommonExtensions.cs
```diff
@@ -17,13 +17,18 @@ namespace Exiled.API.Extensions
     /// </summary>
     public static class CommonExtensions
     {
+        /// <summary>
+        /// Gets the initialized global random class.
+        /// </summary>
+        public static System.Random Random { get; } = new();
```
louis1706, line 23: Use the one from Exiled.Loader

louis1706 (changes requested):

Exiled.API/Extensions/CommonExtensions.cs
FUTURE-SL: image
When adding Exiled.Loader to Exiled.API ☠️

louis1706: oh yeah that normal i haven't thinks to that when i have said it

Exiled.Loader/Features/LoaderMessages.cs
```diff
@@ -81,7 +83,7 @@ public static class LoaderMessages
         /// <returns>The correspondent loader message.</returns>
         public static string GetMessage()
         {
-            if (Loader.Version.ToString().Contains("6.9") || Loader.Random.NextDouble() <= 0.069)
+            if (Loader.Version.ToString().Contains("6.9") || 0.069.Chance())
```
louis1706, line 86: i thinks you need to make 6.9.Chance()

FUTURE-SL: fixed 🧠

NotIntense (approved):
While I don't think this is necessary,  looks fine to me

NaoUnderscore:
`MathExtensions` already exists

PR #2215, opened 2023-11-08
Custom candies

VALERA771:


Exiled.CustomItems/API/Features/CustomCandy.cs
```diff
@@ -0,0 +1,200 @@
...
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">Throws when <see langword="value"/> is less that 0 or above 100.</exception>
+        public override float Weight
+        {
+            get => weight;
+            set
+            {
+                if (value is > 100 or < 0)
+                    throw new ArgumentOutOfRangeException(nameof(value), value, "Value must be above zero and less that 100.");
+
```
louis1706, line 53: Why ?

louis1706 (changes requested):
also have you tested it when you have multiple candy

VALERA771:
I've asked QA to test it cuz I don't have enough time

Exiled.CustomItems/API/Features/CustomCandy.cs
VALERA771: It can equals to 0 but here weight is a chance to pick up a custom candy instead of normal

louis1706: actually you remove the possibility to make it greater than 100 there no need for that

warquys: Use the `Clamp` fonction of unity this will avoid errors that would break the workflow and avoid config errors

VALERA771:
Will be remaked

PR #2217, opened 2023-11-10
Add Notification API

FUTURE-SL:


louis1706:
Sorry this PR can't be accepted it's will break likely all server config (i keep it open in case other devboy thinks otherwise but that my opinion)

NaoUnderscore:
`TextDisplay` already exists.

PR #2225, opened 2023-11-11
Is it a flashlight?

VALERA771:


louis1706:
i don't thinks it's needed there no difference between FlashLight and Lantern

VALERA771:
Yeah, but it's 2 separate items. I think we can create just a base class from which inhernits both flashlight and lantern

Exiled.API/Features/Items/Flashlight.cs
```diff
@@ -54,18 +56,25 @@ public bool Active
...
+            NextAllowedTime = NextAllowedTime
         };
 
         /// <summary>
-        /// Returns the Flashlight in a human readable format.
+        /// Returns the Lantern in a human readable format.
         /// </summary>
-        /// <returns>A string containing Flashlight-related data.</returns>
-        public override string ToString() => $"{Type} ({Serial}) [{Weight}] *{Scale}* |{Active}|";
+        /// <returns>A string containing Lantern-related data.</returns>
```
louis1706, line 77: not a lantern but can be both

louis1706 (changes requested):

louis1706:
> Yeah, but it's 2 separate items. I think we can create just a base class from which inhernits both flashlight and lantern

yeah that what i was thinking

Exiled.API/Features/Items/Flashlight.cs
```diff
@@ -54,18 +56,25 @@ public bool Active
...
+        /// <inheritdoc/>
+        public override Item Clone() => new Flashlight
+        {
+            IsEmittingLight = IsEmittingLight,
+            NextAllowedTime = NextAllowedTime
         };
 
         /// <summary>
-        /// Returns the Flashlight in a human readable format.
+        /// Returns the Lantern in a human readable format.
```
louis1706, line 75: same for this i would this you would seen it sorry

VALERA771:
Breaking changes are allowed? Cuz I'm changing Base property type

louis1706:
> Breaking changes are allowed? Cuz I'm changing Base property type

it's not a breaking change since it's already change inside Halloween also it's Base propperty it's should't break too mush

Exiled.API/Features/Items/Flashlight.cs
```diff
@@ -31,19 +32,20 @@ public Flashlight(FlashlightItem itemBase)
...
         /// </summary>
-        public new FlashlightItem Base { get; }
+        /// <remarks>Can be <see cref="FlashlightItem"/> or <see cref="LanternItem"/>.</remarks>
+        public new ToggleableLightItemBase Base { get; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether the flashlight is turned on.
+        /// Gets or sets a value indicating whether the item is emitting light.
         /// </summary>
-        public bool Active
```
louis1706, line 46: don't do this breaking change is not needed now

louis1706: you could make it Obsolete if you want

Exiled.API/Features/Items/Flashlight.cs
```diff
@@ -31,19 +34,28 @@ public Flashlight(FlashlightItem itemBase)
         /// Initializes a new instance of the <see cref="Flashlight"/> class, as well as a new Flashlight item.
         /// </summary>
         internal Flashlight()
```
louis1706, line 36: add ItemType item argument to choose wish Item will be initialised

VALERA771:
build go brrr

louis1706 (approved):

PR #2226, opened 2023-11-11
Fixed Event StartingRecallEventArgs's IsAllowed issue.

XKaguya:
StartingRecallEventArgs's IsAllowed won't stop the event. Patch at RagdollAbilityBase could solve this.

### DIFF
--- a/Exiled.Events/EventArgs/Scp049/StartingRecallEventArgs.cs
+++ b/Exiled.Events/EventArgs/Scp049/StartingRecallEventArgs.cs
@@ -31,15 +31,15 @@ namespace Exiled.Events.EventArgs.Scp049
         /// <param name="isAllowed">
         ///     <inheritdoc cref="IsAllowed" />
         /// </param>
-        public StartingRecallEventArgs(Player target, Player scp049, Ragdoll ragdoll, bool isAllowed = true)
+        public StartingRecallEventArgs(BasicRagdoll ragdoll, Player player, bool isAllowed = true)
         {
-            Target = target;
-            Player = scp049;
-            Scp049 = scp049.Role.As<Scp049Role>();
-            Ragdoll = ragdoll;
+            Ragdoll = Ragdoll.Get(ragdoll);
+            Target = Ragdoll.Owner;
+            
+			Scp049 = player.Role.As<Scp049Role>();
+            Player = player;
             IsAllowed = isAllowed;
         }
-
         /// <summary>
         ///     Gets the player who's getting recalled.
         /// </summary>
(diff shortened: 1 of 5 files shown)


XKaguya:
The part of `                // base.CurRagdoll
                new(OpCodes.Ldarg_1),
                new(OpCodes.Callvirt, PropertyGetter(typeof(RagdollAbilityBase<Scp049Role>), nameof(RagdollAbilityBase<Scp049Role>.CurRagdoll))),` currently will lead into crash , dont merge before if i can fix this bug.Another option is to give up acquiring Target players, although I know this is not realistic, lol

louis1706:
this will also be call for Scp0492

XKaguya:
> this will also be call for Scp0492 'RagdollAbilityBase'

yes , that is what the arg used for.Actually it will be call by Scp3114, but to doing limits on the Player , then it can be avoided.

XKaguya:
The problem about previous patch's ```IsAllowed``` is not vailed is because NW using non-zero as error code.Meanwhile ```Transpiler``` can't modify the return value.
So the patch must use ```Prefix``` instead of ```Transpiler ```.
Things are all done now.This event can get ```Target```, ```Player```, ```Ragdoll``` and ```IsAllowed```.

louis1706:
> IsAllowed

Transpiler can modify return value ???

XKaguya:
> > IsAllowed
> 
> Transpiler can modify return value ???

uh , letme try that lol , forget its just stack

XKaguya:
> > IsAllowed
> 
> Transpiler can modify return value ???

Alright. now i've rewritten as Transpiler.And tested.

Misaka-ZeroTwo:
Please target dev branch.

Exiled.Events/EventArgs/Scp049/StartingRecallEventArgs.cs
```diff
@@ -1,66 +1,66 @@
...
+        /// <param name="scp049">
+        ///     <inheritdoc cref="Player" />
+        /// </param>
+        /// <param name="ragdoll">
+        ///     <inheritdoc cref="Ragdoll" />
+        /// </param>
+        /// <param name="isAllowed">
+        ///     <inheritdoc cref="IsAllowed" />
+        /// </param>
+        public StartingRecallEventArgsMod(Player player, Ragdoll ragdoll, bool isAllowed = true)
```
Misaka-ZeroTwo, line 34:
```suggestion
        public StartingRecallEventArgs(Player player, Ragdoll ragdoll, bool isAllowed = true)
```

Misaka-ZeroTwo (changes requested):

XKaguya:
> Please target dev branch.

OK I'll do it after i get up

Misaka-ZeroTwo:
You will need to create a separate PR to prevent conflicts.

PR #2236, opened 2023-11-12
Add Remote Admin emojies API

FUTURE-SL:


louis1706:
Can you give some more Description about why and what it's will do because i actually don't getting it

PR #2239, opened 2023-11-13
Update `YamlDotNet` version.

VALERA771:
I think we can start using new YamlDotNet version.

Yeah, it's breaking changes (some features will be removed) and it requiers lots of discussions so I'll put on draft know.

I can cofirm that this update won't broke configs and completly tested


VALERA771:
Removing from darft, will be merged when available

PR #2256, opened 2023-11-14
`[Exiled.Events]` Added `PlacingPickupIntoPocketDimension` event.

SnivyFilms:
Pickups have different behaviors in 106's pocket dimension. They have flags that determine if they have a 50%, 30%, or 0% chance of being deleted, playing sounds when a pickup does leave the pocket dimension, and determines where it will be ejected back into the rest of the facility.

I have no idea how to make a transpiler for this, but my hope is that this makes a solid start.

### DIFF
--- a/Exiled.Events/EventArgs/Map/PocketDimensionPickupEventArgs.cs
+++ b/Exiled.Events/EventArgs/Map/PocketDimensionPickupEventArgs.cs
@@ -1,0 +1,71 @@
+// -----------------------------------------------------------------------
+// <copyright file="PocketDimensionPickupEventArgs.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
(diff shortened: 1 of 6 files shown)


louis1706:
i will see what i can do there

Exiled.Events/EventArgs/Map/PocketDimensionPickupEventArgs.cs
```diff
@@ -0,0 +1,83 @@
...
+        public bool ShouldRemove
+        {
+            get => PocketItem.Remove;
+            set => PocketItem.Remove = value;
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the warning sound for the pickup should be sent.
+        /// </summary>
+        public bool ShouldWarning
```
louis1706, line 65: https://github.com/NikkiGardiner1/EXILED/pull/1

NikkiGardiner1: Merged it into my dev

PR #2281, opened 2023-11-19
Add Camera::Scp079s

FUTURE-SL:


louis1706:
Camera.Get(x => x.IsBeingUsed) 

I will refuse the new propperty since it's actually not needed and can be easily optain

PR #2290, opened 2023-11-21
Some Scp914 changes

VALERA771:
- Added Scp914::Doors
- Moved useless Scp914Controller instance in some events


Exiled.Events/EventArgs/Scp914/UpgradingInventoryItemEventArgs.cs
```diff
@@ -48,7 +46,8 @@ public UpgradingInventoryItemEventArgs(Player player, ItemBase item, Scp914KnobS
         /// <summary>
         ///     Gets the <see cref="Scp914Controller" /> instance.
         /// </summary>
-        public Scp914Controller Scp914 { get; }
+        [Obsolete("Use Scp914::Scp914Controller instead.")]
```
louis1706, line 49: You can remove the Obsolete i thinks we will keep this Scp914 directly in Event since it's make things a lot easier (also you mostly always need it with Scp914Event)

louis1706 (changes requested):

Exiled.Events/EventArgs/Scp914/UpgradingInventoryItemEventArgs.cs
VALERA771: But you can use Scp914 class (what constructor is doing). I don't really think that it needed here

louis1706: it's useless breaking change for now

louis1706 (changes requested):

louis1706 (approved):

PR #2291, opened 2023-11-21
InteractedScp330 event

VALERA771:
Need someone else to test cuz I'm unable to test it. (this and InteractingScp330 event are not calling on my server, idk why, no patch errors)


VALERA771:
Last time it wasn't working, so it was on draft

ALEXWARELLC:
> Last time it wasn't working, so it was on draft

So am I able to perform tests on it?

VALERA771:
You can try

BoltonDev:
You should add `[EventPatch(typeof(Scp330), nameof(Scp330.InteractedScp330))]`

Misfiy:
What happens in the case players hands is severed?
If it still calls, please add a variable for this
If it doesn't, maybe add it or make it another event

VALERA771:
> What happens in the case players hands is severed? 

Just calls. Event does not depend on this

>If it still calls, please add a variable for this If it doesn't, maybe add it or make it another event

It can be easily done by checking if player has a severed hands effect. Do we really need this?

Misfiy:
> It can be easily done by checking if player has a severed hands effect. Do we really need this?

Would be nice to have it be simple on whether players hands were severed
if you do not need to modify your transpiler then cool just add a bool in the event args for if player has severed effect

VALERA771:
I'll close this for now cuz I saw that we are changing this method a lot so I idk how to correctly implement this event

PR #2208, opened 2023-11-05
Add StranglingEventArgs and StranglingFinishedEventArgs + YamlIgnore Properties Fix

Redforce04:
## Adds 2 new Scp3114 Events:
- StranglingEventArgs
  -  Called when an Scp 3114 player begins to strangle another player (deniable).
- StranglingFinishedEventArgs
  - Called when an Scp 3114 player stops strangling another player (not deniable).

Adds the corresponding Transpiler Patches (2), EventArgs, and EventHandlers.

### Problem:  
- Configs utilizing Yaml Ignore may throw exceptions when being serialized, because the CopyProperties extension does not account for this. This primarily happens when a config property is missing a setter. 
  - This is bad practice and utilization of the YamlDotNet api.
## Solutions
  - CopyProperties will ignore properties that have the [YamlIgnoreAttribute], solving the problem. 
  - Static Properties will not be effected.

## Fix minor styling issues and typos
- Scp3114Handlers: `Digiusing` has been corrected to `Disguising`.
- Spacing has been fixed.

PR #2153, opened 2023-10-19
Various changes I have needed

joker-119:
Simple enough  changes, feel free to keep em, or don't.

PR #2294, opened 2023-11-22
Add StranglingEventArgs.

XKaguya:


louis1706:
Conflict between #2294 #2208 #2153

PR #2296, opened 2023-11-23
NPC Features + VSR

creepycats:
Use "ID_Dedicated" to ensure an NPC is VSR Compliant (Uses Unverified, DedicatedHost instances, which will not take up slots or show on the Player List)

Added Multiple Functions for NPC characters to Look at Points, and use a Held Firearm (Shooting, ADS, Flashlight, Reloading)


creepycats:
Going to change smth 1s

PR #2298, opened 2023-11-25
who wrote this?

Artyomus1337:
tested


louis1706 (changes requested):
Can you fix the other documentation that is on the same file ? (Also it is possibly my fault)

Artyomus1337:
yep

Artyomus1337:
that's it

louis1706 (approved):

PR #2303, opened 2023-11-29
Inroducing... "Async" events

VALERA771:
It just adds an overload to a basic `Event` and `Event{T}` classes and allows to add a methods which return `IEnumerator<float>` (coroutines) to be attached as a event method!

It's partly tested on `WaitingForPlayers` event but would be great if somebody else also test it


VALERA771:
I can say tested

PR #2316, opened 2023-12-05
Item events

VALERA771:
Added:
`Player::AddingItem`
`Player::RemovingItem`

Current tasks:
- [x] Fix and test RemovingItem event
- [x] Test AddingItem event
- [x] Add transpiler comments
- [ ] Imagine a way to make exiled methods also trigger these events
- [x] Just in case, test moving InventoryControl patch to events' patches
- [x] Check if player is dropping item and we can not remove it from Items dictionary


louis1706:
> Added: `Player::AddingItem` `Player::RemovingItem`
> 
> Current tasks:
> 
> * [ ]  Fix and test RemovingItem event
> * [ ]  Test AddingItem event
> * [ ]  Add transpiler comments
> * [ ]  Imagine a way to make exiled methods also trigger these events
> * [ ]  Just in case, test moving InventoryControl patch to events' patches
> 
> _p.s. don't even try to execute workflows..._

sorry i always execute workflow before reading

VALERA771:
PR finished. Last task is not really nessesary

NaoUnderscore:
> PR finished. Last task is not really nessesary

I'd like to point out that `apis-rework` already contains `Adding` and `Removing` `Item` events.
You may take a look to how they've been made and eventually merge your changes if needed.

VALERA771:
It would be easier to start from the beggining, I think. So I'll close this PR

PR #2320, opened 2023-12-07
Scp3114AttackAhpFix

louis1706:


TtroubleTT:
This creates a bug where SCP 3114 no longer gets hume shield from attacking anyone

louis1706:
image

PR #2342, opened 2023-12-21
Lots of new API

VALERA771:


Exiled.Events/Config.cs
```diff
@@ -103,5 +103,11 @@ public sealed class Config : IConfig
         /// </summary>
         [Description("Whether or not to log RA commands.")]
         public bool LogRaCommands { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the SCP-956 should be a capibara.
+        /// </summary>
+        [Description("Whether the SCP-956 should be a capibara.")]
+        public bool Is956Capybara { get; set; } = false;
```
louis1706, line 111: Sorry i can't accept that on Exiled config xD Even if i really like it it's would need to be separated plugin lol

louis1706 (changes requested):

Exiled.Events/Config.cs
VALERA771: That's sad 😢

louis1706 (approved):

PR #2355, opened 2023-12-24
SpawnReasons::Vocalize

InziDeveloperMode:
Added new SpawnReason for SCP-1507


louis1706:
This will never be used

PR #2360, opened 2023-12-25
Coffee API

VALERA771:
- Base class
- List control patch (tested)
- Event for ~interacting~ drinking coffee (tested)


Exiled.API/Features/Coffee.cs
```diff
@@ -0,0 +1,122 @@
...
+            set => Base._author = value;
+        }
+
+        /// <summary>
+        /// Gets or sets a color of a drink in a cup.
+        /// </summary>
+        public Color DrinkColor
+        {
+            get => Base._drinkColor;
+            set => Base._drinkColor = value;
```
louis1706, line 99: Color can't be change i thinks (probably)

VALERA771: Maybe, I'll check and maybe change setter

PR #2367, opened 2023-12-28
Fixed installer downloading pre-release version even when set to false

BoltonDev:


louis1706 (approved):

Exiled.Installer/Program.cs
```diff
@@ -280,20 +279,27 @@ static PathResolution TryParse(string s)
...
-                    continue;
+            var enumerable = releases.ToList();
 
-                if (targetVersion.IsPreRelease && !args.PreReleases)
-                    continue;
+            foreach (var release in enumerable)
+            {
+                if (targetVersion != null)
+                {
+                    if(targetVersion == new Version(release.TagName))
```
NaoUnderscore, line 290:
```suggestion
                    if (targetVersion == new Version(release.TagName))
```

Exiled.Installer/Program.cs
```diff
@@ -280,20 +279,27 @@ static PathResolution TryParse(string s)
...
 
-            foreach (Release r in releases)
-            {
-                if (targetVersion != new Version(r.TagName))
-                    continue;
+            var enumerable = releases.ToList();
 
-                if (targetVersion.IsPreRelease && !args.PreReleases)
-                    continue;
+            foreach (var release in enumerable)
```
NaoUnderscore, line 286: Do not use `var`.

NaoUnderscore (changes requested):

Exiled.Installer/Program.cs
BoltonDev: Oh yeah mb. I think it might be a good idea to prevent their use by adding an option in .editorconfig

NaoUnderscore: Done

NaoUnderscore (approved):

PR #2373, opened 2024-01-01
Exiled.CustomRoles/API/Extensions::HasCustomrole

InziDeveloperMode:
-  Tested

### DIFF
--- a/Exiled.CustomRoles/API/Extensions.cs
+++ b/Exiled.CustomRoles/API/Extensions.cs
@@ -61,6 +61,27 @@ namespace Exiled.CustomRoles.API
         public static void Register(this CustomRole role) => role.TryRegister();
 
         /// <summary>
+        /// Checks if a player has a specific custom role.
+        /// </summary>
+        /// <param name="player">The <see cref="Player"/> to check for the custom role.</param>
+        /// <param name="roleName">The name of the custom role to check for.</param>
+        /// <returns><see langword="true"/> if the player has the custom role; otherwise, <see langword="false"/>.</returns>
+        public static bool HasCustomRole(this Player player, string roleName)
+        {
+            if (player is null)
+                throw new ArgumentNullException(nameof(player));
+
+            if (string.IsNullOrEmpty(roleName))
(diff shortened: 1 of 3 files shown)


Exiled.CustomRoles/API/Extensions.cs
```diff
@@ -60,6 +60,25 @@ public static void Register(this IEnumerable<CustomRole> customRoles)
...
+        /// <returns><see langword="true"/> if the player has the custom role; otherwise, <see langword="false"/>.</returns>
+        public static bool HasCustomRole(this Player player, string roleName)
+        {
+            if (player is null)
+                throw new ArgumentNullException(nameof(player));
+
+            if (string.IsNullOrEmpty(roleName))
+                throw new ArgumentException("Role name cannot be null or empty.", nameof(roleName));
+
+            CustomRole customRole = CustomRole.Registered.FirstOrDefault(role => role.Name.Equals(roleName, StringComparison.OrdinalIgnoreCase));
```
VALERA771, line 77: Use CustomRole::TryGet

VALERA771 (commented):
Also, cuz new custom roles API will be released soon, can be useless

InziDeveloperMode:
> Also, cuz new custom roles API will be released soon, can be useless

but is there already this function in the new customrole API? Otherwise I might just switch branches.

InziDeveloperMode:
Because if this function doesn't exist, it would be very useful to include it in exiled.

VALERA771:
> Because if this function doesn't exist, it would be very useful to include it in exiled. 

If it's not there, switching branch won't help. This PR need to be remakes cuz APIs branch contains lots of changes

InziDeveloperMode:
> > Because if this function doesn't exist, it would be very useful to include it in exiled. 
> 
> If it's not there, switching branch won't help. This PR need to be remakes cuz APIs branch contains lots of changes

I explained myself badly then, I meant that I will adapt the code to the new branch using the new customroles API

InziDeveloperMode:
otherwise I will make a new pull request, as you prefer

PR #2375, opened 2024-01-03
Added a new extension method `CustomRole::HasCustomrole`

InziDeveloperMode:


NaoUnderscore:
Do not place it between same methods with different signature in order to maintain the same pattern.
Aside from that, we may consider to make it an extension method instead, which would not be so much useful.
Furthermore, `Pawn` class exists, you could add it to that.

InziDeveloperMode:
There is a small problem in the code, I'll fix it this afternoon.

NaoUnderscore:
Move eventual additions to #2378 contributing to it instead.

PR #2389, opened 2024-01-14
`[Exiled::Events]` Added new commands for installing plugins, changing or printing config values

VALERA771:
`[Exiled::API]` Added extension 
```csharp
StringExtensions::FromSnakeCase(string)
```
for converting strings from snake_case to NotSnakeCase


VALERA771:
Install command tested. Other will be tomorrow

Exiled.Events/Commands/ConfigValue/Set.cs
```diff
@@ -0,0 +1,97 @@
...
+                return false;
+            }
+
+            if (plugin.Config.GetType().GetProperty(propertyName) is not PropertyInfo property)
+            {
+                response = $"Config value not found: {propertyName} ({propertyName.ToSnakeCase()})";
+                return false;
+            }
+
+            object newValue = Convert.ChangeType(arguments.At(1), property.PropertyType);
```
louis1706, line 82: Add a try catch in case the value give is not correct

louis1706 (changes requested):

Exiled.Events/Commands/ConfigValue/Set.cs
VALERA771: I think it'll return null instead of throwing. I'll check that

louis1706: > I think it'll return null instead of throwing. I'll check that

It's throwing if invalid

VALERA771:
All commands have been tested

louis1706 (approved):

Vladislav-CS (approved):

Misfiy (approved):

Vladislav-CS:
Why DoNotMerge?

Misfiy:
@NaoUnderscore (as asked by vlad) why do not merge?

NaoUnderscore:
Because I'd like to wait before merging Plugin Manager/Installer-like features.
We're going to implement it through our Web APIs.

VALERA771:
@NaoUnderscore

PR #2394, opened 2024-01-22
Easy way to Goto AppData

Veslydev:
Uses %UserName% Instahead of (Your user) and make it easy Did in all langues


1EnesBaturKaza:
bad

1EnesBaturKaza:
good

dogy1369:
usefull

NotIntense:
You could also add (for Window users), press Win + R to bring up the Run box, then type %appdata% (not anything else) for it to instantly take you there.

NaoUnderscore:
You may rename it to `User` instead of `UserName`, following MS docs.

Veslydev:
> You may rename it to `User` instead of `UserName`, following MS docs.

User is not working i tried and it was not working, i checked the MS Docs And The best is put UserName on it

XKaguya:
I think `%APPDATA%` is the easiest way to goto `AppData\Roaming`. Which type in explorer will directly into `C:\Users\Administrator\AppData\Roaming`

PR #2412, opened 2024-01-29
`[Exiled::Events]`, `[Exiled::API]` Fixing for new release

VALERA771:


louis1706:
i will handle it since it's just need to merge dev to Api-Rework

PR #2419, opened 2024-02-01
`[Exiled::API]` Adding a game's constant patching

VALERA771:
Adding a transpiler-based editing of game's constants. In current system can be some mistakes with replacing other values that has just the same value cuz we can't check whether the constant is used. (everyone know this tbh).

Opened for all suggestions. Now I can't test it so hope that someone will do this faster.


NaoUnderscore:
Looks interesting, I'll support you if needed by making some integrations with existing `Core` features if needed.
Keep us up to date.

VALERA771:
Patches are partially tested. No errors on setting role while/patching methods.
Current methods could be optimized by using one patch for method and/or checking if value is presented in method body. I'm currently working on second.

NaoUnderscore:
@VALERA771 Keep me up to date

VALERA771:
Now I'm on final test part. I've implemented patching only for methods which have constant value in them. Currently I want to optimize patch cuz I've noticed that there can be some errors in several types. I'm going to fix it and add support for all constants which are presented in Playable SCPs classes

VALERA771:
1. Added a class to edit Readonly fields. Working same as Constants patching so I don't need that it requers tesing.
2. Modified transpiler so now it's looking better.
3. Added some new properties to some Items classes.
4. I'm going to test new version, if test is successful, I'll make it normal PR (not draft).

VALERA771:
I've decided to remove `ReadonlyField{T}` cuz after a long time searching for readonly field that we can edit (and it'll make profit), I haven't found anything.

Misfiy:
@VALERA771 is this ready to be merge?

VALERA771:
It's tested so I think yes

Exiled.API/Features/Core/ConstProperty.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        /// <summary>
+        /// Finalizes an instance of the <see cref="ConstProperty{T}"/> class.
+        /// </summary>
+        ~ConstProperty()
+        {
+            List.Remove(this);
+
+            foreach (MethodInfo methodInfo in TypesToPatch.SelectMany(x => x.GetMethods().Where(y => y.DeclaringType != null && y.DeclaringType == x)))
```
Misfiy, line 52: Could the methods to patch be saved ?

PR #2427, opened 2024-02-02
[Exiled.API] [Addition] Player -> GetNear & GetFar

Misfiy:


louis1706:
i don't like the idea to make it directly inside the API 
there could be many different way to approach this

NaoUnderscore:
Fix build errors

PR #2434, opened 2024-02-04
[Exiled::API] Added AccountAge to Features.Player

ghost:


creepycats:
Just a thought but shouldn't this be only queried once? Every subsequent request is just an extra network call to fetch the AccountAge when you should only need to call it once and store the result. (Unless i'm stupid and stuff)

louis1706:
> Just a thought but shouldn't this be only queried once? Every subsequent request is just an extra network call to fetch the AccountAge when you should only need to call it once and store the result. (Unless i'm stupid and stuff)

i actually don't feel like we should make an request to a website inside a property but i will let Nao see for it

VALERA771:
I think w can use directly steam information and don't request any third-party websites

creepycats:
The issue is Ive looked through Facepunch Steamworks.NET and there isnt a way to get the account age. If there is some secondary steam API dll we can use then try that but I dont think you can get it natively

VALERA771:
We can access it through xml=1 (https://steamcpmmunity.com/profiles/{steamId}/?xml=1)

creepycats:
Still, requesting a website isnt the greatest idea.

Then again not much else you can do.

louis1706:
This has been refused because of the need of an actual HTTP request and it's conld lead to Lag if used during the gameplay

PR #2449, opened 2024-02-10
Added `Scp3114::Slapped` Event

CatyFoxt:
Added Scp3114 Slapped Event


VALERA771:
We prefer to use Hurt method cuz it's more common and covering all such events. I don't really think that this is useful

CatyFoxt:
@VALERA771, I saw that Scp939 has a Clawed event. That is why i created this event. And it is for the helpful for new developers because it is much more spesific for just Slapping. Hurt method can contains all type of hurting that Scp3114 can do.

VALERA771:
> @VALERA771, I saw that Scp939 has a Clawed event. That is why i created this event. And it is for the helpful for new developers because it is much more spesific for just Slapping. Hurt method can contains all type of hurting that Scp3114 can do.

Clawed event is also firied when you claws "air" (aka no target). As far as I can see, your event won't work like this

CatyFoxt:
@VALERA771, You are wrong the event i did is firing every time Scp3114 tries to Slap. This contains **air, targets and every slapping action.**

VALERA771:
Okay, then yeah

Exiled.Events/Patches/Events/Scp3114/Slapped.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// </summary>
+    [EventPatch(typeof(Scp3114), nameof(Scp3114.Slapped))]
+    [HarmonyPatch(typeof(Scp3114Slap), nameof(Scp3114Slap.DamagePlayers))]
+    internal class Slapped
+    {
+        private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions, ILGenerator generator)
+        {
+            List<CodeInstruction> newInstructions = ListPool<CodeInstruction>.Pool.Get(instructions);
+
+            newInstructions.InsertRange(0, new CodeInstruction[]
```
louis1706, line 33: You call your Event Slapp**ed** but it's call before the DamagePlayer method is executed and so we cannot differentiate a player who have get it's damage or not

louis1706 (changes requested):

Exiled.Events/Patches/Events/Scp3114/Slapped.cs
louis1706: also it's do not contain a lot of information

Exiled.Events/EventArgs/Scp3114/SlappedEventArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </param>
+        public SlappedEventArgs(Player player)
+        {
+            Player = player;
+            Scp3114 = Player.Role.As<Scp3114Role>();
+        }
+
+        /// <summary>
+        /// Gets the SCP-3114.
+        /// </summary>
```
VALERA771, line 34:
```suggestion
        /// <inheritdoc/>
```

Exiled.Events/Handlers/Scp3114.cs
```diff
@@ -38,6 +38,11 @@ public static class Scp3114
         /// </summary>
         public static Event<RevealedEventArgs> Revealed { get; set; } = new();
 
+        /// <summary>
+        /// Invoked when SCP-3114 slaps a player.
```
VALERA771, line 42:
```suggestion
        /// Invoked after SCP-3114 slaps a player.
```

Exiled.Events/Handlers/Scp3114.cs
```diff
@@ -72,6 +77,12 @@ public static class Scp3114
         /// <param name="ev">The <see cref="RevealedEventArgs" /> instance.</param>
         public static void OnRevealed(RevealedEventArgs ev) => Revealed.InvokeSafely(ev);
 
+        /// <summary>
+        /// Called when slapped.
```
VALERA771, line 81:
```suggestion
        /// Called after SCP-3114 slaps a player.
```

VALERA771 (commented):
Some doc suggestions to make them "more clawed"

CatyFoxt:
@louis1706, I changed the index of the method and fixed the orders of the using lines.

louis1706:
> @louis1706, I changed the index of the method and fixed the orders of the using lines.

i thinks this place would be better to put the code 
			SlappedEventArgs ev = new(Player.Get(Owner), _syncAttack);
			Handlers.Scp3114.OnSlapped(ev)

			float num = this._humeShield.CurValue + ev.HumeShieldReward;
			this._humeShield.CurValue = Mathf.Min(num, this._humeShield.MaxValue);

louis1706:
you can use the const to take the HumeShield value

Exiled.Events/Patches/Events/Scp3114/Slapped.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                // SlappedEventArgs ev = new(player, AttackResult, target);
+                new(OpCodes.Newobj, GetDeclaredConstructors(typeof(SlappedEventArgs))[0]),
+
+                // Handlers.Scp3114.OnSlapped(ev)
+                new(OpCodes.Call, Method(typeof(Scp3114), nameof(Scp3114.OnSlapped))),
+            });
+
+            for (int z = 0; z < newInstructions.Count; z++)
+                Log.Info($"{newInstructions[z].opcode} {newInstructions[z].operand} : {string.Join(", ", newInstructions[z].labels.Select(x => newInstructions.FindIndex(y => y.operand == (object)x)))}");
```
NaoUnderscore, line 68: Remove debug lines

NaoUnderscore (changes requested):

PR #2453, opened 2024-02-11
Small rework of background api & consistency of it 

louis1706:


Exiled.API/Features/Lift.cs
```diff
@@ -390,7 +390,7 @@ public void Lock(DoorLockReason lockReason = DoorLockReason.Isolation)
         public void Lock(float duration, DoorLockReason lockReason = DoorLockReason.Isolation)
         {
             Status = ElevatorSequence.DoorClosing;
-            ChangeLock(lockReason);
+            Lock(duration, lockReason);
```
VALERA771, line 393: Calling same method?

Exiled.Events/Patches/Generic/TargetOffset.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace Exiled.Events.Patches.Generic
 {
+#pragma warning disable IDE0060 // Supprimer le paramètre inutilisé
```
VALERA771, line 10: What this warning about? Seems to be IDE specific so idk if we really need it

Exiled.API/Features/Lift.cs
louis1706: Yup

Exiled.Events/Patches/Generic/TargetOffset.cs
IkaOverride: I'm kinda late but it says "Remove unused parameter". so it removes parameters being flagged when they aren't used and they do not start with the _ symbol

NaoUnderscore, line 10: I don't want french in Exiled's code 👎🏼

NaoUnderscore (changes requested):

Exiled.Events/Patches/Generic/TargetOffset.cs
louis1706, line 10:
```suggestion
```

NaoUnderscore (approved):

PR #2456, opened 2024-02-14
Made async events safe

Banalny-Banan:
Currently async events are unsafe, because Timing.Run Coroutine() just ignores exceptions, and catch{} is never invoked. I added a manual coroutine enumerator to catch exceptions and log them.


Exiled.Events/Features/Event{T}.cs
```diff
@@ -197,13 +197,37 @@ internal void InvokeAsync(T arg)
...
                 try
                 {
-                    Timing.RunCoroutine(handler(arg));
+                    Timing.RunCoroutine(SafeCoroutineEnumerator(handler(arg), handler));
                 }
                 catch (Exception ex)
                 {
                     Log.Error($"Method \"{handler.Method.Name}\" of the class \"{handler.Method.ReflectedType.FullName}\" caused an exception when handling the event \"{GetType().FullName}\"\n{ex}");
                 }
             }
```
louis1706, line 206: No more need to have the Try Catch here now then

louis1706 (changes requested):

louis1706 (approved):

Exiled.Events/Features/Event{T}.cs
```diff
@@ -195,14 +195,31 @@ internal void InvokeAsync(T arg)
...
+                        break;
+                    current = coroutine.Current;
                 }
                 catch (Exception ex)
                 {
                     Log.Error($"Method \"{handler.Method.Name}\" of the class \"{handler.Method.ReflectedType.FullName}\" caused an exception when handling the event \"{GetType().FullName}\"\n{ex}");
+                    yield break;
                 }
+
+                yield return (float)current;
```
louis1706, line 222: I feel like the Error would come from there and not being catch by the try catch

PR #2462, opened 2024-02-21
List instead of IEnumerable when not needed

louis1706:


Exiled.CustomModules/API/Features/CustomGamemodes/GameState.cs
```diff
@@ -73,7 +73,7 @@ public abstract class GameState : EActor, IAdditiveSettings<GameModeSettings>
         /// <summary>
         /// Gets all <see cref="PlayerState"/> instances.
         /// </summary>
-        public IEnumerable<PlayerState> PlayerStates => playerStates;
+        public List<PlayerState> PlayerStates => playerStates;
```
NaoUnderscore, line 76: No, we have a method to safely add and remove `PlayerState` objects.

NaoUnderscore (changes requested):

Exiled.CustomModules/API/Features/CustomGamemodes/GameState.cs
louis1706, line 76:
```suggestion
        public IEnumerable<PlayerState> PlayerStates => playerStates;
```

PR #2467, opened 2024-02-22
Simplify door/lift api

louis1706:
Removing static Method than shouldn't be there


Exiled.API/Features/Lift.cs
```diff
@@ -264,106 +264,6 @@ public override Quaternion Rotation
...
-        /// <seealso cref="Player.EnableEffect(EffectType, float, bool)"/>
-        public static bool TryMeltPlayer(Player player)
-        {
-            if (player.Position.y is >= 200 or <= -200)
-                return false;
-
-            player.EnableEffect(EffectType.Decontaminating);
-
-            return true;
-        }
```
louis1706, line 365: Why is this removed ?
because now when it's Decontamination NW is constantly checking if it's give Decontaminating Effect so it's now useless to have this method

Exiled.API/Features/Doors/Door.cs
```diff
@@ -128,6 +129,15 @@ public bool IsOpen
             set => Base.NetworkTargetState = value;
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the door is close.
+        /// </summary>
+        public bool IsClose
```
NaoUnderscore, line 135: Not sure whether it's necessary, however just rename it to `IsClosed` anyway.

NaoUnderscore (changes requested):
What about all the `Lock` methods?

Exiled.API/Features/Doors/Door.cs
louis1706, line 135:
```suggestion
        public bool IsClosed
```

NaoUnderscore:
Looks fine I guess

PR #2471, opened 2024-03-04
[Exiled.Events & Exiled.API] Role Spawn Fix

Misfiy:
* Fix spawning in wrong location
* Add TryGetRoleBase<T> by Yamato

Fix requires testing


Misfiy:
This does not fix the issue.

PR #2479, opened 2024-03-11
`[Exiled::Events]` Scp106::PassingDoor event

VALERA771:


louis1706:
Very nice event that was a needed event

- [ ] Is The Event Spam ?
- [ ] Does it correctly prevent client to go trough It ?

also can you fix build error please

VALERA771:
Event is calling multiple times

PR #2488, opened 2024-03-19
Update TeslaGate.cs

Banalny-Banan:
Added TeslaGate.GetClosest()


louis1706:
GetNearestEntities exist now i don't thinks this method is now needed

louis1706:
GetNearestEntity exist and should did the same behaviour

PR #2490, opened 2024-03-22
[API] [Change + Addition] Player Effects

Misfiy:


Exiled.API/Features/Player.cs
```diff
@@ -3310,7 +3307,44 @@ public void ChangeEffectIntensity(EffectType type, byte intensity, float duratio
             if (TryGetEffect(type, out StatusEffectBase statusEffect))
             {
                 statusEffect.Intensity = intensity;
-                statusEffect.ServerChangeDuration(duration, false);
+                statusEffect.ServerChangeDuration(duration, true);
```
louis1706, line 3310: instead of modifying the false to true make it an argument and don't modify the default value (actually false)

louis1706 (changes requested):

PR #2495, opened 2024-03-24
Added `Player::FallDamage` property

louis1706:


warquys:
Why do you use a AccessTools ? AllowUnsafeBlocks is enabled

PR #2511, opened 2024-04-13
[EXILED.API] [Additions & Changes] Added overload for Player.TryRemoveFriendlyFire

Mikihero:


Exiled.API/Features/Player.cs
```diff
@@ -1809,6 +1809,18 @@ public void TrySetCustomRoleFriendlyFire(Dictionary<string, Dictionary<RoleTypeI
         /// <returns> Whether or not the item was able to be added. </returns>
         public bool TryRemoveFriendlyFire(RoleTypeId role) => FriendlyFireMultiplier.Remove(role);
 
+        /// <summary>
+        /// Tries to remove <see cref="RoleTypeId"/> from FriendlyFire rules.
```
BoltonDev, line 1813:
```suggestion
        /// Tries to remove a <see cref="IEnumerable{T}"/> of <see cref="RoleTypeId"/> from FriendlyFire rules.
```

Exiled.API/Features/Player.cs
```diff
@@ -1809,6 +1809,18 @@ public void TrySetCustomRoleFriendlyFire(Dictionary<string, Dictionary<RoleTypeI
...
+        /// </summary>
+        /// <param name="roles"> Roles to remove from friendly fire rules. </param>
```
BoltonDev, line 1815:
```suggestion
        /// <param name="roles"> The <see cref="IEnumerable{T}"/> of <see cref="RoleTypeId"/> to remove. </param>
```

Exiled.API/Features/Player.cs
```diff
@@ -1809,6 +1809,18 @@ public void TrySetCustomRoleFriendlyFire(Dictionary<string, Dictionary<RoleTypeI
...
+        public void TryRemoveFriendlyFire(List<RoleTypeId> roles)
+        {
+            foreach(RoleTypeId roleType in roles)
+            {
+                FriendlyFireMultiplier.Remove(roleType);
+            }
+        }
```
BoltonDev, line 1822:
```suggestion
        public bool TryRemoveFriendlyFire(IEnumerable<RoleTypeId> roles) => roles.All(TryRemoveFriendlyFire);
```

Mikihero: honestly I think using a foreach here just is a bit less confusing and cleaner

BoltonDev: A method using the prefix "Try" should return a bool value to indicate whether or not the method has been executed successfully. 
`roles.All(TryRemoveFriendlyFire);` will return a bool value to indicate if all roles have been removed using the previous method Player::TryRemoveFriendlyFire(RoleTypeId). 
If you want it to be cleaner, you could also do `roles.All(role => TryRemoveFriendlyFire(role));`

PR #2536, opened 2024-04-29
`[Exiled::Permissions]` New feature and overloads

VALERA771:
1. Added a DisabledPermissions so new you can specify permissions that will be reworked from user event when group has a full perm.
2. Added a new overloads for a check permission methods with out response param.

Not tested. If it needed, I'll try it a little bit later cuz im on Mac now.


VALERA771:
Disabled permissions are tested. Other overloads not requier testing cuz they're using same methods just adding a reposnce output

VALERA771:
**IMPORTANT NOTICE**
To add new feature (disabling permissions) in your permissions file you need to re-create (delete and start server) your permissions file.

PR #2545, opened 2024-05-02
Change Config::ConfigType to Separated by default

Misfiy:


louis1706 (approved):

Exiled.API/Enums/ConfigType.cs
```diff
@@ -7,19 +7,27 @@
...
     /// The different types of configuration files distribution.
     /// </summary>
     public enum ConfigType
     {
         /// <summary>
-        /// Default distribution, every plugin will share the same config file.
+        /// Default. Removed in the future.
         /// </summary>
-        Default,
+        [Obsolete("Separated & Merged exist now.")]
```
NaoUnderscore, line 20: Why marking it as obsolete when we can just remove it?

NaoUnderscore (changes requested):

Exiled.API/Enums/ConfigType.cs
louis1706: Possibly to not breaking EVERY SCP:SL config server ?

NaoUnderscore: Then just keep it as is without marking it obsolete.
Since we're going to break all plugins anyway, and people have time to migrate from v8 to v9, it looks fine removing it without using `Obsolete`.

louis1706: > Since we're going to break all plugins anyway, and people have time to migrate from v8 to v9, it looks fine removing it without using `Obsolete`.

why would we keep it as not Obsolete if we deleted it after ?

PR #2580, opened 2024-05-29
Create GettingStarted-KR

Cocoa2219:


Misfiy (approved):
I'll approve since translator agrees on it and they have previously contributed with README-KR.md

PR #2583, opened 2024-05-30
`[Exiled::Events]` Dynamic patch back for FinishingRecall

BoltonDev:
## New event (FindingTargetEventArgs)
It's an Scp049 event which is called when SCP-049 tries to find a target to activate the sense ability.
Event isn't called for each players connected, only for those who are in the vision information of SCP-049.
⚠️ **Not tested yet**
## Dynamic patch is back for FinishingRecall event
It has been disabled due to a bug which is now fixed (https://trello.com/c/V0uHP2eV/5745-overwatch-overwatch-can-get-respawned-by-scp-049)


Misfiy:
Check for IL errors and then mark this as non draft

BoltonDev:
Finally, we can't add an event for that as it's a client side method ... I just discovered that now

Misfiy (approved):

PR #2584, opened 2024-05-30
[API] PrefabHelper

Misfiy:


VALERA771:
I've put requires testing cuz not sure that Network client will work on server :)

BoltonDev (approved):

VALERA771 (approved):

VALERA771:
Maybe add spawn to game entity? Cuz I think all of them (or at least a lot) supports spawning via Network server

BoltonDev:
> Maybe add spawn to game entity? Cuz I think all of them (or at least a lot) supports spawning via Network server

This is not needed since there is already a method in PrefabHelper

VALERA771:
> > Maybe add spawn to game entity? Cuz I think all of them (or at least a lot) supports spawning via Network server
> 
> 
> 
> This is not needed since there is already a method in PrefabHelper

But then method in controller should be removed as I think or just edits so it'll use that method

VALERA771 (approved):
Seems good but why we have spawn methods in workstation and generator?

Misfiy:
> Seems good but why we have spawn methods in workstation and generator?
Multiple others did, and I thought it would be useful to also easily getting a workstation + more easy to find

Exiled.API/Features/Generator.cs
```diff
@@ -216,6 +226,18 @@ public KeycardPermissions KeycardPermissions
             set => Base._requiredPermission = (Interactables.Interobjects.DoorUtils.KeycardPermissions)value;
         }
 
+        /// <summary>
+        /// Spawns a <see cref="Generator"/>.
+        /// </summary>
+        /// <param name="position">The position to spawn it at.</param>
+        /// <param name="rotation">The rotation to spawn it as.</param>
+        /// <returns>The Generator that was spawned.</returns>
```
Vladislav-CS, line 234:
```suggestion
        /// <returns>The <see cref="Generator"/> that was spawned.</returns>
```

Vladislav-CS (changes requested):

Vladislav-CS (approved):

PR #2598, opened 2024-06-04
[API] Primitives & BitwiseExtensions

Misfiy:
# Additions
* Add ``PrimitiveFlags`` to ``PrimitiveSettings`` constructor & ``Primitive::Create``
* ``BitwiseExtensions::SetFlag<T>(T, T, bool)``
* ``AdminToy::Position``

# Changes
* Made EffectTypeExtension use SetFlag
* Fix ``AdminToy:;Scale``

### DIFF
--- a/Exiled.API/Extensions/BitwiseExtensions.cs
+++ b/Exiled.API/Extensions/BitwiseExtensions.cs
@@ -1,0 +1,44 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="BitwiseExtensions.cs" company="Exiled Team">
+// Copyright (c) Exiled Team. All rights reserved.
+// Licensed under the CC BY-SA 3.0 license.
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Exiled.API.Extensions
+{
+    using System;
+    using System.Globalization;
+
+    /// <summary>
+    /// Extensions for Bitwise operations.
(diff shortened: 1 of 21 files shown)


Exiled.API/Extensions/BitwiseExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+            else
+            {
+                flagsInt &= ~flagInt;
+            }
+
+            return (T)(object)flagsInt;
+        }
+    }
+}
```
VALERA771, line 44: Why we even need this method?

VALERA771 (commented):
Left some comments

Exiled.API/Extensions/BitwiseExtensions.cs
Misfiy: Not needed, but makes it simpler for users to yk use rather than having to understand bitwise operations

VALERA771 (approved):
Okay

louis1706 (changes requested):
Make sure than all exiled use these new method when possible

Vladislav-CS:
These SetFlag methods don't make sense in my opinion. Also fixes should be merged into dev branch, not api

Misfiy:
It makes it simpler for beginners etc
We have other extensions to simply stuff too

PR #2614, opened 2024-06-08
dev into api-reworks

ALEXWARELLC:
We went through this.

Everything is conflict-fixed.


Misfiy (approved):

Vladislav-CS:
I think we shouldn't merge dev from alex's repo, but from original

ALEXWARELLC:
> I think we shouldn't merge dev from alex's repo, but from original

Could you elaborate a bit?

VALERA771:
I agree with vlad. Even if we need to merge dev into apis brach we need to use our branches

ALEXWARELLC:
> I agree with vlad. Even if we need to merge dev into apis brach we need to use our branches

@Misfiy

PR #2619, opened 2024-06-10
PlayGunSound at position

Banalny-Banan:
Added a static PlayGunSound to Player class that allows to play the sound at specified position, instead of playing it for a specific player

### DIFF
--- a/Exiled.API/Features/Player.cs
+++ b/Exiled.API/Features/Player.cs
@@ -1506,6 +1506,26 @@ namespace Exiled.API.Features
         public static void ReloadWhitelist() => WhiteList.Reload();
 
         /// <summary>
+        /// Plays a gun sound at the specified position.
+        /// </summary>
+        /// <param name="position">Position to play the sound at.</param>
+        /// <param name="firearmType">The type of firearm to play the sound of.</param>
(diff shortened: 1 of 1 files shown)


Misfiy:
Put this into apis-rework or dev branch

Exiled.API/Features/Player.cs
```diff
@@ -1505,6 +1505,26 @@ public static bool AddToWhitelist(string userId, bool isPermanent)
...
         public static void ReloadWhitelist() => WhiteList.Reload();
 
+        /// <summary>
+        /// Plays a gun sound at the specified position.
+        /// </summary>
+        /// <param name="position">Position to play the sound at.</param>
+        /// <param name="firearmType">The type of firearm to play the sound of.</param>
+        /// <param name="maxDistance">The maximum distance the sound can be heard from.</param>
+        /// <param name="audioClipId">The audio clip ID to play.</param>
+        public static void PlayGunSound(Vector3 position, ItemType firearmType, byte maxDistance = 45, byte audioClipId = 0)
```
Misfiy, line 1515: since this is not for a specific player and seems global it should be in Map class

PR #2625, opened 2024-06-12
`[Exiled::API]` `[Exiled::Events]` Danger fixes and additions

ghost:
- Added DangerTypeExtensions to Exiled.API.Extensions
- Added DangerType.None to the DangerType enum
- Added EncounteredPlayer to the ChangingDangerState event (can be null if it isn't an encounter danger)
- Fixed the ChangingDangerState event for all encounters and the damage dangers
- Player.Dangers now returns an empty array if the Scp1853 effect isn't enabled (to prevent issues and errors)
- Added Player.GetDanger(DangerType) and Player.TryGetDanger(DangerType, out DangerStackBase)

### DIFF
--- a/Exiled.API/Enums/DangerType.cs
+++ b/Exiled.API/Enums/DangerType.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------
+// -----------------------------------------------------------------------
 // <copyright file="DangerType.cs" company="Exiled Team">
 // Copyright (c) Exiled Team. All rights reserved.
 // Licensed under the CC BY-SA 3.0 license.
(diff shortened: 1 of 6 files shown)


IkaOverride:
it should be good (I hope because I'm not used to doing big pr) but hey I tested and it works on my machine! pls review code

obvEve (changes requested):
Add ``DangerStackBase GetDanger(DangerType)`` in Player

IkaOverride:
null is returned when the Scp1853 effect isn't active (because it caused issues)

IkaOverride:
new commits are tested

obvEve (approved):

PR #2635, opened 2024-06-22
Changes

Misfiy:


EXILED.props
```diff
@@ -1,52 +1,52 @@
...
+
+    <!-- Disable warning about disabled generation of xml files on debug build -->
+    <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+        <GenerateDocumentationFile>True</GenerateDocumentationFile>
+        <NoWarn>$(NoWarn);SA0001</NoWarn>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
+    </PropertyGroup>
```
VALERA771, line 50: Bruuuuuh

VALERA771 (approved):

EXILED.props
Misfiy: ngl no clue why it did that

louis1706: for me this is fine it's only space change so we don't really care lol

PR #2637, opened 2024-06-23
`[Exiled::Events]` Add ChangingKnobSetting EventPatch attribute

ghost:
Fixes the method not being patched if only the ChangingKnobSetting event is subscribed (which results in the event not being called)


IkaOverride:
tested

obvEve (approved):

PR #2652, opened 2024-07-01
`[Exiled::Events]` "Fix" nw skill issue

ghost:
I don't know why it shows I've deleted and recreated the whole file
The northwood skill issue is they didn't really unsubscribe from some events so they can still execute and call the ChangingDangerEvent when they shouldn't


IkaOverride:
tested

obvEve (approved):

PR #2667, opened 2024-07-02
[API, Modules] Possible null fixes

Misfiy:


Exiled.CustomModules/CustomModules.cs
```diff
@@ -91,8 +91,6 @@ public override void OnEnabled()
                 StaticActor.CreateNewInstance<PickupTracker>();
             }
 
-            SubscribeEvents();
```
Misfiy, line 94: These are called in base.OnEnabled()

Misfiy:
There are still null issues

ALEXWARELLC (approved):

PR #2675, opened 2024-07-03
[Events] Fix ammo not setting on role spawn

ghost:


IkaOverride:
how it works: MonoMods.Utils has a great extension for Dictionary.AddRange! It's even used in another exiled file

IkaOverride:
(the issue was with .AddRange from exiled not working with the dictionary)

IkaOverride:
and it's tested

PR #2718, opened 2024-07-16
Added FeatureGroups (dynamic register)

theDevJade:
Allows you to easily register all listeners in a class and other features!

Like this:
```csharp
var combat = new FeatureGroup("combat");
combat.Supply(Health.Health.HealthIE, Health.Health.HealthUpdate); // Supplying Coroutines
combat.Supply(_limbEvents, _otherHealth); // Supplying classes containing events
combat.Register();
```

And an example of what your event class would look like

```csharp
[Listener()]
public void DoctorInstakill(HurtingEventArgs args)
 {
     if (args.Attacker.Role != RoleTypeId.Scp049) return;
     if (args.Player.IsGodModeEnabled) return;
     args.IsAllowed = false;
     args.Player.Kill(DamageType.Scp049);
 }
```


Exiled.Events/Extra/Events/EventAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Exiled.Events.Extra.Events
+{
+    using System;
+
+    /// <summary>
+    /// An attribute to mark listeners.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, Inherited = false)]
+#pragma warning disable SA1649
+    public sealed class ListenerAttribute : Attribute
```
xNexusACS, line 17: instead of using pragma here, why don't you rename the file to the class name, so you don't need pragmas here

Exiled.Events/Extra/CoroutineManager.cs
```diff
@@ -0,0 +1,79 @@
...
+        private static readonly Dictionary<string, CoroutineHandle> RunningCoroutines = new();
+
+        /// <summary>
+        ///     Starts a single coroutine and keeps track of it.
+        /// </summary>
+        /// <param name="coroutine">The coroutine to start.</param>
+        /// <param name="key">A unique key to identify the coroutine. Defaults to the name of the caller method.</param>
+        public static void StartCoroutine(Func<IEnumerator<float>> coroutine, string key)
+        {
+            if (key != null && RunningCoroutines.ContainsKey(key))
```
xNexusACS, line 30: key can't be null here, since you need to specify one already in the parameter

Directory.Build.props
```diff
@@ -8,4 +8,9 @@
     <ItemGroup>
         <AdditionalFiles Include="$(MSBuildThisFileDirectory)\stylecop.json"/>
     </ItemGroup>
+
+    <PropertyGroup>
+        <!-- Define the EXILED_REFERENCES property -->
```
xNexusACS, line 13: Also remove this

PR #2742, opened 2024-07-21
Remove Log Line from Verified Patch

xNexusACS:


NaoUnderscore:
That was my fault for releasing the stable version with the Resonance installer(s).
Going to close this.

PR #2750, opened 2024-07-22
[Resonance] Quaternion fixes

ITeMbI4:
### Changes

- Make rotation optional for `Pickup.Spawn`, `Pickup.CreateAndSpawn`, `Projectile.Spawn`, `Projectile.CreateAndSpawn` and `Ragdoll.CreateAndSpawn`
- Fix zero `Quaternion`


ITeMbI4:
A bit of context, why I delete `default` from `Quaternion`:

When you create a `Quaternion` with `default` value, it will be - `new Quaternion(0f, 0f, 0f, 0f)`.
So, `myRotation *= Quaternion.AngleAxis(90f, Vector3.up);` will fail to rotate an object, because our rotation is invalid.

So, I just change all zero `Quaternion` to a proper one - `identity = new Quaternion(0f, 0f, 0f, 1f)`

xNexusACS (approved):

PR #2753, opened 2024-07-24
Fixing CustomModules

theDevJade:
Implements a custom TypeConverter for modulepointers using reflection. Needs testing and possible modifications


ALEXWARELLC (commented):
Please implement Unit Tests or include any tests run.

theDevJade:
This new version does not use Type Converters, I have tested it works efficiently and no errors. This version instead uses node deserializers with a branching deserializer tree based on the type. So far it covers RoleSettings & CustomRole<>'s. But more can be added easily, documentation is also added. @ALEXWARELLC

NaoUnderscore (approved):

PR #2731, opened 2024-07-18
`[Exiled::API]` Rework MirrorExtension

NotZer0Two:
Removed all Mirror Extension from MirrorExtensions.cs and Positioned Them inside the classes Extension

PR #2756, opened 2024-07-27
`[EXILED::API]` Adding SendFakeSceneLoading

NotZer0Two:
Adds a new way to load scenes to the client from the server without any client mod, what i could document was in the server, its only present "Facility" but by using [AssetRipper](https://github.com/AssetRipper/AssetRipper) there are 6 scenes in total removing the 2 (NewMainMenu and Facility) that we know there are in total other 4 scenes never seen

They are
**MainMenuRemastered**
Seems some rework or the old menu still semi-working

**FastMenu**
Old menu for the server list

**PreLoader**
Is the loading menu

**Loader**
Black menu before the actual main menu


NotZer0Two:
I've made SceneType for making sure people don't type it wrong or just don't softlock their client if they only want to explore but if in the future some of them are wrong or bugged people can use the first method with string to load or for newer scenes, i will still updated it

Exiled.API/Enums/ScenesType.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Its Loading screen at the start of the game.
+        /// ! Will cause crash when trying joining servers !
+        /// </summary>
+        PreLoader,
+
+        /// <summary>
+        /// A black menu before loading the <see cref="NewMainMenu"/>
+        /// </summary>
+        Loader,
```
ALEXWARELLC, line 46: English better please.

NotZer0Two: Probably better than before

VALERA771:
I have doubts about moving methods from `MirrorExtensions` to other classes

NotZer0Two:
Its something already done, i don't know why github imports old one pr i've made https://github.com/Exiled-Team/EXILED/pull/2731

xNexusACS (changes requested):
Fix build conflicts

6hundred9:
cool

PR #2758, opened 2024-07-27
[Resonance] Remove List from GameEntity

ITeMbI4:
### Changes

- Remove `List`, `GetNearestEntity`, `GetNearestEntities`, `GetFarthestEntity` and `GetFarthestEntities`, because it's causing TPS drops on the server
- Remove all unnecessary usings

### DIFF
--- a/Exiled.API/Extensions/ItemExtensions.cs
+++ b/Exiled.API/Extensions/ItemExtensions.cs
@@ -15,7 +15,6 @@ namespace Exiled.API.Extensions
     using Features.Items;
     using InventorySystem;
     using InventorySystem.Items;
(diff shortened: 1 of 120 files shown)


ITeMbI4:
TPS fix is still under testing, will update once testing will be completed

theDevJade:
Could i see tps before the change and after? @ITeMbI4

ITeMbI4:
Before the fix there were a lot of situations, then TPS drops to 10-30 and stays there for some time. After a fix, it's got stable.
It only happens after 17 players on the server. @theDevJade

ALEXWARELLC (approved):

PR #2760, opened 2024-07-27
[Resonance] Fix flags set

ITeMbI4:
### Changes

- Fix `EffectCategory` setter (previously always return None)
- Fix firearms, don't enable flashlight when created


VALERA771:
Are you sure that you've fixed every such case?

ITeMbI4:
Yes

VALERA771 (approved):

xNexusACS (approved):

PR #2764, opened 2024-08-07
Review

warquys:
I don't want to impose my decisions. But, I want this to be change or at least be heard.
Some stuffs seem to me to have been rushed. So i added comamnt and do litle edition.
I don't know how you prefer to resolve the problem. So i pointed them with comment for the most.

Why not discord:
Discussions here will be easier than discord to keep up with change.

I also try to use the new Enum. But realy i do not get who it work. 

When I fold the code, it seems to me that there is an error in the management of super classes.
And when i try, This isn't working, I don't know if it's me or the code. So I place my experiences here:
<details>
<summary>Interactive Test</summary>
<div>

```cs
> #reset core
Resetting execution engine.
Loading context from 'CSharpInteractive.rsp'.
> using System.Reflection;
>    /// <summary>
     /// An interface for all enum classes.
     /// </summary>
     public interface IEnumClass
     {

     }
> public abstract class UniqueUnmanagedEnumClass<TSource, TObject> : IComparable, IEquatable<TObject>, IComparable<TObject>, IComparer<TObject>, IConvertible, IEnumClass
        where TSource : unmanaged, IComparable, IFormattable, IConvertible, IComparable<TSource>, IEquatable<TSource>
        where TObject : UniqueUnmanagedEnumClass<TSource, TObject>
    {
        private static SortedList<TSource, TObject> values;
        private static int nextValue = int.MinValue;
        private static bool isDefined;

        private string name;

        /// <summary>
        /// Initializes a new instance of the <see cref="UniqueUnmanagedEnumClass{TSource, TObject}"/> class.
        /// </summary>
        public UniqueUnmanagedEnumClass()
        {
            values ??= new();
            TypeCode code = Convert.GetTypeCode(typeof(TSource).GetField("MinValue").GetValue(null));

            // @Nao If the value is not an Uxxxxx it will get an overflowLike if it use an Int16, Byte or SByte
            // Maybe use a long for nextValue and do "nextValue = typeof(TSource).GetField("MinValue").GetValue(null)"
            // it also be a struct containing only unmanged structThe best solution is proably to
            // look for upper version of the C# to resolve this with the new numeric interfaces.
            if (code is TypeCode.UInt16 or TypeCode.UInt32 or TypeCode.UInt64)
                nextValue = 0;

            lock (values)
            {
                TSource value;
                do
                {
                    value = (TSource)Convert.ChangeType(nextValue++, code);
                }
                while (values.ContainsKey(value));

                Value = value;

                // @nao If the value is register when a new instance is created.
                // Maybe put the ctor in protected to avoid exteranal code (outisde the enum) to create new values
                values.Add(value, (TObject)this);
            }
        }

        /// <summary>
        /// Gets all <typeparamref name="TObject"/> object instances.
        /// </summary>
        public static IEnumerable<TObject> Values => values.Values;

        /// <summary>
        /// Gets the value of the enum item.
        /// </summary>
        public TSource Value { get; }

        /// <summary>
        /// Gets the name determined from reflection.
        /// </summary>
        public string Name
        {
            get
            {
                if (isDefined)
                    return name;

                IEnumerable<FieldInfo> fields = typeof(TObject)
                    .GetFields(BindingFlags.Static | BindingFlags.GetField | BindingFlags.Public)
                    .Where(t => t.FieldType == typeof(TObject));

                foreach (FieldInfo field in fields)
                {
                    TObject instance = (TObject)field.GetValue(null);
                    instance.name = field.Name;
                }

                isDefined = true;
                return name;
            }
        }

        /// <summary>
        /// Implicitly converts the <see cref="UniqueUnmanagedEnumClass{TSource, TObject}"/> to <typeparamref name="TSource"/>.
        /// </summary>
        /// <param name="value">The value to convert.</param>
        public static implicit operator TSource(UniqueUnmanagedEnumClass<TSource, TObject> value) => value.Value;

        /// <summary>
        /// Implicitly converts the <typeparamref name="TSource"/> to <see cref="EnumClass{TSource, TObject}"/>.
        /// </summary>
        /// <param name="value">The value to convert.</param>
        public static implicit operator UniqueUnmanagedEnumClass<TSource, TObject>(TSource value) => values[value];

        /// <summary>
        /// Implicitly converts the <see cref="UniqueUnmanagedEnumClass{TSource, TObject}"/> to <typeparamref name="TObject"/>.
        /// </summary>
        /// <param name="value">The value to convert.</param>
        public static implicit operator TObject(UniqueUnmanagedEnumClass<TSource, TObject> value) => values[value];

        /// <summary>
        /// Casts the specified <paramref name="value"/> to the corresponding type.
        /// </summary>
        /// <param name="value">The enum value to be cast.</param>
        /// <returns>The cast object.</returns>
        public static TObject Cast(TSource value) => values[value];

        /// <summary>
        /// Casts the specified <paramref name="values"/> to the corresponding type.
        /// </summary>
        /// <param name="values">The enum values to be cast.</param>
        /// <returns>The cast object.</returns>
        public static IEnumerable<TObject> Cast(IEnumerable<TSource> values)
        {
            foreach (TSource value in values)
                yield return UniqueUnmanagedEnumClass<TSource, TObject>.values[value];
        }

        /// <summary>
        /// Casts the specified <paramref name="values"/> to the corresponding type.
        /// </summary>
        /// <typeparam name="T">The type to cast the enum to.</typeparam>
        /// <param name="values">The enum values to be cast.</param>
        /// <returns>The cast <typeparamref name="T"/> object.</returns>
        public static IEnumerable<T> Cast<T>(IEnumerable<TSource> values)
            where T : TObject
        {
            foreach (TSource value in values)
                yield return Cast<T>(value);
        }

        /// <summary>
        /// Casts the specified <paramref name="value"/> to the corresponding type.
        /// </summary>
        /// <typeparam name="T">The type to cast the enum to.</typeparam>
        /// <param name="value">The enum value to be cast.</param>
        /// <returns>The cast <typeparamref name="T"/> object.</returns>
        public static T Cast<T>(TSource value)
            where T : TObject => (T)values[value];

        /// <summary>
        /// Safely casts the specified <paramref name="value"/> to the corresponding type.
        /// </summary>
        /// <param name="value">The enum value to be cast.</param>
        /// <param name="result">The cast <paramref name="value"/>.</param>
        /// <returns><see langword="true"/> if the <paramref name="value"/> was cast; otherwise, <see langword="false"/>.</returns>
        public static bool SafeCast(TSource value, out TObject result) => values.TryGetValue(value, out result);

        /// <summary>
        /// Safely casts the specified <paramref name="values"/> to the corresponding type.
        /// </summary>
        /// <param name="values">The enum value to be cast.</param>
        /// <param name="results">The cast <paramref name="values"/>.</param>
        /// <returns><see langword="true"/> if the <paramref name="values"/> was cast; otherwise, <see langword="false"/>.</returns>
        public static bool SafeCast(IEnumerable<TSource> values, out IEnumerable<TObject> results)
        {
            results = null;

            List<TObject> tmpValues = new List<TObject>();
            foreach (TSource value in values)
            {
                if (!UniqueUnmanagedEnumClass<TSource, TObject>.values.TryGetValue(value, out TObject result))
                    return false;

                tmpValues.Add(result);
            }

            results = tmpValues;
            return true;
        }

        /// <summary>
        /// Retrieves an array of the values of the constants in a specified <see cref="UnmanagedEnumClass{TSource, TObject}"/>.
        /// </summary>
        /// <param name="type">The <see cref="UnmanagedEnumClass{TSource, TObject}"/> type.</param>
        /// <returns>An array of the values of the constants in a specified <see cref="UnmanagedEnumClass{TSource, TObject}"/>.</returns>
        public static TSource[] GetValues(Type type)
        {
            if (type is null)
                throw new NullReferenceException("The specified type parameter is null");

            if (!type.IsSubclassOf(typeof(UniqueUnmanagedEnumClass<TSource, TObject>)) && type.BaseType != typeof(UniqueUnmanagedEnumClass<TSource, TObject>))
                throw new Exception("The specified type parameter is not a UniqueUnmanagedEnumClass<TSource, TObject> type.");

            return typeof(TSource)
                .GetFields(BindingFlags.Static | BindingFlags.Public | BindingFlags.GetField)
                .Where(field => field.FieldType == typeof(TSource))
                .Select(field => (TSource)field.GetValue(null))
                .ToArray();
        }

        /// <summary>
        /// Parses a <see cref="string"/> object.
        /// </summary>
        /// <param name="obj">The object to be parsed.</param>
        /// <returns>The corresponding <typeparamref name="TObject"/> object instance, or <see langword="null"/> if not found.</returns>
        public static TObject Parse(string obj)
        {
            foreach (TObject value in values.Values.Where(value => string.Compare(value.Name, obj, StringComparison.OrdinalIgnoreCase) == 0))
                return value;

            return null;
        }

        /// <summary>
        /// Converts the <see cref="UnmanagedEnumClass{TSource, TObject}"/> instance to a human-readable <see cref="string"/> representation.
        /// </summary>
        /// <returns>A human-readable <see cref="string"/> representation of the <see cref="UnmanagedEnumClass{TSource, TObject}"/> instance.</returns>
        public override string ToString() => Name;

        /// <summary>
        /// Determines whether the specified object is equal to the current object.
        /// </summary>
        /// <param name="obj">The object to compare.</param>
        /// <returns><see langword="true"/> if the object was equal; otherwise, <see langword="false"/>.</returns>
        public override bool Equals(object obj) =>
            obj != null && (obj is TSource value ? Value.Equals(value) : obj is TObject derived && Value.Equals(derived.Value));

        /// <summary>
        /// Determines whether the specified object is equal to the current object.
        /// </summary>
        /// <param name="other">The object to compare.</param>
        /// <returns><see langword="true"/> if the object was equal; otherwise, <see langword="false"/>.</returns>
        public bool Equals(TObject other) => Value.Equals(other.Value);

        /// <summary>
        /// Returns a the 32-bit signed hash code of the current object instance.
        /// </summary>
        /// <returns>The 32-bit signed hash code of the current object instance.</returns>
        public override int GetHashCode() => Value.GetHashCode();

        /// <summary>
        /// Compares the current instance with another object of the same type and returns
        /// an integer that indicates whether the current instance precedes, follows, or
        /// occurs in the same position in the sort order as the other object.
        /// </summary>
        /// <param name="other">An object to compare with this instance.</param>
        /// <returns>
        /// A value that indicates the relative order of the objects being compared.
        /// The return value has these meanings: Value Meaning Less than zero This instance precedes other in the sort order.
        /// Zero This instance occurs in the same position in the sort order as other.
        /// Greater than zero This instance follows other in the sort order.
        /// </returns>
        public int CompareTo(TObject other) => Value.CompareTo(other.Value);

        /// <summary>
        /// Compares the current instance with another object of the same type and returns
        /// an integer that indicates whether the current instance precedes, follows, or
        /// occurs in the same position in the sort order as the other object.
        /// </summary>
        /// <param name="obj">An object to compare with this instance.</param>
        /// <returns>
        /// A value that indicates the relative order of the objects being compared.
        /// The return value has these meanings: Value Meaning Less than zero This instance precedes other in the sort order.
        /// Zero This instance occurs in the same position in the sort order as other.
        /// Greater than zero This instance follows other in the sort order.
        /// </returns>
        public int CompareTo(object obj) =>
            obj == null ? -1 : obj is TSource value ? Value.CompareTo(value) : obj is TObject derived ? Value.CompareTo(derived.Value) : -1;

        /// <summary>
        /// Compares the specified object instance with another object of the same type and returns
        /// an integer that indicates whether the current instance precedes, follows, or
        /// occurs in the same position in the sort order as the other object.
        /// </summary>
        /// <param name="x">An object to compare.</param>
        /// <param name="y">Another object to compare.</param>
        /// <returns>
        /// A value that indicates the relative order of the objects being compared.
        /// The return value has these meanings: Value Meaning Less than zero This instance precedes other in the sort order.
        /// Zero This instance occurs in the same position in the sort order as other.
        /// Greater than zero This instance follows other in the sort order.
        /// </returns>
        public int Compare(TObject x, TObject y) => x == null ? -1 : y == null ? 1 : x.Value.CompareTo(y.Value);

		/// <inheritdoc/>
        TypeCode IConvertible.GetTypeCode() => Value.GetTypeCode();

        /// <inheritdoc/>
        bool IConvertible.ToBoolean(IFormatProvider provider) => Value.ToBoolean(provider);

        /// <inheritdoc/>
        char IConvertible.ToChar(IFormatProvider provider) => Value.ToChar(provider);

        /// <inheritdoc/>
        sbyte IConvertible.ToSByte(IFormatProvider provider) => Value.ToSByte(provider);

        /// <inheritdoc/>
        byte IConvertible.ToByte(IFormatProvider provider) => Value.ToByte(provider);

        /// <inheritdoc/>
        short IConvertible.ToInt16(IFormatProvider provider) => Value.ToInt16(provider);

        /// <inheritdoc/>
        ushort IConvertible.ToUInt16(IFormatProvider provider) => Value.ToUInt16(provider);

        /// <inheritdoc/>
        int IConvertible.ToInt32(IFormatProvider provider) => Value.ToInt32(provider);

        /// <inheritdoc/>
        uint IConvertible.ToUInt32(IFormatProvider provider) => Value.ToUInt32(provider);

        /// <inheritdoc/>
        long IConvertible.ToInt64(IFormatProvider provider) => Value.ToInt64(provider);

        /// <inheritdoc/>
        ulong IConvertible.ToUInt64(IFormatProvider provider) => Value.ToUInt64(provider);

        /// <inheritdoc/>
        float IConvertible.ToSingle(IFormatProvider provider) => Value.ToSingle(provider);

        /// <inheritdoc/>
        double IConvertible.ToDouble(IFormatProvider provider) => Value.ToDouble(provider);

        /// <inheritdoc/>
        decimal IConvertible.ToDecimal(IFormatProvider provider) => Value.ToDecimal(provider);

        /// <inheritdoc/>
        DateTime IConvertible.ToDateTime(IFormatProvider provider) => Value.ToDateTime(provider);

        /// <inheritdoc/>
        string IConvertible.ToString(IFormatProvider provider) => ToString();

        /// <inheritdoc/>
        object IConvertible.ToType(Type conversionType, IFormatProvider provider) => Value.ToType(conversionType, provider);
    }

> public class Enum1 : UniqueUnmanagedEnumClass<uint, Enum1>
{
    /// <summary>
    /// Represents an invalid custom role.
    /// </summary>
    public static readonly Enum1 None = new();

    public static readonly Enum1 Foo = new();

    public static readonly Enum1 Bar = new();
}

public class Enum2 : UniqueUnmanagedEnumClass<uint, Enum1>
{
    /// <summary>
    /// Represents an invalid custom role.
    /// </summary>
    public static readonly Enum1 None = new();

    public static readonly Enum1 Foo = new();
    
        public static readonly Enum1 Bar = new();
}

public class Enum1Ex : Enum1
{
    public static readonly Enum1 ExtenedValue = new();

    public static readonly Enum1Ex ExtenedValue2 = new();
}

public class Enum2Ex : Enum2
{
    public static readonly Enum2 ExtenedValue1 = new();

    public static readonly Enum2Ex ExtenedValue2 = new();
}
> Enum1.Bar.Name
"Bar"
> Enum1Ex.Bar.Name
"Bar"
> Enum1Ex.ExtenedValue.Name
null
> Enum1Ex.ExtenedValue2.Name
null
> Enum2Ex.ExtenedValue1.Name
System.TypeInitializationException: The type initializer for 'Enum2Ex' threw an exception.
> Enum2Ex.ExtenedValue2.Name
System.TypeInitializationException: The type initializer for 'Enum2Ex' threw an exception.
> Enum2Ex.ExtenedValue2
System.TypeInitializationException: The type initializer for 'Enum2Ex' threw an exception.
> try { _ = Enum2Ex.ExtenedValue2.Name; } catch (Exception e) { WriteLine(e?.InnerException?.ToString() ?? "None"); }
System.InvalidCastException: Unable to cast object of type 'Enum2' to type 'Enum1'.
   at Submission#4.UniqueUnmanagedEnumClass`2..ctor()
   at Submission#5.Enum2..ctor()
   at Submission#5.Enum2Ex..cctor()
> public class Enum1 : UniqueUnmanagedEnumClass<uint, Enum1>
{
    /// <summary>
    /// Represents an invalid custom role.
    /// </summary>
    public static readonly Enum1 None = new();

    public static readonly Enum1 Foo = new();

    public static readonly Enum1 Bar = new();
}

public class Enum2 : UniqueUnmanagedEnumClass<uint, Enum2>
{
    /// <summary>
    /// Represents an invalid custom role.
    /// </summary>
    public static readonly Enum1 None = new();

    public static readonly Enum1 Foo = new();
    
    public static readonly Enum1 Bar = new();
}

public class Enum1Ex : Enum1
{
    public static readonly Enum1 ExtenedValue1 = new();

    public static readonly Enum1Ex ExtenedValue2 = new();
}

public class Enum2Ex : Enum2
{
    public static readonly Enum2 ExtenedValue1 = new();

    public static readonly Enum2Ex ExtenedValue2 = new();
}
> Enum1.Bar.Name
"Bar"
> Enum1Ex.Bar.Name
"Bar"
> Enum1Ex.ExtenedValue1.Name
null
> Enum1Ex.ExtenedValue2.Name
null
> Enum2Ex.ExtenedValue1.Name
null
```
</div>
</details>


NaoUnderscore:
Interesting, we aren't supporting multiple inheritance levels, we must fix it.

Exiled.CustomModules/API/Features/RespawnManager.cs
```diff
@@ -174,21 +174,24 @@ private void OnPreRespawningTeam(PreRespawningTeamEventArgs ev)
         {
             PreviousKnownTeam = NextKnownTeam;
 
+            // @Nao are you sur that is "or SpawnableTeamType" and not "or not SpawnableTeamType"
```
NaoUnderscore, line 177: I'm sure

Exiled.CustomModules/API/Features/RespawnManager.cs
```diff
@@ -174,21 +174,24 @@ private void OnPreRespawningTeam(PreRespawningTeamEventArgs ev)
...
-            if (customTeam.TeamsOwnership.Any(t =>
-                    t == (ev.NextKnownTeam is SpawnableTeamType.ChaosInsurgency ?
-                        Team.ChaosInsurgency : Team.FoundationForces)))
+            if (customTeam.TeamsOwnership.Any(t => t == (Team)ev.NextKnownTeam))
             {
                 ev.MaxWaveSize = customTeam.Size;
                 return;
             }
 
+            // @Nao, it cool to use the event system for this. But if an other plugin
```
NaoUnderscore, line 191: Right

Exiled.API/Features/Core/Components/TickComponent.cs
```diff
@@ -24,7 +24,7 @@ public sealed class TickComponent : EObject
         /// The default fixed tick rate (60 per second).
         /// </summary>
 #pragma warning disable SA1310
-        public const float DEFAULT_FIXED_TICK_RATE = 0.016f;
+        public const float DEFAULT_FIXED_TICK_RATE = 1f / 60f;
```
ALEXWARELLC, line 27: While this produces the same result, it just produces roughly the same result. As seen below, the difference is miniscule. If you feel that this works better, it would need to be tested and monitored in a controlled environment for an extended period of time.

image

Exiled.API/Features/Core/GameEntity.cs
```diff
@@ -44,11 +44,6 @@ protected GameEntity(GameObject gameObject)
             // List.Add(this);
         }
 
-        /// <summary>
-        /// Finalizes an instance of the <see cref="GameEntity"/> class.
-        /// </summary>
-        ~GameEntity() => List.Remove(this);
-
```
ALEXWARELLC, line 51: Why is this being removed?

Exiled.API/Features/Core/Generic/EBehaviour.cs
```diff
@@ -60,6 +60,8 @@ protected EBehaviour()
         /// </remarks>
         protected virtual void FindOwner()
         {
+            // @Nao T is a GameEntity, why not add in inisde of it the method a abstract.
```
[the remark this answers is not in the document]
ALEXWARELLC: If Get's are static, you shouldn't need to get it using `GetMethod`. Unless directly calling it through the class doesn't work, you shouldn't need to make reflection calls.

Exiled.API/Features/Core/Generic/Singleton.cs
```diff
@@ -35,6 +38,8 @@ public Singleton(T value)
         /// <summary>
         /// Finalizes an instance of the <see cref="Singleton{T}"/> class.
         /// </summary>
+        // @Nao Probably never call, it self reference. Maybe an future issue doing memory leek
```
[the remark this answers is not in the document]
ALEXWARELLC: First: Read the English dictionary, please.
Second: This runs before the class is ultimately disposed, so in theory, there shouldn't be a memory leak, but if you're so unsure, you can implement a disposable method, but it's generally best to use `IDisposable`s with `using` blocks.

Exiled.API/Features/Core/Generic/Singleton.cs
```diff
@@ -72,13 +81,7 @@ public static bool TryGet<TObject>(out TObject instance)
...
         public static bool Destroy(T @object)
         {
-            if (Instances.TryGetValue(@object, out Singleton<T> _))
-            {
-                Instances[@object] = null;
-                return Instances.Remove(@object);
-            }
-
-            return false;
+            return Instances.Remove(@object);
```
ALEXWARELLC, line 84: The old code, while may seem less efficient, is safer, considering the `ArgumentNullException` can occur creating issues further down the line. `TryGetValue` returns a boolean, which, in this context, stops bad, or null values from causing an exception.

Exiled.API/Features/Respawn.cs
```diff
@@ -23,6 +23,9 @@ namespace Exiled.API.Features
     /// </summary>
     public static class Respawn
     {
+        private const string MtfChopperGameObjectName = "Chopper";
+        private const string ChaosVanGameObjectName = "CIVanArrive";
```
ALEXWARELLC, line 27: I don't see how moving immutable strings to a constant variable is any different to just hard-referencing the exact same string.

Exiled.CustomModules/API/Features/CustomRoles/CustomTeam.cs
```diff
@@ -35,6 +35,9 @@ namespace Exiled.CustomModules.API.Features.CustomRoles
     /// </remarks>
     public abstract class CustomTeam : CustomModule
     {
+        private const string VanilaConfigMtfMinTime = "minimum_MTF_time_to_spawn";
+        private const string VanilaConfigMtfMaxTime = "maximum_MTF_time_to_spawn";
+
```
ALEXWARELLC, line 40: Again, I don't see how moving immutable strings to a constant variable is any different to just hard-referencing the exact same string.

ghost (commented):
I'm gonna ignore the Grade 12 spelling.

Exiled.API/Features/Core/Generic/EBehaviour.cs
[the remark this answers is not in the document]
warquys: Sorry, exiles use CSharplang 9 and the static interface method is only available in 11.

So if using reflection is the solution. Can we add a check like this.
```cs
            if (typeof(T).IsAssignableFrom(method.ReturnType))
```
And throw if false. This will help to better understand what is wrong and avoid null values in `Owner`.

Exiled.API/Features/Core/Generic/Singleton.cs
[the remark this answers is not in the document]
warquys: The only solution to make it disposable is to call, 
```cs
public static bool Destroy(T @object)
```
If this is done dtor code becomes useless.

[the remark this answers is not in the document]
warquys: @ALEXWARELLC 

1. I don't know how I should interpret it...

2. Making the singleton disposable is not realy usefull. I find it strange the fact that the singleton is an instance. But adding a disposable singleton make this realy strange. 
Most of the time when I use a singleton it is to have an object reference becomes "static". And keep it alive a long time.
Adding disposable and use it with `using` wil result to have short live singleton. Not really useful if you considere to share the instance between several plugins.

I say memory leak bc, if we create sub 
```cs
> GC.Collect();
public class SelfStoring
{
    public SelfStoring()
    {
    }


    ~SelfStoring()
    {
        Console.WriteLine("dtor call!");

    }
}

new SelfStoring();
new SelfStoring();
new SelfStoring();
new SelfStoring();
> GC.Collect();
dtor call!
dtor call!
dtor call!
dtor call!
> public class SelfStoring
{
    static List<SelfStoring> store = new List<SelfStoring>();

    public SelfStoring()
    {
        store.Add(this);
    }


    ~SelfStoring()
    {
        store.Remove(this);
        Console.WriteLine("dtor call!");

    }
}

new SelfStoring();
new SelfStoring();
new SelfStoring();
new SelfStoring();
> GC.Collect();
> GC.Collect(GC.MaxGeneration, GCCollectionMode.Aggressive);
```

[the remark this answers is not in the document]
ALEXWARELLC: I had a stroke reading this,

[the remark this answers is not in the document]
warquys: Of course a memory leak will only occur if a plugin starts creating singltons. If no memory is generated then there will be no problem. However, removing a reference in the dtor does not make sense (expressed in the case of a weak reference).

Exiled.API/Features/Core/Generic/UniqueUnmanagedEnumClass.cs
```diff
@@ -53,6 +57,9 @@ public UniqueUnmanagedEnumClass()
                 while (values.ContainsKey(value));
 
                 Value = value;
+
+                // @nao If the value is register when a new instance is created.
+                // Maybe put the ctor in protected to avoid exteranal code (outisde the enum) to create new values
```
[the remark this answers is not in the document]
warquys: `internal protected` can be cool if Exiled need to create dedicated new instance in an helper class

Exiled.API/Features/Core/Generic/Singleton.cs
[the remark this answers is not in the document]
ALEXWARELLC: > Of course a memory leak will only occur if a plugin starts creating singltons. If no memory is generated then there will be no problem. However, removing a reference in the dtor does not make sense (expressed in the case of a weak reference).

👍 Understood.

Exiled.CustomModules/API/Features/RespawnManager.cs
warquys: Is it not better to do `is not uint` to avoid [casting exeption](https://github.com/warquys/EXILED/blob/1841bbb35953e5a5e02138017317d1a413d5cb34/Exiled.CustomModules/API/Features/RespawnManager.cs#L181) ?

Exiled.API/Features/Core/Components/TickComponent.cs
warquys: For spelling errors: I did not have a suitable spell checker on the computer where I wrote this code and is not my mothertongue.

Roslyne can do math optimisation and avoid floating-point error.
In my country floating-point error are see durring the Grade 12 or first year when pushing Bachelor's degree...
0.016f * 60f result in 0.96;
(1f / 60f) * 60f result in 1.

Exiled.API/Features/Core/GameEntity.cs
warquys: Never call, [same reason](https://github.com/Exiled-Team/EXILED/pull/2764#discussion_r1712646105).

Exiled.API/Features/Core/Generic/Singleton.cs
warquys: What do you mean by "effective"? Many optimizations can be made. Memory, speed, maintainability... ? 
Can you be a little more specific.

And no the old code is not safer for null
```cs
> Dictionary<Object, int> dic = new();
. var obj = new Object();
. dic[obj] = 10;
> dic.ContainsKey(obj)
true
> dic.ContainsKey(null)
System.ArgumentNullException: Value cannot be null. (Parameter 'key')
  + System.ThrowHelper.ThrowArgumentNullException(System.ExceptionArgument)
  + Dictionary<TKey, TValue>.FindValue(TKey)
  + Dictionary<TKey, TValue>.ContainsKey(TKey)
> dic.TryGetValue(null, out _)
. // do "_" and not "var _"
. // "var _" result to create a new var call "_"
System.ArgumentNullException: Value cannot be null. (Parameter 'key')
  + System.ThrowHelper.ThrowArgumentNullException(System.ExceptionArgument)
  + Dictionary<TKey, TValue>.FindValue(TKey)
  + Dictionary<TKey, TValue>.TryGetValue(TKey, out TValue)

```

if you whant to avoid `ArgumentNullException` do a check beffor.

Exiled.API/Features/Respawn.cs
warquys: facilitate maintenance, avoid searching everywhere in the class for names that might change. And allows you to use the same name everywhere and edit only in one place.

Exiled.CustomModules/API/Features/CustomRoles/CustomTeam.cs
warquys: https://github.com/Exiled-Team/EXILED/pull/2764#discussion_r1712663457

Exiled.API/Features/Respawn.cs
ALEXWARELLC: > facilitate maintenance, avoid searching everywhere in the class for names that might change. And allows you to use the same name everywhere and edit only in one place.

Since these values are only ever used once, I don't see the need for it, but I understand how it's more efficient for maintainability.

Exiled.CustomModules/API/Features/CustomRoles/CustomTeam.cs
ALEXWARELLC: https://github.com/Exiled-Team/EXILED/pull/2764#discussion_r1712665478

Exiled.API/Features/Respawn.cs
warquys: You only need to look up in the class if the GameObject name change and not in the all class. 
And that is use full to avoid magical value.

warquys:
I do not get the point of [TypeCastObject](https://github.com/Exiled-Team/EXILED/blob/f620d33dc9ad3462eaaa962ded8881e77cfcda58/Exiled.API/Features/Core/TypeCastObject.cs#L14) do the C# default `is`, `(T)` and `as` has limitation ? Or is it to add new "Castable" that are not of the same type. 

I know that implementing cast operators don't allow `is` and `as`. However, if the method requires a specific type. There is no problem because the cast will take place before the call.

warquys:
@NaoUnderscore I rewrite the Singleton, are you ok ?

NaoUnderscore:
> I do not get the point of [TypeCastObject](https://github.com/Exiled-Team/EXILED/blob/f620d33dc9ad3462eaaa962ded8881e77cfcda58/Exiled.API/Features/Core/TypeCastObject.cs#L14) do the C# default `is`, `(T)` and `as` has limitation ? Or is it to add new "Castable" that are not of the same type.
> 
> I know that implementing cast operators don't allow `is` and `as`. However, if the method requires a specific type. There is no problem because the cast will take place before the call.
> 
> Is it to do runtime editable castable object ? If yes, where do you need this ? I do not see an usage of ITypeCast in any method. And the class using TypeCastMono and TypeCastObject only use the `is`, `(T)` and `as` with static generique.

That's a way to make everything more versatile, so we can safely define our own `Cast` overloads, which means that if we want to change something in `TypeCaseObject` that's going to be reflected to all classes using it, it's useful in case of custom modules.

NaoUnderscore:
> @NaoUnderscore I rewrite the Singleton, are you ok ?

Go ahead, I've made some changes to `StaticActor` and their generic version as well.

warquys:
I no longer have time to work on this project.
Sorry.

Exiled.CustomModules/API/Features/RespawnManager.cs
NaoUnderscore: Nah, we're kinda safe on it.

PR #2766, opened 2024-08-11
`[EXILED::API]` Adding MoveNetworkIdentityObject

NotZer0Two:
Lets the user move NetworkIdentities it can be bugged for the current culling system, in theory you can move object with a radius of 500m (Very small) and that will let them load in, could be useful for moving objects that you don't want in the facility to spawn or for example the Bowl of Candy


NotZer0Two:
image

This is what you can do

Exiled.API/Features/Player.cs
```diff
@@ -4107,6 +4107,23 @@ public void SendFakeSceneLoading(string newSceneName)
         /// <param name="newSceneName">The new Scene the client will load.</param>
         public void SendFakeSceneLoading(ScenesType newSceneName) => SendFakeSceneLoading(newSceneName.ToString());
 
+        /// <summary>
+        /// Moves object for the player.
```
NaoUnderscore, line 4111:
```suggestion
        /// Moves an object for the player.
```

Exiled.API/Features/Server.cs
```diff
@@ -356,5 +356,25 @@ public static void ChangeSceneToAllClients(string newSceneName)
         /// </summary>
         /// <param name="scene">The new Scene the client will load.</param>
         public static void ChangeSceneToAllClients(ScenesType scene) => ChangeSceneToAllClients(scene.ToString());
+
+        /// <summary>
+        /// Moves object for all the players.
```
NaoUnderscore, line 361:
```suggestion
        /// Moves an object for all players.
```

NaoUnderscore (changes requested):

NaoUnderscore:
Fix conflicts

Monaldcry7788 (approved):

xNexusACS (approved):

PR #2786, opened 2024-10-23
Recoded Player.PreAuthenticating wia Patch

Marikider:
Recoded PreAuthenticating event wia patch, so IsAllowed will work.

### DIFF
--- a/Exiled.Events/EventArgs/Player/PreAuthenticatingEventArgs.cs
+++ b/Exiled.Events/EventArgs/Player/PreAuthenticatingEventArgs.cs
@@ -103,7 +103,7 @@ namespace Exiled.Events.EventArgs.Player
         /// <summary>
         /// Gets a value indicating whether the player can be authenticated or not.
         /// </summary>
-        public bool IsAllowed { get; private set; } = true;
+        public bool IsAllowed { get; set; } = true;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


xNexusACS (changes requested):

xNexusACS (approved):

xNexusACS (changes requested):
Fix build issues

xNexusACS (dismissed):

xNexusACS (approved):

xNexusACS:
Just some questions I have before merging this, what happens if `IsAllowed` is set to false? The player auth is rejected and go back to the main menu or the player is stuck in a loop in the loading screen?

Is this also fully tested? All properties returning the correct value, also making sure that this doesn't cause any issue with whitelists and server max slots.

Marikider:
When IsAllowed is set to false, player goes back to lobby with Server Not Responding Message.  All Values in event args are same with original event are correct. Just tested it on reserved slots and whitelists, it is working and not cousing any troubles with it.

xNexusACS:
> When IsAllowed is set to false, player goes back to lobby with Server Not Responding Message. All Values in event args are same with original event. Just tested it on reserved slots and whitelists, it is working and not cousing any troubles with it.

nice

PR #2802, opened 2024-12-26
14.0

Logic-Gun:
some changes (shot dont works)

Someone need rewrite code on transpiller to RespawningTeam Event (To get players and add DeployingTeamRole event)


Logic-Gun:
image
remained

